PR #43, opened 2015-01-24
Add domain check to flip

zaxtax:


stuhlmueller:
Thanks for the pull request!

I wonder if we should be even more strict and do `if (val !== true && val !== false) { return -Infinity; }`.

zaxtax:
Stricter is always better. Though JavaScript's inability to distinguish ints and doubles means a little more work to do that sometimes.

PR #32, opened 2015-01-31
Change test suite implementation to use .wppl and .json files

pedir:
Currently, the tests in [test-inference.js](https://github.com/probmods/webppl/blob/master/tests/test-inference.js) are such that the model code is included as a string in the test file. This makes it difficult to add tests, in particular for bigger models. Replace this with a `tests` folder that contains .wppl files with the model code, and corresponding .json files with the desired return distribution and tolerance (or expectation, etc). Change the test runner to list all .wppl files in the directory and run a test for each.


pedir:
Change the target branch from master to dev.

PR #49, opened 2015-02-20
Stacktrace

pedir:
Add basic stack tarcing to webppl files so that exceptions contain information regarding the position of the error in the original wppl file. Right now both the the new webppl stack trace and the old compiled js stack trace are shown side by side.


pedir:
The new branch (https://github.com/probmods/webppl/tree/stack_trace) contains a more updated version of webppl stack trace.

PR #85, opened 2015-04-13
Fix 'grunt hint' command.

pedir:
Fix some of the jshint errors including styling issues and missing semicolons.


stuhlmueller:
Not high priority, but it could be good to deal with the remaining issues that jshint returns (possibly by making jshint less strict) and then to add jshint to the build process, like we do for gjslint.

PR #93, opened 2015-05-04
Add regressions examples to nodeunit tests.

pedir:
Refactor test-inference.js to add linear and logistic regression examples to unit test.


tests/test-inference.js
```diff
@@ -43,19 +43,44 @@ function runDiscreteSamplingTest(test, code, expectedHist, numSamples, tolerance
...
+
+function runDistributionStatisticsTest(test, testData) {
+  var k = function(s, erp) {
+    var normHist = getHist(erp);
+    var testMean = util.expectation(normHist);
+    var testStd = util.std(normHist);
+    var expected = testData.expected;
+    var allOk = true;
+    if (testMean - testStd < expected.mean - expected.std ||
+        testMean + testStd > expected.mean + expected.std) {
```
stuhlmueller: I'm confused by this condition. I expected something like

    testMean < expected.mean - expected.std

or

    testMean < expected.mean - tolerance

It seems that you are comparing (actual mean - actual std) to (expected mean - expected std). Shouldn't we expect these quantities to be approximately equal, and so the test to fail half the time, even if it has the correct mean and std?

pedir:
First, I implemented it similar to what you mentioned (`testMean < expected.mean - tolerance`) but the problem was that both mean and std deviated from their expected value rather considerably. 

For instance when I run the logistic regression 20 times, the min std of those runs are ~0.00003 and max of ~ 0.414 (for its mean it's 0.73 and 0.99 respectively) so I don't think there's actually one "true" mean and "std" that we can compare the tests to since the shape of the distribution heavily depends on the observed chains. That's why I changed it to more of a "confidence interval" check, i.e. if the mass of the test distribution falls between a larger interval (`expected.mean - expected.std`) we would accept.

Maybe a better approach is to check whether (`testStd < some upper bound for expected.std`) and (`abs(testMean - expected.mean) < tolerance`)? Or we could run the test several times and aggregate the results into one average mean and average std and check those with expected mean and std within a tolerance, however, this would slow down our tests.

stuhlmueller:
Are you saying that MCMC doesn't converge for the number of samples we are taking, and so we get different means and stds for different runs? If you discard burn-in and run it for longer, does this still hold true?

pedir:
They do! I was just using not enough samples, sorry about that. I updated the comparison method.

stuhlmueller:
It seems that this still sometimes fails - maybe you need even more samples? Run it a bunch of times and check that it reliably passes, with room to spare?

pedir:
I made the threshold larger so tests shouldn't fail anymore. I'll fix the threshold once we merge the MH burn in logic.

stuhlmueller:
I'll merge this for now, but remember to revisit this later and make the tolerance threshold smaller once burn-in is added to mcmc.

Issue #36, opened 2015-01-19
sampleWithFactor not working


williamleif:
The sampleWithFactor function is broken in both master and dev branches. This means the dippl.org examples using that function are currently broken as well.

stuhlmueller:
The cause of this issue is that `Enumerate.prototype.sampleWithFactor` and `Enumerate.prototype.sample` use the score function by calling it in non-tail position. More precisely, `Enumerate.prototype.sampleWithFactor` constructs a helper function as follows:

    coroutine.sample(s,cc,a,dist,params,
                     function(v){
                       var ret;
                       scoreFn(s, function(s, x){ret = x;}, a+"swf", v);
                       return ret;});

`Enumerate.prototype.sample` calls this helper function in a `for` loop. This isn't supported in cps style and causes problems now that we are using trampolining. For example, a compiled `scoreFn` may look like this:

    function (globalStore, _k50, address, v) {
      var _return = _k50;
      var _k51 = _return;
      var _test52 = v;
      _test52 ? _trampoline = function () {
        _k51(globalStore, 0);
      } : _trampoline = function () {
        _k51(globalStore, -Infinity);
      };
    }

So, it never returns and assigns a function to the global `_trampoline` variable, which would return a value to the continuation (`function(s, x){ret = x;}` above) if called. Assigning a local continuation to the global `_trampoline` variable is problematic, because it may override the actual top-level continuation.

One way to fix this is to use `cpsForEach` in `Enumerate.prototype.sample`, but maybe there is a simpler solution?

For now, I have commented-out the broken `Enumerate.prototype.sampleWithFactor` implementation, which makes `sampleWithFactor` semantically correct again (but no different from using sample and factor individually).

PR #95, opened 2015-05-05
sampleWithFactor Fix

null-a:
This fixes `sampleWithFactor` using `cpsForEach` as discussed in #36. I tested it by running `hmmIncremental` and `hmmSampleWithFactor`, the outputs from which are identical.

I'm not currently extending the stack address when calling `extraScoreFn` as my first impression was that it's not necessary. Am I right?

### DIFF
--- a/examples/hmmSampleWithFactor.wppl
+++ b/examples/hmmSampleWithFactor.wppl
@@ -16,7 +16,7 @@ var hmmRecur = function(n, states, observations) {
   var newObservations = observations.concat([newObs])
   return ((n == 1) ?
           {states: newStates, observations: newObservations} :
-          hmmRecur(n - 1, states, observations));
+          hmmRecur(n - 1, newStates, newObservations));
 }
 
 var hmm = function(n) {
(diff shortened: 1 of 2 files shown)


stuhlmueller:
Neat, thanks!

Since `extraScoreFn` shouldn't make any calls to `sample`, it's not necessary to pass the address.

I wonder if running `cpsForEach` in each sample statement is significantly slower than the previous `for (var s in supp)` procedure. If so, it could be better to leave the previous `sample` implementation unchanged and just move copy over your new implementation to `sampleWithFactor`.

null-a:
Thanks. I did do a quick benchmark to check it wasn't horrendously slow, but I agree it makes sense to do this. Good call!

stuhlmueller:
Thanks for the change!

Two minor comments: 

1. Rename `queueContinuation` to `enqueueContinuation`? I think it's a bit clearer. My first thought was that `queueContinuation` might call a continuation.

2. Could you refer to `env.coroutine` instead of assigning `coroutine = this` and referring to `coroutine`? This shouldn't change anything, and is more in line with how we're accessing the coroutine elsewhere at the moment.

Other than that, looks great!  If you make those changes, feel free to merge the pull request yourself.

null-a:
No problem! Good suggestions - I've made the changes and merged.

PR #99, opened 2015-05-07
more efficient MH

iffsid:
- use hashmap for efficiency (similar to DanR's version)
 - other assorted minor efficiencies
 - modularize so reuse is better (in future branches/commits to merge into dev)
   [deepcopytrace moved into MH to faciliate exports to other places where mh may be used]

 Changes to smc.js are superficial and were needed to make tests work because of
 dependence on old MH stuff. Once this is merged, smc can be updated to use new MH
 and pf with var#factors.

### DIFF
--- a/src/inference/mh.js
+++ b/src/inference/mh.js
@@ -9,125 +9,157 @@ var erp = require('../erp.js');
 
 module.exports = function(env) {
 
-  function findChoice(trace, name) {
-    if (trace === undefined) {
-      return undefined;
-    }
-    for (var i = 0; i < trace.length; i++) {
-      if (trace[i].name === name) {
-        return trace[i];
-      }
-    }
-    return undefined;
+  function makeTraceEntry(s, k, name, erp, params, currScore, choiceScore, val, reuse) {
+    return {k: k, name: name, erp: erp, params: params, score: currScore,
+      choiceScore: choiceScore, val: val, reused: reuse, store: s};
   }
 
-  function acceptProb(trace, oldTrace, regenFrom, currScore, oldScore) {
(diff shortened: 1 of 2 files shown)


ngoodman:
We should coordinate this with @dritchie, since we will probably want to merge his fast incremental MH into dev when it is ready.

iffsid:
Sure.
Incremental MH is implemented as a separate inference algorithm in the `incremental-mh` branch, so merging (the git operation, not conceptually) into dev shouldn't cause issues. Most of the changes in this pull-request match those implemented as hashMH in the `incremental-mh` branch.

iffsid:
@dritchie just pushed hashmh into dev as part of the merge of incremental-mh. Is there any reason to have the old (inefficient) implementation of mh around instead of having one of these replace it?
If we're planning on replacing old mh with the hashmh-type implementation, then we should be able to just pull this commit in without trampling on any of @dritchie 's changes.

stuhlmueller:
Given that we now already have hashmh and the store handling fix, what extra things does this pull request do? Maybe make new (smaller) pull requests on top off the current dev head so that we can have those extra things for hashmh? (In general, it's much easier to merge smaller pull requests.)

iffsid:
The pull request makes hashmh the default mh implementation. It also makes hashmh more modular so that it can be used within pfRejuv, which buys quite a decent amount, particularly for models that have a reasonably large number of `sample` statements in a trace. Those are the only changes. The only reason why smc.js was touched in this commit was to move the old inefficient mh helper functions into smc.js so as to keep the tests valid.

stuhlmueller:
Ok. Make a pull request against head just for the refactoring for now, and close this one?

Issue #94, opened 2015-05-04
Refactor tests: run all inference algorithms on a core set of models


stuhlmueller:
This will help us check that inference implementations don't forget handle some basic constructs (like structure change).

null-a:
I've made a start on this. The basic approach is to split apart the model definitions and inference procedures [like so](https://github.com/null-a/webppl/commit/e1a4e159e67e24690f278086b5299d6ab0c41d31#diff-e0a743b6daf0785c82a7f8635fffc632R10).

One problem with this is that I'll end up duplicating the models currently in `examples`, though not having the test suite depend in stuff in `examples` might be a good thing.

Feedback welcome!

null-a:
I've done a little more on this, it's currently on a branch in my fork:

https://github.com/null-a/webppl/commits/refactor-inference-tests

null-a:
This now includes all of the tests currently specified by `tests/test-data/models/*.wppl`.

The new test runner allows us to specify models, inference algorithms, and which combinations of the two to test. Adding additional models which we test all inference algorithms against should be straight-forward.

The intention is that it's also possible to add tests of the type mentioned in #66. I've added [testing of the normalization constant](https://github.com/probmods/webppl/commit/69a5b319ad50b28aba91bf33c4842c9ec3aa5caa) as an example.

stuhlmueller:
Looks good to me. I don't think it's a problem that we're decoupling examples and tests again, even if it leads to some duplication for now. It would be good to still run all the examples as part of the CI tests, even if we don't check their return distribution, to at least verify that changes don't break any of the examples in obvious ways.

PR #107, opened 2015-05-27
MHdrift

pedir:
This adds drifting proposals to MH. Also adds burn-in, graphing of the trace, gweke test, and refactors SMC to use MH methods when possible.


pedir:
The travis-ci failure seems to be just an npm warning not related to these changes. The tests and lints are all passing.

stuhlmueller:
Agree that the travis failure looks unrelated (would be good to fix it, though).

Can you add the equivalent of testDriftLinearRegression.wppl but make it use SMC? (Just to verify that drift proposals also work as part of SMC.)

pedir:
I fixed the npm warning on dev so that shouldn't be an issue anymore.

I also added a test for drift PFRj very similar to the linear regression one for the MHDrift. They will hopefully be consolidated when we get around issue #94.

PR #113, opened 2015-06-15
HashMH Store Handling Fix

null-a:
HashMH uses something like rejection sampling to find an initial trace with a non-zero probability. The store isn't reset across these executions leading to incorrect results.

I discovered this when I added store handling inference tests for HashMH as part of #94.

This change fixes the problem and doesn't appear to break things.


iffsid:
This is also corrected in #99 -- mh is essentially replaced by hashmh with factored-out code.

Issue #66, opened 2015-03-24
Make test suite check expected values of continuous examples


stuhlmueller:

iffsid:
We could also add other such checks to the tests -- e.g., the normalization constant from particle filters.

stuhlmueller:
Good idea, should do that.

stuhlmueller:
This is partially implemented now (for drift ERP test cases), but should be extended to more models and inference algorithms.

Issue #96, opened 2015-05-05
Add rejection sampling


stuhlmueller:
Requires an upper bound on the total factor score per execution (easiest is 0, with restriction to negative factor values).

null-a:
Have I understood this correctly:

It's tricky to reject samples early without the restriction that factor values are all negative. With such a restriction, once `scoreSoFar` drops below the `logUniform + maxScore` threshold no future `factor` statements can rescue the execution. When positive factor values are allowed, it's hard to say anything useful about what the rest of the program will do, so we'd have to run it to the end before deciding whether to accept or reject.

I've played with the current code a little. I noticed that we need to `JSON.stringify` the properties of `hist`, but other than that it seemed to be working OK. What's broken about it?

stuhlmueller:
As I understand it, for rejection to work, you need an upper bound on the total achievable factor score (`maxScore` in my code; this can be 0 when all factors are negative). This allows you to reject traces based on how far below `maxScore` their score is. The simplest way to implement this is to do a single big rejection step at the end. I tried to be clever and do early rejection, but I think my math is broken and so the resulting distributions aren't right in general. (I haven't spent much time on this -- let me know if I should look into this and fix it.)

PR #114, opened 2015-06-16
Refactored Inference Tests

null-a:
I think this is ready to merge.

The main change is that models and expected results are now defined independently of inference algorithms and there's a fairly simple way to say which combinations of the two to test. This either addresses #94 or makes it much easier to do so.

I've added a [new test](https://github.com/probmods/webppl/commit/4f78e7cafad3a2722ee3509eb7e3715dabfa0813) to check the expectations of continuous models across more algorithms and the [normalization constant](https://github.com/probmods/webppl/commit/69a5b319ad50b28aba91bf33c4842c9ec3aa5caa) of particle filter algorithms. (#66)

The test suite still runs the programs in `examples` but the expected results of these are no longer checked. I'm happy to turn models into new tests where we think that's useful.

I've added tests for `AsyncPF`, `IncrementalMH` and `HashMH`. One of the `AsyncPF` tests fails intermittently with an exception, I'll create an issue for that.

Other than that we're running the same tests as before.


stuhlmueller:
Looks great! 

I like the way you're handling the models in the examples folder now (checking that the return value is an erp, even if we don't check the distribution).

On the test group specifications, for example this one:

    {
      name: 'ParticleFilter',
      args: [1000],
      only: ['simple', 'store', 'gaussianMean', 'varFactors1', 'varFactors2'],
      hist: { tol: 0.1 },
      logZ: { check: true, tol: 0.05 },
      store: { hist: { tol: 0 }, args: [100] },
      gaussianMean: { mean: { tol: 0.3 }, std: { tol: 0.3 }, args: [10000] },
      varFactors1: { args: [5000] }
    }

There are two kinds of entries in this object, right? Ones that apply to all tests within the group (`name`, `args`, `only`, `hist`, `logZ`) and then test-specific ones (`store`, `gaussianMean`, `varFactors1`)? If that's true, then I wonder whether it would be clearer to separate these two levels in the spec, e.g. like this:

    {
      name: 'ParticleFilter',
      args: [1000],
      hist: { tol: 0.1 },
      logZ: { check: true, tol: 0.05 },
      tests: {
        simple: {},
        store: { hist: { tol: 0 }, args: [100] },
        gaussianMean: { mean: { tol: 0.3 }, std: { tol: 0.3 }, args: [10000] },
        varFactors2: {},
        varFactors1: { args: [5000] }
      }      
    }

Or like this:

    {
      name: 'ParticleFilter',
      args: [1000],
      only: ['simple', 'store', 'gaussianMean', 'varFactors1', 'varFactors2'],
      hist: { tol: 0.1 },
      logZ: { check: true, tol: 0.05 },
      settings: {
        store: { hist: { tol: 0 }, args: [100] },
        gaussianMean: { mean: { tol: 0.3 }, std: { tol: 0.3 }, args: [10000] },
        varFactors1: { args: [5000] }
      }
    }

I also wonder whether we should merge `test-sampling.js` with the inference tests. This would result in a bunch of small files, but could reduce code duplication. What do you think? (This could be a separate pull request.)

null-a:
> Looks great!

Thanks!

> There are two kinds of entries in this object, right?

Yeah, that's right.

The stuff you're suggesting should go under `settings` or `tests` contains model specific arguments to be passed either to the inference algorithm (`args`) or to individual tests performed on the ERP returned from inference. e.g. `hist` contains args passed to the check which calls `histsApproximatelyEqual`. If there are no model specific settings then the settings at the top level (i.e. the object with the `name` property) are used.

I've been un-easy about exactly the point you make, and since you've picked up on it straight away then I agree it should be changed. I'll take a look and add commits to this PR.

I'll also take a look at pulling in `test-sampling.js`. I had thought they were best kept separate, but looking again merging them doesn't seem unreasonable.

null-a:
I've made a couple of changes.

1. I think the inference test definitions are more readable now, I'm interested to hear what you think!

2. I've merged `test-sampling.js` into `test-inference.js`. The tests are very slightly different as the code under test is now wrapped in a function and an ERP is created out of the results. This is so that the inference and sampling tests have a common interface. In essence we're still testing the same thing I think. I've re-used the `geometric` model from the inference tests for the sampling test to save duplicating two very similar models.

I think it would have been cleaner to implement this by adding a new `ForwardSampling` inference algorithm, something [like this](https://github.com/null-a/webppl/commit/b44a8f471849e683c2c35ce8db80d7464cad70bc). I didn't go this way in the end because it's not quite testing the same thing as the current tests[1] and I wasn't sure it was a great idea having an extra inference function hanging around which is only used for testing. I thought I'd mention it though, just in case you think it's a better approach.

[1] Multiple samples are generated within a single webppl program, rather than by running the webppl program multiple times.

stuhlmueller:
> I think the inference test definitions are more readable now, I'm interested to hear what you think!

Looks good.

> I think it would have been cleaner to implement this by adding a new ForwardSampling inference algorithm, something like this. I didn't go this way in the end because it's not quite testing the same thing as the current tests[1] and I wasn't sure it was a great idea having an extra inference function hanging around which is only used for testing.

This seems better to me. I don't think the difference in what we are testing matters much (either way will check whether our sampling distributions are what they should be). 

In principle, you don't even need a special forward sampling procedure – without factors, the current particle filtering implementation should reduce to forward sampling. However, it doesn't seem great for this use case - there is too much additional structure that could in principle introduce errors. Another possibility would be to finish the rejection sampling implementation that I started a while ago (see `rejection` branch; currently broken), which without factors reduces to forward sampling in a more straightforward way. You could also use your forward sampling implementation for now and replace it with rejection when we add it.

null-a:
> You could also use your forward sampling implementation for now and replace it with rejection when we add it.

I like it. I've updated the tests to use `ForwardSample`, I'll change them over to `Rejection` once #96 is in.

I think this is ready to merge, unless you can think of anything else?

PR #121, opened 2015-06-22
Rejection Sampling

null-a:
This now defaults to rejecting at the end of an execution, but can optionally be set to reject incrementally. (#96)

As well as requiring an upper bound on the total factor score, incremental rejection requires each individual factor score to be <= 0. (For those factor statements we see, this is checked at run time.)

I think this is correct, do you agree?

(I've also replaced `ForwardSampler` with `Rejection` in our inference tests.)


stuhlmueller:
Haven't had a chance to look at it in detail yet, but in the meantime: add a test case or two that exercise rejection with non-zero bounds on the total score?

null-a:
Sorry for going quiet on this.

> add a test case or two that exercise rejection with non-zero bounds on the total score

Sure, good call - I'll take care of it next week.

stuhlmueller:
Thanks for adding the `upweight` test! It would be good to add another test with negative non-zero bound, so that we have confidence that incremental rejection also does the right thing in general. For that test case, I'd use factor scores that don't differ quite as much between executions, so that we can test that incremental rejection actually gives us the right relative probabilities of different executions (and doesn't just work in the setting where one execution is ruled out entirely). It's fine if the test case is a bit more complex - we can compute the ground truth using enumeration, after all.

null-a:
> It would be good to add another test with negative non-zero bound

Done!

Issue #38, opened 2015-01-23
Enable debugger; statements in webppl code


stuhlmueller:
Pass through `debugger;` statements to compiled code. This requires that cps deals with nodes of type `DebuggerStatement`.

null-a:
This is a little tricky. It looks like the cps transform turns *everything* into an expression making it difficult to include statements such as `debugger`.

One way to make this work is to side-step the problem by wrapping `debugger` statements in a function. After compilation, a `debugger` statement then becomes something like:

```
var _dummy1 = (function (globalStore) { debugger; })(globalStore);
```

This isn't ideal, but it works. (Code [here](https://github.com/probmods/webppl/compare/dev...null-a:debugger).)

I can think of 2 alternatives/improvements:

1. Figure out how to include statements in cps directly. It's not clear to me what this would involve.
2. Have a post-processing step which takes the current output and removes the redundant assignment and wrapper function.

stuhlmueller:
> One way to make this work is to side-step the problem by wrapping debugger statements in a function. 

What are the drawbacks of this approach? A single extra function call on the stack? Does this lead to some environments becoming inaccessible from the debugger, or is it just an extra click?

null-a:
I think user unfriendlyness was my concern, but you're right it's just  
an extra call on the stack and an extra click.

PR #126, opened 2015-06-30
Enable debugger statements

null-a:
This enables `debugger` statements in webppl code.

Since the earlier discussion in #38 I've included a post-processing step to remove the immediately invoked wrapping function.

(`debugger` statements which appear as the final statement in a block will still be wrapped but that's fine.)

I think this is good to merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,16 +83,13 @@ To debug WebPPL programs running in Chrome, enable [pause on JavaScript exceptio
 
     // 1. Install node-inspector (only need to do this once)
     npm install -g node-inspector
-    
-    // 2. Compile your webppl program to Javascript
-    webppl my-program.wppl --compile --out my-program.js
-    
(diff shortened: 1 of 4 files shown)


stuhlmueller:
Do `debugger;` statements within models actually stop the computation for you at the right point? I just tried it, and for me it stopped at the eval statement:

screenshot 2015-07-02 16 02 25

I don't know if this is caused by your change, but if this isn't just happening for me, it would be important to fix for debugging to be useful.

null-a:
> Do debugger; statements within models actually stop the computation for you at the right point?

Yes. I think upgrading node-inspector will fix this for you. I'm seeing the same thing as you with versions prior to v0.10.1.

stuhlmueller:
Updated node-inspector, now it works. Neat!

Issue #122, opened 2015-06-23
--require-header command line arg isn't fully compatible with internal headers


stuhlmueller:
Internal headers are functions that are called with env:

    headerModule = require('pathToHeader.js')(env)

This allows access to `env.coroutine`, among other things. Command line arg headers currently don't have this access.

PR #127, opened 2015-06-30
Pass `env` to additional headers

null-a:
This addresses #122.

Obviously this breaks things for anyone already using the current set-up. We could think about backwards compatibility or informative error messages if you think this will be an issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,18 +146,25 @@ Using the example of reading and writing CSV files:
 
 ### Additional header files
 
-Sometimes, it is useful to define external functions that are able to access the store, continuation, and address arguments that are present at any point in a webppl program but usually not exposed to the user. Let's use the example of a function that makes the current address available in WebPPL:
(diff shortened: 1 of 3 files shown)


stuhlmueller:
Looks good! I don't think backwards compatibility matters in this case (not sure anyone who isn't working on a project that I'm involved in has ever used this feature).

hawkrobe:
I wanted to report two issues I've been experiencing with the requireHeader function:

1. After browsifying webppl, it fails to load in the browser with a ```Cannot find module './header.js'``` error. This was resolved by commenting out ```requireHeader()``` function and changing the require statement for header.js back to ```require('header.js')```

2. The [issue with dippl and forest in Safari and Firefox](https://github.com/forestdb/forestdb.org/issues/6) appears to trace back to this function.

I don't understand the issue well enough to diagnose the actual problem -- I just wanted to report these observations!

null-a:
> I wanted to report two issues I've been experiencing with the requireHeader function

Thanks, I'll take a look!

Issue #123, opened 2015-06-25
Package/module system


stuhlmueller:
It would be useful to have a standardized way of importing webppl extensions such as coarse-to-fine inference and visualization tools (#88). Here are some thoughts on what this could look like:

A package can provide each of the three `require-*` types supported by the webppl command line tool: headers (with functions that take store, continuation, address as args), external JavaScript files (that don't take store, continuation and address as args), and webppl files. 

A package could simply be a node.js package where the package.json file contains additional information about where these files are located, like this:

    {
      "name": "webppl-viz",
      "description": "Visualization tools for webppl",
      ... (more standard package.json entries) ...
      "webppl": {
        headers: ['src/viz-header.js'],
        external: ['src/viz-base.js', 'src/viz-scatter.js'],
        wppl: ['src/viz.wppl'],
      }
    }

The intended meaning of importing the package is to include these external files as if we had run webppl with the corresponding `require-*` arguments.

Installation could happen via npm: `npm install -g webppl-viz`.

Importing packages could happen either via special primitives in the source code of webppl programs (e.g., `import('viz')` imports the `webppl-viz` package) and/or on the command line (`webppl foo.wppl --require viz` imports the package).

Issue #34, opened 2015-01-15
Get eval working


longouyang:

stuhlmueller:
To do this, we need a header function that applies all transforms to the string of interest, uses Javascript eval to evaluate the transformed string, and then passes the returned object to the continuation.

stuhlmueller:
In fact, there's already `webppl_eval` [here](/probmods/webppl/blob/dev/src/main.js#L99), but it's currently not quite appropriate for this purpose (e.g. doesn't take store).

null-a:
Does #157 close this?

feste:
Nope, it still doesn't take a store I think.

stuhlmueller:
Hmm. My first thought was that this should be closed (since `webpplEval` now takes `s, k, a, code` as arguments, so the first argument is the store). However, `run` evaluates the code with an empty store:

    eval.call(global, compiledCode)({}, k, '');

So, `webpplEval` as it is will empty the store for all subsequent computation. Similarly, the address is reset. This is a bug. To fix this, `run` could take store and address as optional arguments and pass them on to the result of the eval call. However, it may be simpler to just call `compile` and `eval` directly in `webpplEval`, and pass on store, address, and continuation there.

null-a:
Closed by #177.

PR #133, opened 2015-07-02
[WIP] Import webppl packages

hawkrobe:
This implements the spec for a package/module system proposed in #123. 

I've provided a functioning example in examples/webppl-module. A package can be required when running webppl code from the command-line by calling, e.g.,

```webppl model.wppl --require path/to/module/package.json```

and can be bundled into the webppl.min.js used in the browser by calling the ```browserify-with-package.js``` script like so:

```node browserify-with-package.js /path/to/module```

The command-line --require flag could replace the trio of require-header, require-js, and require-wppl if we chose, but I left them in for now. 

I'd like to note that the browserify bundling is a hack at best: essentially, the script uses the package.json file to generate an ```additionalReqs.js``` file which appends the webppl headers and pre-imports the external javascript modules in a way that makes browserify happy. Then ```additionalReqs.js``` is imported at the top of ```main.js,``` and you get global access to these functions from there. I'd be open to other ways of solving this, but browserify is quite constraining. 

Also note that while the command-line interface handles all three types of headers, ```browserify-with-package.js``` script does not currently handle the fancy ones with access to internals. This doesn't seem as important, but it's something to keep an eye on in the future!

### CODE REVIEW

**Comment by hawkrobe** [comment]
@stuhlmueller @null-a I've finished everything I set out to do with this, and I'd appreciate a code review before merging! 

As I said above, I think the command-line interface is fairly simple (although it might be more robust to how you supply the path -- you currently need to point it directly to the package.json file). The messy bit is the browserify bundling process!

Thanks!

**Comment by stuhlmueller** [comment]
I haven't looked at the code yet, but wouldn't it be easy to just require the path to the module directory to be specified, not the path to package.json?

Also, if we're imagining that modules get installed via "npm install moduleName" (or "npm -g install moduleName"), wouldn't it be sufficient to just require the module name, and then let node figure out the path using the same mechanism it uses for importing ordinary node modules?

**Comment by hawkrobe** [comment]
Good points!

The problem with the first is the ```getRequirePath()``` function in the 
command-line webppl script -- it will throw a "cannot resolve module" 
error if you pass in a path to a directory instead of a file. We could 
change the logic of it to check whether the string you've passed in ends 
with a "/" and append "package.json" to it for the check? We can't just 
bypass the function, because it may be a relative path that needs resolving.

And I think the second is doable, but seems a bit harder (i.e. requiring 
someone to read a bunch of npm documentation) and I think it would 
constitute a discrete unit of work.

Andreas Stuhlmüller wrote:
>
> I haven't looked at the code yet, but wouldn't it be easy to just 
> require the path to the module directory to be specified, not the path 
> to package.json?
>
> Also, if we're imagining that modules get installed via "npm install 
> moduleName" (or "npm -g install moduleName"), wouldn't it be 
> sufficient to just require the module name, and then let node figure 
> out the path using the same mechanism it uses for importing ordinary 
> node modules?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/probmods/webppl/pull/133#issuecomment-118195447>.
>

-- 
Sent with Postbox <http://www.getpostbox.com>

**Comment by null-a** [comment]
I've now refactored this. The basic approach is the same, here's what's changed:

* You no longer include `package.json` in the path given on the command line, instead do something like: `./webppl prog.wppl --require examples/webppl-module`.
* The regular mechanism for specifying package entry points is used instead of the `webppl.external` entry in `package.json`. i.e. include an `index.js` file or specify a `main` entry in `package.json`. This makes all regular node packages valid webppl packages. For example, this works: `./webppl prog.wppl --require node_modules/underscore`.
* `src/browser.js` is now the entry point for `browserify`.
* Packages are now included in the browser bundle by a browserify plugin. Use it like so: `browserify -t [./src/bundle.js --require node_modules/underscore --require examples/webppl-module] -t brfs src/browser.js > webppl.js`
* The number of bundled packages in shown in the console alongside the `webppl loaded` message.
* I've removed the command line options to include individual header/js/wppl files. Packages are so easy to create that it doesn't seem worth the effort to keep it around long-term. Let me know if you think otherwise.

One thing I need to give some thought to is what should happen if you pass a package name without a path as a `--require` argument. If we just pass this to `require` then I *think* node will start looking in `./webppl`'s `node_modules` directory and walk up the directory tree. I don't think it will look at packages installed with `npm install -g`. This doesn't seem very useful for us.

Another thing that might be worth is thinking about is how we approach testing packages. We'll presumably want to do this (e.g if we split out inference algorithms) but I'm not sure how we'd go about it.

**Comment by null-a** [comment]
### Todo

- [ ] Make sure dippl still works
- [ ] Update `README`

**Comment by stuhlmueller** [comment]
Nice. Thanks for writing the browserify plugin!

> The regular mechanism for specifying package entry points is used instead of the `webppl.external` entry in `package.json`. i.e. include an `index.js` file or specify a `main` entry in `package.json`.

Good idea!

> The number of bundled packages in shown in the console alongside the `webppl loaded` message.

We could show the names of included packages, too.

> I've removed the command line options to include individual header/js/wppl files. Packages are so easy to create that it doesn't seem worth the effort to keep it around long-term. Let me know if you think otherwise.

Seems fine. I'll think about this again once you update the README with a description of the new approach, and once I've played with the package system a bit.

> One thing I need to give some thought to is what should happen if you pass a package name without a path as a `--require` argument. If we just pass this to `require` then I *think* node will start looking in `./webppl`'s `node_modules` directory and walk up the directory tree. I don't think it will look at packages installed with `npm install -g`. This doesn't seem very useful for us.

Yes, it looks like node's `require` doesn't search globally installed packages. Since globally installed packages are mainly for command line scripts (like core webppl), we might not want to change this behavior.

Let's think about how we expect people to use our package system. One use case goes like this:

1. (I install webppl using `npm install -g webppl`)
2. I create a `my-webppl-project/` folder and run `npm init` in this folder
3. I locally install any packages I want to use for this project, e.g. `npm install webppl-viz`
4. I create my webppl file `my-webppl-project/my-model.wppl`
5. I run `webppl my-webppl-project/my-model.wppl`

In this case, I'd want webppl to search in `my-webppl-project/node_modules` even if I'm executing `webppl` from another directory.

In another use case, step 3 is replaced with a global package installation. If we don't want to store those packages in the global npm package directory, we could give special meaning to `~/.webppl/` in the user's home directory, require the user to install packages there, and always search this path for packages:

    npm install --prefix ~/.webppl webppl-viz

Based on these use cases, first searching for packages by looking for the next-higher `package.json` based on the model location, then searching the global `/.webppl/` folder seems like a reasonable strategy to me. What do you think?

> Another thing that it might be worth us thinking about is how we approach testing packages. We'll presumably want to do this (e.g if we split out inference algorithms) but I'm not sure how we'd go about it.

Agreed. This should follow existing node testing practices (run `npm test` to run the tests, which are stored in a `tests/` folder). Is the difficulty that we can't easily `require` the global webppl location in general? For travis-ci, we can simply `npm install` it locally when we run the tests.

The current pull request isn't handling the general case of recursive package dependencies yet, right?

**Comment by null-a** [comment]
> Based on these use cases, first searching for packages by looking for the next-higher package.json based on the model location, then searching the global /.webppl/ folder seems like a reasonable strategy to me. What do you think?

Sounds good to me! I've implemented it and started to describe how it works in the `README`. I think it's at a point where you could have a play. `webppl`'s `--verbose` flag shows a little bit of information from the package look-up routine which might be useful if you hit any bugs.

Things work a similar way with `browserify` except that we only look in `~/.webppl`.

Packages can also be specified by path in both cases, see the `README` for info on that.

I've tried to make this all work cross-platform but I've only tried it on OS X.

> Is the difficulty that we can't easily require the global webppl location in general?

I was trying to figure out how someone would test an inference algorithm which they had in a package without reimplementing all the testing infrastructure we have in the core library. Perhaps we worry about things like this later, when we actually need them.

> The current pull request isn't handling the general case of recursive package dependencies yet, right?

That's right.

**Comment by stuhlmueller** [comment]
I tried it -- seems to work pretty well. ([Here's the package](https://www.npmjs.com/package/webppl-isosmc) I used to experiment.)

Two small issues I ran into:

First, in my module folder, I tried to run `webppl --require . examples/trivial.wppl` to give it a try, but this doesn't work:

    /Users/stuhlmueller/Projects/webppl/src/pkg.js:53
          throw 'Could not find WebPPL package: ' + name;
                                                  ^
    Could not find WebPPL package: .

Would it be difficult to make this work?

Second, for globally installed packages (i.e. packages in `~/.webppl/`), we should search `~/.webppl/node_modules/`. This is where `npm install --prefix ~/.webppl <package-name>` puts packages.

> I was trying to figure out how someone would test an inference algorithm which they had in a package without reimplementing all the testing infrastructure we have in the core library. Perhaps we worry about things like this later, when we actually need them.

I think it's fine to postpone this. Naively, it seems like we could just export the relevant functions from webppl (possibly moving some of the infrastructure from `tests/` to `src/`), but maybe I'm missing something.

**Comment by null-a** [comment]
Paths have to begin with `/`, `./` or `../`, a convention I [borrowed](https://nodejs.org/api/modules.html#modules_all_together) from Node's `require`. This should work:

```
webppl --require ./ examples/trivial.wppl 
```

That rule seems overly strict. It seems safe to relax it such that we consider `.` to be a path also, so I'll change it.

> we should search ~/.webppl/node_modules/. 

Oops! I'll fix that too.

erindb:
I used `browserify` with a [`webppl-viz`](https://github.com/erindb/webppl-viz) package and it was awesome! :) Thanks!

null-a:
When I updated the README I took out the detailed CSV example as I found it tricky to get the basics of how packaging works across at the same time. I'm happy to have another attempt if retaining that specific example is important.

stuhlmueller:
We use this functionality pretty frequently, but there's not reason for the code itself to be part of the README. How about we create a `webppl-csv` npm package (as part of #146) and, once we have that, you add a brief note to your new README section on install remote packages, using that as an example?

null-a:
I've looked over this and I don't see any reason why it couldn't be merged.

Before we do, a quick question: Is the distinction between `run` and `webpplEval` in `main.js` important? I guess `run` (which is called from the browser) has access to the variables in the `main` module and that `webpplEval` doesn't, but it's not clear to me why that's useful. Also, we add `webpplEval` to the global scope in node, and there's a comment asking whether we should do the same for the browser. Should we?

If you don't want to think about this now, I'll happily move this to a fresh issue.

stuhlmueller:
> Before we do, a quick question: Is the distinction between `run` and `webpplEval` in `main.js` important?

At some point, `webpplEval` was used to provide `eval` functionality within webppl code. I don't know what the current state is of that - as you point out, it's not available in the browser at the moment. So, my hunch would be to merge `run` and `webpplEval` for the moment, and to add something like this back in as part of #34 if necessary.

null-a:
> merge run and webpplEval for the moment

Done!

I've done a very basic check to ensure things still work in the browser and I'll test it more tomorrow, but if this looks OK to you do you want to merge it?

stuhlmueller:
Looks good. One minor nitpick:

> The name of the variable is derived from the package name by replacing all occurrences of `-` with `_`.

It would be more Javascript-y if we turned the package name into camelcase, e.g. transforming `my-package` to `myPackage`.

null-a:
> It would be more Javascript-y if we turned the package name into camelcase

Good catch! Done.

PR #137, opened 2015-07-02
fix to MAP estimating in marginalizeERP

iffsid:


stuhlmueller:
Thanks! We should have tests so that we catch bugs like this in the future: #138

Issue #135, opened 2015-07-02
Enable easy use of IncrementalMH


dritchie:
Incremental MH needs a couple of additional source transforms. These add unnecessary overhead for other inference algorithms, so they're only performed if a flag is passed to webppl.compile. To make this easier to use, it'd be nice to detect whether the program being compiled uses Incremental MH and turn the flag on if it does. One simple approach would be to scan the (pre-header-inclusion) AST for callees named 'IncrementalMH.' That's not 100% correct, since the user could always alias IncrementalMH to something else, though. A more conservative approach might be to look for a (non-comment) token 'IncrementalMH' anywhere in the program.

Issue #64, opened 2015-03-24
Factor out repeated code in inference algorithms


stuhlmueller:
There is a lot of overlap in pf, pmcmc, smc; and in mh and smc. One way to combat this would be to just have a single smc algorithm, and implement all of the  algorithms mentioned above as parameterizations of this uber algorithm. This has some advantages, but doesn't feel entirely right, based on what smc currently looks like; it would be better if we had smaller compositional pieces.

stuhlmueller:
Also, instrument inference algorithms with hooks such that we can pass in functions to collect and compute diagnostics without the diagnostics code being part of the core algorithms.

null-a:
I've made an [initial attempt](https://github.com/probmods/webppl/compare/dev...null-a:refactor-trace) to refactor mh and smc. The basic idea is to structure things around functions which operate on traces. For example, there's a new [`MHKernel`](https://github.com/null-a/webppl/blob/refactor-trace/src/inference/mhkernel.js#L47) function which takes a webppl program and a trace from a previous execution and returns a new trace using MH. With this in hand you can do [MCMC](https://github.com/null-a/webppl/blob/refactor-trace/src/inference/mh2.js#L15) with a fairly simple iteration which accumulates results in a histogram, and we can do [rejuvenation](https://github.com/null-a/webppl/blob/refactor-trace/src/inference/particlefilter2.js#L172) in smc without duplicating any of the MH propose/accept/reject logic.

One thing I'm keeping in mind is Noah's [suggested interface](https://github.com/probmods/webppl/issues/86#issuecomment-94001674) for SMC. This gets us part of the way there, although it's not clear to me how much work it would take to make it possible to plug in other kernels.

This is a very rough sketch (lots of rough edges, I'm not certain how well it extends to the other algorithms, I don't know if it incurs a performance hit, etc.), but do you think it's heading in a useful direction?

null-a:
My [initial attempt](https://github.com/probmods/webppl/compare/dev...null-a:refactor-trace) at refactoring some of the inference algorithms (more detail in previous comment) factors out the (rejection) strategy used to initialize the Markov chain. At present this *increases* the amount of code (because the initialization strategy is a new coroutine) but it seems possible that going this way yields further opportunities to simplify things that will leave us with less code once we're finished.

The idea of "make traces more central" that was mentioned in the planning meeting sounds superficially similar to this approach. I'll investigate further and report back once I've understood incrementalization.

null-a:
[...]

PR #142, opened 2015-07-03
Enable caching transforms based on simple AST test.

null-a:
This is an initial attempt at #135. The transforms are enabled for any program which includes the identifier `IncrementalMH`.


stuhlmueller:
Since this transforms the entire program, it would be good to test that the caching transform doesn't break any of the other inference algorithms. We could have a version of the `simple.wppl` test case that includes a `IncrementalMH` query that does (almost) nothing.

null-a:
That's a great suggestion. I'm pretty sure a few such tests would fail at the moment as a few of the inference coroutines don't define an `incrementalize` method. I assume they should all do something like `Coroutine.prototype.incrementalize = env.defaultCoroutine.incrementalize`. I'll add the tests, fix any failures and add a commit to this PR.

Another thought: At present if you run incremental MH without the transforms there's no indication that anything is wrong. (We've had passing tests for incremental MH *without* the transforms since I refactored the inference tests.) I think we could do with a test or runtime error/warning to catch this, I'm just not sure how best to tackle it yet.

dritchie:
"At present if you run incremental MH without the transforms there's no 
indication that anything is wrong."
Just to be clear: the transforms help IncrementalMH be more efficient, 
but they're not actually necessary for it to be a correct inference 
algorithm.

Adding |Coroutine.prototype.incrementalize| to all the inference 
coroutines is what I did in the incremental-mh branch, but it still rubs 
me the wrong way. It feels wrong and non-modular for every coroutine to 
define an extra method just because of a feature that one of them 
happens to have. I don't have a better suggestion at the moment, though.

-- D

null-a:
>  the transforms help IncrementalMH be more efficient, but they're not actually necessary for it to be a correct inference algorithm.

Since the tests were passing that was my assumption, thanks for confirming!

>  It feels wrong and non-modular for every coroutine to define an extra method

I totally agree. In the absence of a straight-forward solution though, I think it's reasonable to live with it temporarily and clean it up as part of #64.

Issue #115, opened 2015-06-16
incorrect acceptance probability using MH with custom proposer 


hawkrobe:
Suppose we have the following ERP generator, which assigns uniform probability to all integers between 1 and N. It has a custom proposer that suggests a very small subset of those integers with uniform probability. (Yes, this is a ridiculous ERP, but the same issue arises in more useful, complex cases).

```
var makeERP = function(maxNum) {                                   
  var mySampler = function(){                                   
        return Math.floor(Math.random * maxNum);                   
  };                                                             
  var myScorer = function(params, val){                         
        return Math.log(1/maxNum);                                 
  };                                                             
  var myProposer = new ERP(                                      
        function sample(params) {                                  
            var intervalWidth = Math.floor(maxNum / 10);           
            return Math.floor(Math.random() * 10 * intervalWidth); 
        },                                                         
        function score(params, proposedVal) {                      
            return Math.log(1/10);                                 
        }                                                          
  );                                                             
  var erp = new ERP(mySampler, myScorer,                       
                      {proposer: myProposer});                     
  return erp;
};

module.exports = {                                                                      
    makeERP : makeERP                                                                   
}      
```                                                                                 

Suppose we use this in a very simple model, like the following:

```
var myERP = customProposer.makeERP(10000);
                                        
var model = function() {                
    var mat = sample(myERP);            
    return mat;                         
};                                      
                                        
MH(model, 10, true);      
```

We expect that it would uniformly sample 10 values from the proposer distribution, since all samples have the same score under the base ERP, and the proposer is symmetric.

When we run this at the command line, we see that after the first sample, all proposals are rejected. When we print out the details of mh.acceptProb(), we see that it's using the custom proposer's scorer to compute the forward probability but the original ERP's scorer to compute the backward probability. Because the forward probability is so much larger, the acceptance probability is about 0.0001 (and this obviously gets smaller as the maxValue we use increases).

This seems to be a bug in the integration of custom proposers into MH -- both the forward and backward probabilities should use the custom proposer scores.

stuhlmueller:
First, two minor comments:

There is a small bug in the example: Math.random should be Math.random():

    var mySampler = function(){                                   
      return Math.floor(Math.random() * maxNum);                   
    };

Also, the third parameter in the MH call is now burn-in, so let's change the call:

    MH(model, 10, 0, true);

Then, let's adjust the custom proposer so that we can actually explore the full support of the ERP:

    var myProposer = new ERP(                                      
[...]

Issue #46, opened 2015-02-10
Add MH integration tests


stuhlmueller:
In particular, Geweke tests [as described here](http://arxiv.org/abs/1412.5218).

null-a:
Might this have caught #115? If not, what would - unit tests?

stuhlmueller:
Integration tests would have caught #115 if we had used a model with a custom proposer that doesn't assign the same forward/reverseChoiceScore to each ERP proposal. If we had such tests for models that exercise substantial parts of the system and if refactored our probability calculations to be clearer (#145), I would be a lot more confident that there aren't more bugs lurking in the MH algorithm.

PR #144, opened 2015-07-07
Fix custom proposer acceptance probability

stuhlmueller:
Issue #115. @ngoodman or @null-a, please review and merge.


hawkrobe:
Do you think we could put in the modified example from #115 as a test, to catch this kind of thing in the future? I don't think there's any test coverage of custom proposers right now, right?

stuhlmueller:
Yes, that's a good idea in principle. It needs a little bit of thought, since the current example takes long to converge, but at the same time, running it for very many iterations could have masked the current error (since it only happens at the first step).

null-a:
Looks good!

> I don't think there's any test coverage of custom proposers right now, right?

The `drift`inference test exercises custom proposers via `gaussianDriftERP` but it didn't show up the error.

> running it for very many iterations could have masked the current error (since it only happens at the first step).

I don't think it was limited to the first step in the `drift` model but it was still masked.

I've bundled the problem of improving testing in with #46 for now.

PR #152, opened 2015-07-26
Variable number of factors for ParticleFilterRejuv

iffsid:
This nominally sorts out #118.
When particles have different number of factors, at the resampling stage, the rejuvenation is only carried out on the still-`active` particles. There may be a case for adopting a different scheme where the `inactive` particles get rejuvenated again as well.
The final rejuvenation still works as it should, over all the particles.

### DIFF
--- a/src/inference/mh.js
+++ b/src/inference/mh.js
@@ -7,6 +7,7 @@ var _ = require('underscore');
 var assert = require('assert');
 var erp = require('../erp.js');
 var diagnostics = require('./mh-diagnostics/diagnostics.js')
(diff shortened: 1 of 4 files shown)


stuhlmueller:
Looks good.

One minor comment: Can we always use curly braces for conditionals?

For example, change

    if (i < this.particleIndex)
      this.resampleParticles(); // Updated all particles; now wrap around

to

    if (i < this.particleIndex) {
      this.resampleParticles(); // Updated all particles; now wrap around
    }

It's a bit clearer, and more consistent with the rest of our codebase.

PR #157, opened 2015-08-07
wrap run so that it can be used inside webppl programs

erindb:
useful for meaning functions in RSA models, especially for [counterfactuals](http://forestdb.org/models/tug-o-war-explanations.html).


null-a:
One tiny suggestion is that perhaps this could live in `headerUtils.js`?

erindb:
That does sound more reasonable, but is `run` available in `headerUtils.js`? (I couldn't tell if it was possible to require `main` from `headerUtils`...)

null-a:
Ah, you're probably right! You *could* `require('./main')` but it looks like that might not work because of circular `require` statements. We could change that but it might be more trouble than it's worth.

stuhlmueller:
Having it in `headerUtils` would in principle be better, but I agree that it's likely not worth the trouble.

It would be good to add a test case that uses `webpplEval` before this is merged.

erindb:
it seemed silly to run webpplEval through all the inference algorithms, since it's deterministic... so I just checked the result of running it on "1+1". does that seem OK?

stuhlmueller:
I was thinking that you could just test it using a single inference algorithm, such as rejection (which is probably the simplest), by specifying `func: 'Rejection',` in the test case. But making an extra folder for deterministic tests is good, too, perhaps even better. This will make it really easy to add more such tests, and they will be quick to run. 

Rename the `util-scripts` folder to `deterministic`, and `test-webppl-util-scripts.js` to `test-deterministic.js`? 

Also, is there some way to factor out duplicate structure in `test-inference.js` and `test-deterministic.js`? The `getModelNames`, `loadModel`, `loadExpected` and `generateTestCases` functions in the two files look very similar.

erindb:
we could combine `test-inference.js` and `test-deterministic.js` into a single `test-webppl.js` and just treat scripts in `test-data/deterministic/` differently from those in `test-data/models/`?

stuhlmueller:
That would raise the question why `test-examples` isn't subsumed by `test-webppl`, since the examples are also webppl scripts. Overall, having separate `test-deterministic` and `test-inference` files, with common structure factored out to `test-util`, seems slightly better to me, but feel free to take a different route.

erindb:
That sounds good. The name `test-util` is taken, I think, for tests having to do with `webppl`'s `util.js`. It's not obvious to me what would be a good way to factor out that code (where it should go such that grunt doesn't expect it to be running tests?). Also, lots of other test files have similar repeated code. Would it make sense to factor all of that out in this commit, too?

stuhlmueller:
Ah, you're right! It can't be `test-util` then.

We could have common structure live in `src/util.js` if it's not a lot of code, or in `tests/util.js` if it's more code (or `tests/helper.js`, `tests/common.js`, or something else; not sure what the best name is). We can be more specific in `Gruntfile.js` about where the tests live, so that `grunt` doesn't look in files that don't begin with `test-`.

It would be good to factor out common structure from other test files, too. If you don't want to do this now, we can also merge your current request (with the file renaming suggested above) and you can create an issue for refactoring the tests.

PR #158, opened 2015-08-11
Some extra functionality

iffsid:
Transferring some extraneous development in the `ad` branch


stuhlmueller:
Thanks for making a pull request!

In general, it would be better to split up different kinds of functionality into different pull requests, so that it's easier to merge parts.

Comments:

- I wrote `makeDeltaERP` at some point to speed up constant-valued ERPs. It looks like your new `categoricalScore` is faster than the old one (great!), but `categoricalSample` still calls `multinomialSample`, so maybe don't remove `makeDeltaERP` yet?
- I'm not sure that `liftERP` should be part of the main header. I don't have a strong opinion on this, so if you think it should be in there, we can have it in there. In that case, use more descriptive variable names? `be`, `bep`, `augS`, `liftedS`, `aggP` all require the reader to do some work.
- When is it a good idea to use the `noisyEq` function? Isn't it better to do `factor(gaussianERP.score([mu, sigma], v))`? It's surprising to see Gaussian density values flowing into a Bernoulli as a probability. For example, `flip(Math.exp(gaussianERP.score([1, .1], 1)))` would flip a coin with weight `3.989422804014326`.
- Also not sure that `stochasticCache` should be in main webppl (as opposed to a package), but don't have strong opinions here either. I guess the fact that we have `cache` in main makes it somewhat reasonable to have `stochasticCache` as well. At some point we'll do a cleanup and everything that doesn't seem like it should be in core will move to a package.
- In `getOpt`, check `optObject[option] !== undefined` instead of checking that `optObject[option]` is truthy? Otherwise, we will use the default value when the option value is `0`.

iffsid:
Thanks!

- I fixed up the naming (liftERP) and property checks (getOpt).
- `deltaERP` -- added in a specialization for the sampler in categoricalERP when the support is the singleton list.
- `noisyEq` -- I only added it in because I wanted an apples to apples comparison of all the diagnostics used in `bher` for the `ad` test programs. I so agree that the straight `factor` use is saner though; I just put this in for comparison purposes. But I take your point (6ab4352079e4bbb6df6e37cb748773f05ef6f6d5)
- `liftERP` and `stochasticCache` locations -- agreed, but I think these are the simplest places for them, for the time being at least.

stuhlmueller:
Thanks!

    for (var key in extraParams) {
      if (_.has(extraParams, key)) {

Use `_.each?`?

    console.log('there');
    console.log('here');

Remove these?

Add a test for `liftERP`, using the inference-tests infrastructure? If you can add one for `stochasticCache`, that would be good, too. These can be very basic tests, just so that we know that these functions don't break in obvious ways when we change webppl in the future.

> `noisyEq` -- I only added it in because I wanted an apples to apples comparison of all the diagnostics used in `bher` for the `ad` test programs. I so agree that the straight `factor` use is saner though; I just put this in for comparison purposes.

I don't think we need it in core webppl, even if it's useful for comparison.

iffsid:
Thanks; fixed
- `for..if` to `_.each`
- removed logging
- removed noisyEq

Also added simple tests for `liftERP` and `stochasticCache` (behaviour as `cache`).
A more rigourous test for `stochasticCache` needs a little more thinking. (You/I can open an issue related to it if necessary).

Issue #119, opened 2015-06-21
Unify treatment of importance distributions and custom MH proposers


stuhlmueller:
Currently, importance distributions can be added to erps within webppl, e.g. like this:

    var coin = withImportanceDist(FairCoin, makeCoin(.2))

By contrast, ERPs with custom MH proposers can currently only be constructed outside of webppl, e.g. like this:

    var gaussianDriftERP = new erp.ERP(
        erp.gaussianERP.sample,
        erp.gaussianERP.score,
        {proposer: gaussianProposer});

In the end, both approaches result in an ERP with an extra field (`importanceERP` and `proposer`, respectively), each of which has another ERP as its value. For importance sampling, this ERP is currently unparameterized, whereas for proposers, it takes the previous ERP value and the base ERP params as arguments.

It could be good to make the specification of proposers and importance distributions more analogous, maybe allowing both to be specified from within webppl.

null-a:
#494 switches importance distributions to use the guide program style, partly addressing this issue.

Does anyone have thoughts about how drift kernels should look using this style? The obvious thing seems to be to try something like:

```
sample(Gaussian(0,1), {
  guide: function(prev) { return Gaussian(prev, 0.1) };
})
// for which we can also provide a convenient wrapper:
gaussianDrift(0, 1, 0.1);
```

Calling this kernel function in cps from handwritten inference code will probably be a pain, but at first glance it seems like it would work.

If we *do* choose to go this way, it probably makes sense for me to skip #231 and #308 and jump straight to this?

dritchie:
I'm not a huge fan of the keyword `guide` being overloaded to mean different things in different contexts (i.e. importance dist vs. proposal dist), but otherwise I'm onboard with this approach.

In support of another (probably more long term) goal, it would be worth thinking about how to make this interface support more arbitrary proposals (i.e. compound stochastic functions beyond just sampling from a single distribution), while computing the correct acceptance probabilities / importance weights.

(There's also the as-of-yet-untouched question of how to support custom proposals to multiple variables, but that's a whole other can of worms, I think)

null-a:
> I'm not a huge fan of the keyword guide being overloaded to mean different things in different contexts 

Great, yes, using a different keyword seems like a sensible idea to me.

ngoodman:
@null-a i like your suggestion, with `guide` changed to `driftKernel` (or something like that). and yes, might as well go straight to this (skipping #231 and #308).

[...]

PR #161, opened 2015-08-12
Combine duplicated isErp functions.

null-a:
I need `isErp` in #64 so it would be nice to put it somewhere re-usable.

I considered putting it in `erp.js` rather than `util.js` but that felt a little odd because `isErp` becomes available at the top-level in webppl code but it isn't in CPS. I'm open to better suggestions if you have any?

I've checked this works from the command line with and without the `--compile` option.

A nice (though small) side-effect of this is I think we could use `util.isErp` in dippl [here](https://github.com/probmods/dippl/blob/15257640c73961cb7f38285347389ffa88f37e9d/assets/js/custom.js#L16).

### DIFF
--- a/src/erp.js
+++ b/src/erp.js
@@ -567,6 +567,14 @@ var makeMultiplexERP = function(vs, erps) {
   );
 };
 
+function isErp(x) {
+  return x && _.isFunction(x.score) && _.isFunction(x.sample);
+}
+
+function isErpWithSupport(x) {
+  return isErp(x) && _.isFunction(x.support);
+}
+
 module.exports = {
   ERP: ERP,
   bernoulliERP: bernoulliERP,
@@ -584,5 +592,7 @@ module.exports = {
   uniformERP: uniformERP,
   makeMarginalERP: makeMarginalERP,
   makeCategoricalERP: makeCategoricalERP,
-  makeMultiplexERP: makeMultiplexERP
+  makeMultiplexERP: makeMultiplexERP,
+  isErp: isErp,
+  isErpWithSupport: isErpWithSupport
 };
--- a/src/util.js
+++ b/src/util.js
@@ -165,14 +165,6 @@ function getOpt(optObject, option, defaultValue) {
       defaultValue;
 }
 
-function isErp(x) {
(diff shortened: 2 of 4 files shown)


stuhlmueller:
You could place it in `erp.js` and (if you need it in webppl) [wrap it in `header.wppl`](https://github.com/probmods/webppl/blob/dev/src/header.wppl#L59), like some of the other functions `erp.js` is exporting. It seems that your feeling of oddness equally applies to `makeMarginalERP`, `makeCategoricalERP`, and `makeMultiplexERP`, so maybe figuring out the right place for these functions is a separate issue?

Also, would it make sense to define `isErp` as follows?

    var isErp = function(x){ return x instanceof ERP; }

null-a:
> maybe figuring out the right place for these functions is a separate issue

Fair enough.

> Also, would it make sense to define isErp as follows?

I did go that way in #64 originally, but came away from it as `isErp(withImportanceDist(...)) == false`.

stuhlmueller:
> I did go that way in #64 originally, but came away from it as `isErp(withImportanceDist(...)) == false`.

This suggests that we should change how `withImportanceDist` is implemented. We can leave `isErp` as it is in the pull request and open an issue for that, if you don't want to deal with it now (either way is fine with me).

null-a:
> This suggests that we should change how withImportanceDist is implemented.

I'm not so sure. My usual way of thinking about this is that in a language like JS saying that an ERP is any object with a couple of functions on it is a good way to. The `ERP` constructor thingy is just one way of creating such an object. I'm expressing a preference for duck-typing rather than class based inheritance type stuff I guess.

stuhlmueller:
My intuitions go the other way. The considerations are (1) you could have an object with `sample` and `score` functions, but where these functions don't have the same meaning as for ERPs and (2) the `isinstance` version is more robust to future interface changes. I think these are weak arguments, so I'm happy to go with the duck-typing solution, too.

null-a:
I only have weak arguments too. How about we stick with this for now, but consider getting `instanceof` working as part of #119? (Or sooner if the opportunity arises.)

stuhlmueller:
Sounds good.

null-a:
Great, thanks!

PR #162, opened 2015-08-12
ERP serialization and printing

iffsid:
Some further functionality to serialize, deserialize, and print erps


null-a:
A bunch of somewhat overlapping thoughts that may or may not make sense for the use cases you have in mind:

* I think `makeMarginalERP` expects scores not probabilities, in which case `erpFromString` needs `util.logHist`.
* I was initially confused by the fact that `toJSON` returns an object. I had a very strong expectation that this would return a string. Perhaps this is idiomatic JS and I'm just not aware of it? If not, perhaps we can think of a better name?
* One idea I had is to rename `toJSON` to e.g. `toObj` (or something better) and have a new `toJSON` function which takes care of calling `JSON.stringify` on the result of `toObj`. We could also rename `erpFromString` to `erpFromJSON` so that the symmetry is more obvious. i.e. `erpFromJSON(myErp.toJSON()) == myErp`, kinda.
* Having `toJSON` and `print` communicate via the magic string `<erp>` seems brittle. Unless there's a sane JSON/object representation of an ERP without support I guess I'd `return` undefined or `throw` an error. (And not write to standard out.)
* I've been passing `this` as the 3rd argument to methods like `_.forEach` instead of doing things like `var cc = this;`. (This way saves having to name the variable if nothing else.) For readability, we should maybe consider picking a conventional way to do this. Any thoughts?

stuhlmueller:
Great review, thanks Paul!

> I was initially confused by the fact that toJSON returns an object. I had a very strong expectation that this would return a string. 

Yes, `toJSON` should return a string ([compare to Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON)).

> I've been passing `this` as the 3rd argument to methods like `_.forEach` instead of doing things like `var cc = this;`. (This way saves having to name the variable if nothing else.) For readability, we should maybe consider picking a conventional way to do this. Any thoughts?

Good idea. I didn't know you could do this via a third argument to functions like `each`. In cases where we are using such functions, this seems like an elegant way. In other cases, I mostly use `bind`:

    function(){
      // ... use 'this' here ...
    }.bind(this).

Are there other good approaches? If not, this two-case approach could be our convention.

iffsid:
- `makeMarginalERP`: ah yes. In this case, I think it would be better to serialize the score and exponentiate when printing.
- The idea is that whatever `toJSON` returns is also serialized when `stringify` is called [ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify). I return an object because it composes nicely with `print` (doing a sort).
- `<erp>`: agreed; magic string not best idea -- I only put that in because the old print function displayed it when printing out an erp without support.
- `this`: In this case it looks like the third argument is possibly the best solution. I usually pull it out into a separate variable to make reading the code and the context cleaner, but it's not something that needs to be done here.

iffsid:
> Yes, toJSON should return a string (compare to Date).

The `toJSON` behaviour I referred to seems different from the one with `Date`. [Here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify), it specifies that the `toJSON` method customizes `JSON.stringify` such that it's result is then serialized; so all that `toJSON` does is convert your previously unserializable object into something that is serializable. This could be a string, but is not required to be.

stuhlmueller:
Can you find an example where someone implemented `toJSON` in a way that doesn't return a string? Based on a cursory search, it seems to me that the standard behavior is to return a string.

iffsid:
One example is [here](http://www.knockmeout.net/2011/04/controlling-how-object-is-converted-to.html)
Based on the discussion [here](https://stackoverflow.com/questions/20734894/difference-between-tojson-and-json-stringify), this seems like well-defined behaviour.

stuhlmueller:
Thanks for finding the example and discussion! I now think your use of `toJSON` is fine.

iffsid:
Cool!
I fixed up the `makeMarginalERP`, `this` invocation, and the `<erp>` string dependence.

null-a:
> I now think your use of toJSON is fine.

Me too!

This makes me a bit nervous:

`this.sample.name.match(/Sample/).index`

I imagine it works for all ERP at the moment but we could easily break it accidentally. Maybe be more defensive here and don't assume that the name contains `Sample`?

Also, I think I'd be inclined to put that bit of logic in `print` rather than `toJSON` as it seems more to do with having nice output than been a meaningful serialisation of the ERP.

null-a:
> Are there other good approaches?

Non come to mind.

> If not, this two-case approach could be our convention.

Sound good.

iffsid:
> This makes me a bit nervous:

It only holds for erps without support. As I understand it, we can't programmatically construct such an erp (I mean as a result of inference), so it must be from an erp that was explicitly constructured. We could adopt a policy that such erps are always constructed with named functions like `erpTypeSample`, `erpTypeScorer`, etc.

If we don't want to assume such conformance, then maybe we just print `erp.sample.name`? With the small caveat that the functions used to build the erp are always named. Thoughts?

> Also, I think I'd be inclined to put that bit of logic in print rather than toJSON ...

I put it in toJSON because of its use in writing to `localStorage`. If you don't provide somthing for transformation of the erp for serialization, it just becomes `{}`, and I wanted the written data to be debuggable at the very least; hence the informative name goes in as the string.

stuhlmueller:
A more robust alternative to looking at the name of the `sample` function may be to check whether the ERP object has `name` property, and use that if it's there. Do that in all cases, not just when the support is `undefined`. So, a serialized ERP would be an object with (up to) two properties, `name` and `support`. If the ERP has neither `name` nor `support`, throw an error in `toJSON`.

iffsid:
I agree. I think the `name` property is the right way to go about it.

iffsid:
ok. I believe now one should be able to call print on *any* erp and expect reasonable output.
I added a `name` field and a `parameterized` field (and related method `parameterize`) to erps to tell if they can be serialized. `bernoulliERP` for example, cannot be serialized because it needs parameters. The `parameterize` method allows you to now call `bernoulliERP.parameterize([0.5]).print()` and get the right output.

The only caveat to all this is that if you serialize to a string, and recover, `name` is set back correctly, but the actual sampler/scorer is got from `makeMarginalERP`; so the specialization that was done for `deltaERP` is lost when deserialzing from string. It's not wrong; just (possibly) inefficient. 

One way to deal with this involves revamping `erp.js` so that all the samplers and scorers are available in the global scope of the file, save the _names_, and rebuild erps from those names. I think that is better left for later though.

null-a:
Very interesting!

* I don't think `parameterize` should be a mutating operation.
* I would return `{ name: this.name, erp: {} }` from `toJSON` for ERPs without support/parameterized ERPs.
* Naming-wise, it's odd that calling `parameterize()` makes `this.parameterized == false`. One idea is to go with something like `fixParameters()` instead.

It would be remiss of me not to ask: Do we actually *need* this functionality or have we been unwittingly led to it because of some of our earlier choices? I can see that serializing marginal ERPs is useful, but if that's *all* we need we can probably do something simpler than this.

Having said that, it looks like `parameterize` could be useful with `withImportanceDist` where (at least in the test suite) we have special functions to create unparameterized versions of e.g. `bernoulliERP`. (See `importance.wppl` for an example.) Then again, maybe we can come up with an even better solution for this when we think about #119.

iffsid:
- mutation: drat, that was unintended. fixed.
- toJSON now always returns an object `{name: *, erp: {*}}` unless it errors out
- changed `parameterize` to `fixParameters`

> Do we actually need this functionality ...

I wanted to be able to save and restore parameter-fixed erps if possible. It's probably not essential for non-marginal/categorical erps, but it's useful as in the importanceDist case.

> ... it looks like parameterize could be useful with withImportanceDist ...

That was one of the reasons I put it in. The other was uniformity across all erps

null-a:
> changed parameterize to fixParameters

This still *sounds* like a mutating operation, I now think `withParameters` would be much better. Sorry.

I also noticed that `binomialERP` doesn't have a name, not sure if that's intentional?

Other than that, this looks good to me!

stuhlmueller:
> returns {name: \*, erp: {*}}

Use `dist` instead of `erp` here (since the whole object is an `erp` object)?

What are the lines with `baseERPEntries` for? If they're not essential, consider getting rid of them, as it would make the code cleaner?

If we go through with this change, auxiliary ERP parameters will be a lot more common (since most ERPs will have `parameterized` set to `true`). In this case, I think the ERP definitions would be more readable if we used

    var gaussianERP = new ERP({
      sampler: gaussianSample,
      scorer: gaussianScore,
      name: 'gaussianERP',
      parameterized: true
    });

instead of

    var gaussianERP = new ERP(
      gaussianSample,
      gaussianScore,
      {
        name: 'gaussianERP',
        parameterized: true
      }
    );

What do others think?

stuhlmueller:
Should `parameterized` be set to `true` by default, since this is the more common case?

stuhlmueller:
Or should we have a `parameterless` property (or `unparameterized`, or `static`, or something like that), which we set to `true` in the few cases where the ERP's distribution actually doesn't depend on parameters?

null-a:
> Should parameterized be set to true by default, since this is the more common case?

Seems reasonable. I'm not sure I have a strong opinion about `parameterized` vs. `parameterless` at the moment.

I also thought the original idea of trying to take `name` from the sample function was a reasonable thing to do, I just thought we needed better handling of the case where it was missing. Maybe we could try and extract the name in `ERP()` and error if it's missing. We could also support a `name` arg if we think that's useful. (e.g. If we create ERPs reusing a sampling function.) This might allow us to remove most or all of the `name: 'somenameERP'` lines.

stuhlmueller:
I'd be happy with a solution where the `name` parameter is set by (1) using an explicit parameter if it's given, (2) trying to extract it on ERP creation from the sampling function if it's not given, and (3) throwing an error if neither of 1 and 2 is the case. I'm also happy with a solution where we don't do (2).

In either case, explicitly naming all parameters to ERPs (as in my comment above) could be good, but we can also leave that for a separate pull request.

iffsid:
> I now think withParameters would be much better.

Fixed. Also fixed missing binomialERP properties.

> Use dist instead of erp ...

Fixed.

> Should parameterized be set to true by default ...

I think this would be the right thing to do. Fixed.

> What are the lines with `baseERPEntries` for?

Lifted erps have this additional property that stores the equivalence class elements and raw probability values.

As for the question of naming, I think the separate property is a reasonable solution. It removes dependence on named functions being used as samplers or scorers; a requirement that really has no need to be enforced at all. 

I agree with @stuhlmueller that, if we choose to do so, changing the erp interface is probably left to a separate pull request.

stuhlmueller:
Thanks!

> Lifted erps have this additional property that stores the equivalence class elements and raw probability values.

This is too special-purpose to be part of the core ERP methods. Consider extracting this functionality and `liftERP` itself into a separate `webppl-lift` package.

iffsid:
I took out the refs to `baseERPEntries` for now.

The specifics of the `baseERPEntries` property aren't important. The question is whether properties/methods can be added to erps outside the webppl core (`erp.js`). If they can, then the serialization methods must be able to act on them.

We can deal with this in one of two ways:
- modify the erp interface and the serialization such that one can iterate over auxillary params for the erp, enabling saving and restoring for any arbitrary (serializeable) property/method. E.g. extend the erp object to be `{name: *, dist: {*}, auxillary: {*}}`
- outsource the extra serialization work to a separate package (that also, perhaps, implements the adding of the aforementioned properties to the erp object. This might cause problems with composing multiple such packages though.

Thoughts?

null-a:
> Thoughts?

I wonder if trying to make it possible to serialize/deserialize any ERP (with a support) *with high-fidelity* is  solving a problem which is more general than the one we have. There seems to be a sweet spot where we get nearly all the functionality without introducing much complexity:

In particular, we could *think* of this as though:

* We only support (de)serializing categorical ERP.
* To aid serialization, some ERP can be converted to categorical ERP.

The difference with where we're currently heading been the acknowledgement that serialization support is limited, and might loose fidelity. (We preserve the distribution of course, but perhaps we lose meta-data, optimized sampling routines, etc.)

This could be as simple as:

```
ERP.prototype.toJSON = function() {
  if (!this.support || this.parameterized) { throw 'Cannot serialize ERP.' }
  var support = this.support([]);
  var probs = support.map(function(s) { return Math.exp(this.score([], s)); }, this);
  return [probs, support];
};

var erpFromString = function(s) {
  return makeCategoricalERP.apply(null, JSON.parse(s));
};
```

(This also has the advantage of not including each element of the support twice in the serialized representation. We might want to fix that even if we don't/can't take this stripped down approach.)

This approach implies that we wouldn't try to serialize extra properties of ERP. In the case of `liftERP`, can we make do with the ability to serialize either the input to `liftERP` or the returned ERP without its `baseERPEntries`?

null-a:
I've also been thinking about naming. If I understand correctly we have names:

1. As an aid to debugging.
2. To use when printing to the console.

I'm not sure this justifies their inclusion.

* Point 1 relates to having `{}` end up in local storage. It sounds like having the name around is a way of learning something about the state of the program at some point in the past, but perhaps having serialization fail early (e.g. by throwing an error as above) might be preferable. If the problem is that debugging webppl programs is hard, then we should fix that.
* Point 2: I rarely print anything other than marginals to the console, so this doesn't seem super useful.
* When we deserialize we call `makeMarginalERP` and then change the `name` of the resulting ERP. I can imagine this leading to confusion.

Unless I've misunderstood or there are other reasons to have it, I think I'd drop the naming stuff.

stuhlmueller:
This proposed simplification (essentially limit serialization to categorial, don't introduce names) sounds good to me.

iffsid:
My necessity for serialization comes from situations where I need to
 1. augment erps with additional properties, like `baseERPEntries` for liftedERPs
 2. save and restore such erps from localStorage/disk, without losing the additional properties.

If the overhead for saving the additional properties is too much, then they could just be saved separately from the erp and restored in a similar manner. It's not ideal, but should work.

I'm not wedded to the idea of naming; if people feel we can forego the naming, then that's fine by me.

stuhlmueller:
> outsource the extra serialization work to a separate package (that also, perhaps, implements the adding of the aforementioned properties to the erp object). 

This still sounds like the best solution for your specific problem. This way, you don't have to compromise, but also don't need to add special-case code to core webppl.

> This might cause problems with composing multiple such packages though, since the modification to the serialization may not be associative (although I'm finding it hard to visualize such a scenario).

It's not clear that you need to consider composing multiple such packages at all at the moment - just write a package that does exactly what you need, and think about composition later if it ever becomes necessary.

iffsid:
Fixed up the naming and simpler serialization.
Now throws an error if attempting to serialize an erp requiring parameters (`parameterized`), or without support.

stuhlmueller:
Looks good to me. Add a test case checking that `erpFromString(erp.toJSON())` is equivalent to `erp` for some simple erp?

null-a:
>  then they could just be saved separately from the erp and restored in a similar manner. It's not ideal, but should work.

At least the `toJSON` interface means we can do `var str = JSON.stringify({ erp: marginalERP, extras: [] })` and have the right thing happen. Admittedly there's a little bit of a mismatch going the other way, since:

```
JSON.parse(str).erp // => [probs, support]
```

I guess we could modify `makeCategoricalERP` to take such pairs if need be.

iffsid:
For augmenting the serialization process, I simply do something like
``` js
var sr = erp.toJSON().concat([extras])
// => sr = [probs, support, extras]
```
`makeCategoricalERP` already takes an `extraParams` argument that just plugs into the `auxParams` argument in the erp interface.

For my purposes, in a separate package, I rewrite the serializer (in my package) as
``` js
// my-package-src.js
var oldtoJSON = ERP.prototype.toJSON;
ERP.prototype.toJSON = function() {
  var basicJSON = oldtoJSON.bind(this)();
  var newJSON = basicJSON.concat([{baseERPEntries: this.baseERPEntries}])
  this.toJSON = function() {return newJSON};
  return newJSON;
}
```
``` js
// test.wppl
var erpA = liftERP(categoricalERP([0.2, 0.3, 0.4, 0.1], [1, 2, 9, 10]),
                   function(s) {return s > 5 ? 'A' : 'B';});
var erpB = erpFromString(JSON.stringify(erpA))
erpB.baseERPEntries
// =>  { '"B"': 
//       { vs: [ 1, 2 ],
//         ps: [ -1.6094379124341003, -1.2039728043259361 ] },
//       '"A"': 
//       { vs: [ 9, 10 ],
//         ps: [ -0.916290731874155, -2.3025850929940455 ] } }
```

iffsid:
I made a minor change resulting from both @null-a and @stuhlmueller's comments.
The JSON returned by toJSON is now actually an object `{erp: [probs, support]}`. This makes catching errors easier; otherwise any array of two sub-arrays could be converted to an erp.

The augmenting (as in my previous comment) has also been updated -- latest code is available in my webppl package that provides modifications to the erp object, [`webppl-erpMods`](https://github.com/iffsid/webppl-erpMods)

iffsid:
Added simple test to check that serialization doesn't break anything.

stuhlmueller:
Thanks!

> The JSON returned by toJSON is now actually an object `{erp: [probs, support]}`.

Why not `{probs: ..., support: ...}`?

iffsid:
> Why not `{probs: ..., support: ...}`?

Primarily because I would like to avoid calling special functions like `makeCategoricalERP` from external libraries, and instead just modify the JSON serialization. E.g: the [erp lifting](https://github.com/iffsid/webppl-erpMods/blob/master/src/lifterp.js) package just augments the JSON object.

Having it as a list means that in `erp.js`, I can just use `makeCategoricalERP.apply` on the serialized JSON (as a list) to get back a webppl erp object, and not worry about explicitly dealing with any additional fields the serialized JSON had.

stuhlmueller:
If you're overwriting `toJSON` in your library, why not also provide a custom `erpFromJSON` function in your library? `makeCategoricalERP` is exported, so why not use it?

I don't think we should choose a less intuitive representation for JSON-ified ERPs just because it makes writing certain sorts of libraries slightly easier. If you feel strongly about not redefining `erpFromJSON` in your library, you can make it pass on any extra fields in the JSON-ified ERP as `auxParams` to `makeCategoricalERP`.

iffsid:
I was somewhat uncomfortable with redefining globally available standalone functions; less so with prototype definitions because of looking-up-in-prototype-chain that happens for objects. But perhaps that is not actually a concern.

In any case, I made a change such that the JSON object is now `{erp: {probs: *, support: *}}`.
I explicitly add the `erp` key to act as a type check (An explicit type name is better instead of having to check on `probs` and `support`).

You can potentially attach any extra properties to the `erp` key besides `probs` and `support`, and it should be handled correctly through `makeCategoricalERP`.

stuhlmueller:
> I explicitly add the `erp` key to act as a type check (An explicit type name is better instead of having to check on `probs` and `support`).

I don't see how it's better -- if you're worried that people will pass in objects that have properties `probs` and `support` but where probs and support aren't of the right type, I'd think you should also be worried that people will pass in an object with property `erp` that isn't an object of the right kind. It seems equally good and more straightforward to just serialize to `{probs: *, support: *}`.

iffsid:
True, but it's a little more painful to pass in an object with `erp`, `probs`, and `support`, rather than just the latter two.

Another reason for having it is that we could attach properties specific to the serializers and de-serializers to the json object. Helps with extensibility. Like when saving `multiplexERP`s or on `kde`'d erps, where the object can be
``` js
// where support is a list of unparameterized distributions themselves
{erp: {probs: *, support: *, erpAux: *}, mixture: true}
```
OR
``` js
// the actual distribution is unparameterized with just `vs` and `ps`
{erp: {probs: *, support: *, erpAux: *}, kdeKernel: 'epanechnikov', kdeBW: 1.2}`
```
where the additional fields don't really belong together with the `probs` and `support` keys, or are a relevant to the `makeCategoricalERP` function.

In any case, it's not really that much of a stretch to have `erp` is it? Makes the json object simpler and less confusing to display too.

stuhlmueller:
I'm not convinced -- if people write special-case (de)serializers in the future, it seems that they can just as well use extra properties that are part of the top-level object. However, this is a tiny issue we're talking about here, so I don't think it's worth spending much time on. Maybe @null-a can take another quick look at this and if it looks ok to him, it's fine with me as well and we can merge it.

null-a:
*Root node*

I have a very slight preference to not include a root `erp` node. The check in `erpFromJSON` doesn't seem so bad either way and viewing the result of serialization seems like the kind of thing usually done with enough context to know what you're looking at. Other than those reasons, having this seems redundant. In the future it may become apparent that having this is useful, but so long as we're not too worried about making breaking changes we can worry about that if and when it happens I think. It hardly seemed worth changing, but since it only takes a few seconds I went ahead and did it under the assumption I had the casting vote.

*Naming*

Something I feel more strongly about is the naming of the deserialization functions. It seems to me that the function which takes a string of JSON to an ERP should be called `erpFromJSON` rather than `erpFromString`. I suspect that this would be easier for a new user to guess at, or to find in our (not yet existent) documentation.

Another way to see that the current naming might be a problem is to imagine that we also serialize to another format such as XML. In that case it seems clear that `erpFromString` is problematic.

Would it be possible to fold `erpFromString` and `erpFromJSON` into a single function? Something like:

````javascript
var erpFromJSON = function(json) {
  var obj = JSON.parse(json);
  if (!obj.probs || !obj.support) {
    throw 'Not an ERP JSON object!'
  }
  return makeCategoricalERP(obj.probs,
                            obj.support,
                            _.omit(obj, 'probs', 'support'));
}
````

**Comment by iffsid** [comment]
_Root_

I have a preference for the presence of the `erp` key; it makes the kinds of extensions I'd like to make to the erp serializers easier to separate from actual erp properties themselves (serializing with kde, as opposed to lifting, for example).

However, it's not critical that it be present; perhaps we can revisit this issue sometime later when it looks like it might be worth extending the erp interface more. I'm fine with the change.

_Naming_

I'm thinking of the operations from the point of view of types. The result of JSON.stringify, as I can tell, is not of type `JSON`, but of type `string`. Hence deserializer called `erpFromString`.

Keeping with this interpretation, the serialization process proceeds as:
```
                  JSON.stringify
           toJSON               stringify
ERP ----------------> JSON ----------------> string
```
and to mirror this, the deserialization process proceeds as:
```
                 erpFromString
         JSON.parse          erpFromJSON
string -------------> JSON ----------------> ERP
```

Are there any particular instances where it's actually common to assume the result of JSON.stringify is a  JSON object? From the [definition](https://en.wikipedia.org/wiki/JSON) and the documentation for toJSON, it seems to be the case that the `string` interpretation is correct.

**Comment by null-a** [comment]
> The result of JSON.stringify, as I can tell, is not of type JSON, but of type string.

Yup, it's a Javascript `string`, but we also know it's a string of valid `JSON`. Why include just the type in the function name and not this more specific information when doing so would seem to help for the reasons I suggested?

> Are there any particular instances where it's actually common to assume the result of JSON.stringify is a JSON object?

My understand is that the stuff in the string is in JSON format, and the reset are just Javascript object.

```
                  JSON.stringify
           toJSON               stringify
ERP ----------------> Object ----------------> string (JSON format)
```
and to mirror this, the deserialization process proceeds as:
```
                 erpFromString
         JSON.parse          erpFromJSON
string (JSON format) -------------> Object ----------------> ERP
```

**Comment by iffsid** [comment]
> Yup, it's a Javascript string, but we also know it's a string of valid JSON.

That's true. (As I understand it, the only distictions are `function`, `-Infinity`, `Infinity`, and `NaN` and unicode stuff).

However, if you want to change `erpFromString` to `erpFromJSON`, then you lose the symmetry of `toJSON` giving you a JS Object and `erpFromJSON` acting on a JS object. If you really feel strongly about this, I could change `erpFromString` to `erpFromJSONString` to make things explicit. Then
```
               erpFromJSONString
         JSON.parse          erpFromJSON
string -------------> JSON ----------------> ERP
```
would be cleaner in referring to type and underlying format.

**Comment by iffsid** [comment]
Actually, to avoid all this naming confusion, would people be happy simply with the following?
```
serializeERP :: ERP -> JSON string
deserializeERP :: JSON string -> ERP
```
This, I think is better because as it turns out, we can't simply call `JSON.stringify` on an erp because the JSON format does not handle `Infinity` and `-Infinity` (it replaces them with `null`), so we need a little augmenting serializer and deserializer for those values; which are conveniently arguments for `JSON.stringify` and `JSON.parse`.

**Comment by null-a** [comment]
>  then you lose the symmetry of toJSON giving you a JS Object and erpFromJSON acting on a JS object

To restore symmetry I would give up on mirroring Javascript's `toJSON` semantics and have it return a string rather than an object.

> Actually, to avoid all this naming confusion, would people be happy simply with the following?

**Comment by iffsid** [comment]
> To restore symmetry I would give up on mirroring Javascript's toJSON semantics and have it return a string rather than an object.

This would make a lot of things unnecessarily painful - any augmenting or manipulation of the serialized erp (pretty printing, lifting, etc) would have to parse it back into an object and then re-stringify it. This seems like way too much work just to deal with naming issues.

**Comment by null-a** [comment]
> Actually, to avoid all this naming confusion, would people be happy simply with the following?

Looks good to me!

Out of interest, when do you imagine we'll serialize `(-)Infinity`?

**Comment by iffsid** [comment]
`(-)Infinity` could be part of the support of a distribution

**Comment by iffsid** [comment]
Actually, this `JSON.stringify` and `(-)Infinity` issue is probably affecting someone without them knowing.
(I've so far dealt with it explicitly wherever I need)
The keys for the histograms are built with this; so are the keys for the function arguments in caches.
We might have to raise an issue to make sure that we change the usage to a more general `serialize` function, globally available, that handles this correctly.

**Comment by null-a** [comment]
> The only caveat with the new naming scheme is that you'd have to handle the call appropriately. Instead of simply calling JSON.stringify to construct the key, you'd have to check if it was an erp, and call serializeERP appropriately.

Are you talking about keys for caches and histograms here too, or something else?

**Comment by iffsid** [comment]
> Are you talking about keys for caches and histograms here too, or something else?

Any location where we construct an `Object` key by calling `JSON.stringify` would need to handle the case of `(-)Infinity` serialization. That's why I proposed that we could construct a global function, say, `serialize`, that handled this issue.
``` js
function InfToJSON(k, v) {
  if (v === Infinity) {
    return 'Infinity';
  } else if (v === -Infinity) {
    return '-Infinity';
  } else {
    return v;
  }
}

function serialize(obj) {
  return JSON.stringify(obj, InfToJSON);
}
```
We'd then use serialize everywhere we currently use `JSON.stringify` to create keys.

`serializeERP` would then not be necessary, as it would be doing the same as `serialize`.

null-a:
> serializeERP would then not be necessary, as it would be doing the same as serialize.

I think we'd still need something like `deserializeERP` though, right? So perhaps a matching `serializeERP` would be useful?

How about we have `util.(de)serialize` to do JSON (de)serialization with `Infinity` handling from within Javascript, and expose `(de)serializeERP` via the header for WebPPL?

iffsid:
> I think we'd still need something like deserializeERP though, right?

Yes, it should be included; at least for symmetry's sake. Currently, `(de)serializeERP` are exported via `header.wppl`.

As for `(de)serialize`, might it not be better to have it available unqualified similar to 'display`, `cache`, etc? Perhaps by adding it to `headerUtils.js` (it really ought to be available both in the front-end code and the back-end core code)?

null-a:
> might it not be better to have it available unqualified

Sure, it might! Cross that bridge when we get to it?

iffsid:
Sure :-)
I added them to util for now.

null-a:
This looks good to me, I think we should merge it.

@stuhlmueller: Are you happy?

PR #163, opened 2015-08-13
Handle calls to sample which omit the params arg in IncrementalMH

null-a:
We sometimes call `sample` omitting `params` when it would be empty. For examples see the test models `cache.wppl` and `importance.wppl`.

This change makes that work with IncrementalMH.

Testing this the via the `cache` model is somewhat indirect, but the strategy of creating a new test called something like `sampleWithoutParams` seems like it would get out of hand pretty quickly.

I've not added this test to AsyncPF as I believe it doesn't work on models without factor statements.


stuhlmueller:
What if `params` is `undefined`, but `this.params` wasn't undefined? More generally, what if the length of the old and new parameter arrays doesn't match? I don't know if we have ERPs with optional parameters right now, but they're easy to imagine, so unless we explicitly disallow this, it would be good if the check in `registerInputChanges` were robust to that.

null-a:
Good point!

This should fix that and it includes a test which triggered the bug when `this.params` xor `params` was `undefined`.

Adding a test for the case where `params` and `this.params` are both arrays but differ in length is harder. I think I'd need to add a contrived ERP tailored for the job. I'm not sure that's a good way to go.

stuhlmueller:
Thanks! Make the test a little more strict, like this

    var model = function() {
      var x = flip(0.5);
      var y = sample(FairCoin, x ? [0] : undefined);
      var z = sample(FairCoin, x ? undefined : [0]);
      return y && z;
    };

and check that the return distribution is 3/4 false, 1/4 true?

> Adding a test for the case where `params` and `this.params` are both arrays but differ in length is harder. I think I'd need to add a contrived ERP tailored for the job. I'm not sure that's a good way to go.

Yeah, I don't think that's necessary.

null-a:
> Make the test a little more strict

Done!

Issue #156, opened 2015-08-06
SMC returns #particles * (#rejuvenation-steps + 1) samples


stuhlmueller:
`#samples = #particles * #rejuvenation-steps` would be more intuitive.

Example:

    var y = ParticleFilterRejuv(
      function(){
        return uniform(0, 1);
      }, 
      3, // particles
      5  // rejuvenation steps
    );
    
    y.support().length

This returns 18 = 3 * (5 + 1).

The explanation for the additional sample is that for each particle, we run SMC until we're at the end of the program, which gives us one sample per particle; then we additionally take #rejuvenation-steps MH samples, each of which gives us another sample.

null-a:
Do you imagine achieving this by only returning the samples from the rejuvenation steps? (Taking `rejuvSteps - 1` steps seems like it would be more confusion than what we have now.)

stuhlmueller:
Yes, if there's a way to do this and still have it return a sample in the case of 0 rejuvenation steps. If there is no elegant way, then the current solution of `#particles * (#rejuvenation-steps + 1)` samples isn't that bad either.

null-a:
Closed by #169.

Issue #155, opened 2015-08-06
MH and SMC throw error for models without erps


stuhlmueller:
Example:

    // Doesn't work:
    
    var x = MH(function(){
      return 1;
    }, 10);
    
    print(x);
    
    // Works:
    
    var y = MH(function(){
      return flip(.5);
    }, 10);
    
    print(y);

Error:

    Uncaught TypeError: Cannot read property 'score' of undefined
    MH.exit @ webppl.min.js:35

null-a:
For reference this also affects `HashMH` and `IncrementalMH`. The existing `deterministic` test case triggers the bug.

null-a:
#169 resolved this problem for `MH` and `SMC` and we no longer have `HashMH`.

Issue #151, opened 2015-07-15
Support query table in MH


stuhlmueller:
It's currently supported in `HashMH` and `IncrementalMH`, but not `MH`. This can happen as part of refactoring #64.

Issue #149, opened 2015-07-13
Bug in PFRejuv Store Handing


null-a:
I expect [this test](https://github.com/null-a/webppl/commit/c20c5d6ed04dc35aeeca973035e33dc400a6f6cf) to always return `{ 'true': 1 }` but PF with rejuvenation is a little off:

```
Expected: { true: 1 }
Actual: { false: 0.02580645161290322, true: 0.9741935483870968 }
✖ PFRj - store2

Expected: { true: 1 }
Actual: { false: 0.019801980198019802, true: 0.9801980198019803 }
✖ PFRjAsMH - store2
```

I guess this is a problem with the store handling, but I've not tracked it down yet. I don't think my work-in-progress refactoring of MH & SMC (#64) suffers from this, though I'm not sure we'd want to wait for that to be merged before having a fix for this.

stuhlmueller:
My hunch would be to work on the refactoring instead of tracking down the bug, unless you think it won't take long to fix it, or that it will take multiple months for the refactoring to make it into dev.

null-a:
Closed by #169.

PR #169, opened 2015-08-25
Refactored MH & SMC

null-a:
As discussed in #64. I don't think there are any significant changes left to make so now seems a good time to open a PR.

**Changes**

* `MH` and `HashMH` are replaced by a new `MCMC` routine which is parameterized by a transition kernel. The important features of both `MH` and `HashMH` have been retained, and [performance is expected to be very similar](https://github.com/probmods/webppl/issues/64#issuecomment-133002277) to that of `HashMH`. The intention is that this can be used with other kernels though at present there is only `MHKernel`. (I've not tried to make this work with `HMC` or `IncrementalMH` yet.)
* `ParticleFilter` and `ParticleFilterRejuv` are replaced by a new `SMC` routine. This optionally takes a transition kernel and uses it to do rejuvenation. Performance is very similar [or better](https://github.com/probmods/webppl/issues/64#issuecomment-134285022) than the corresponding `ParticleFilter` and `ParticleFilterRejuv` methods.

I've also endeavored to improve readability and reduce duplication where I can. I hope this is evident when you look at the code, but if not let me know and I'll point out some areas where I think things have improved.

**Interface**

Using the new methods looks like this:

```
MCMC(model, { samples: 1000 });
MCMC(model, { samples: 1000, lag: 5, burn: 100, kernel: MHKernel });
MCMC(model, { samples: 1000, verbose: true }); // Shows current iteration and acceptance ratio.
MCMC(model, { samples: 1000, onlyMAP: true });
MCMC(model, { samples: 1000, justSample: true });

SMC(model, { particles: 1000 });
SMC(model, { particles: 1000, rejuvSteps: 10 });
SMC(model, { particles: 1000, rejuvSteps: 10, rejuvKernel: MHKernel });
```

The hope is that this will work nicely with the interface changes proposed in #86. I've created wrappers `MH`, `ParticleFilter` and `ParticleFilterRejuv` for backwards compatibility.

**Minor Changes**

* `MCMC` takes an option `samples` which specifies how many samples you want back, rather than the number of iterations to perform. The number of iterations is computed from `samples`, `lag` and `burn`.
* The `justSample` option of `MCMC` returns the MAP estimate, not just the samples. This part of the interface still needs work. (See [#86](https://github.com/probmods/webppl/issues/86#issuecomment-100244114), #140.)
* I've reworked the handling of variable numbers of factors in the particle filter. Instead of checking whether each particle is active when manipulating `particleIndex` I partition the particles into two sets - active and complete. This implements the same strategy but might be easier to understand, though I'm not sure. I left it in for people to have a look at.
* I've removed the `varFactors` tests from `ParticleFilterRejuv`. The `varFactors1` test wasn't actually doing any rejuvenation and it fails when this is corrected. ([Related discussion](https://github.com/probmods/webppl/issues/20#issuecomment-133129248).)
* We no longer have the `diagnostics` option which was part of `MH`. Instead `justSample` can be used to get all of the samples in order to perform convergence checks etc.

**Also Fixed**

* #51
* #120
* #145
* #147
* #149
* #151
* #155 (`IncrementalMH` needs the fix too.)
* #156


stuhlmueller:
Thanks Paul! I've started looking over the code and will have comments before the end of the week.

stuhlmueller:
I've looked over everything. Overall I think this represents a great step towards making webppl cleaner. 

**Main comment**

If I could only give one comment, I'd say that I wish that the interface and/or names for `MCMC`/`runMarkovChain` and for `SMC`/`ParticleFilter` were more intuitive. The names for the two elements in each of these pairs seem close to exchangeable, but they have different interfaces, and without looking it up, I find it difficult to remember what exactly `ParticleFilter` and `runMarkovChain` do. I don't think the interface of `MCMC` and `SMC` needs changing, so:

  1. Do we need `ParticleFilter`, or can we fold it into `SMC`? Without looking at the code, one would guess that `ParticleFilter` doesn't include rejuvenation, whereas `SMC` does, but in fact both do, so maybe we can just merge them?
    
  2. A clean way to deal with repeated application of MCMC kernels would be to use a higher-order function `iterateKernel: (Kernel, N) -> Kernel` that takes a kernel and produces a new kernel that applies the original kernel `N` times. Unfortunately, `runMarkovChain` does some extra bookkeeping. Could this happen somewhere else?

**Other comments**

On `src/inference/initialize.js`:

- It's a little unintuitive that calling `var x = new Initialize(s, k, a, wpplFn)` has the side effect of changing `env.coroutine`. Same for `MHKernel`. I realize that this is what's happening in the current implementation, too, but I wonder if we should have a method that does this more explicitly, like `install`? On the other hand, if we never create objects like `Initialize` without immediately installing them as the new coroutine, maybe the current approach is okay.

- How would you feel about replacing method definitions like

        Initialize.prototype.run = function() { ... }
        
        Initialize.prototype.sample = function() { ... }

    with

        Initialize.prototype = {
        
          constructor: Initialize,
        
          run: function() { ... },
        
          sample: function() { ... }
          
        }

    To me, this style feels a bit cleaner, as it groups together the methods more explicitly and makes it easier to scan the list of methods. If you endorse this change, but don't want to do it now, I can also make a separate pull request later.

On `src/inference/mcmc.js`: 

- On first reading, the semantics of `lag` aren't clear. Does `lag == 1` mean that after every sample, there's a lag of 1 sample that doesn't get used, or does it mean that we use every sample? Since the default is 1, I assume it's the latter, but I find the other semantics slightly more intuitive.

- Should `MAPEstimator` be in this file? It seems more general.

On `src/inference/mhkernel.js`:

- Looking at `MHKernel` made me wonder once more whether we can eventually implement inference coroutines in webppl (#54). In my mind, a kernel is a map `Trace -> Trace`, and it feels like our code could be more readable and modular if we could stay closer to that idea. (This isn't a proposal for any change in the current pull request.)

- `MHKernel.prototype.sample = function(s, cont, address, erp, params, forceSample) {` - we should standardize on calling the continuation `k` and the address `a` wherever possible. See also `ParticleFilter.prototype.factor`, `ParticleFilter.prototype.rejuvenateParticles`, and probably others. I assume that these are from before the refactor, but we might as well fix it as part of the refactoring.

- `if (prevChoice.val === val) { return this.cont(this.oldTrace, Math.random() < 0.5); }`

    1. Can we have the convention of always using new lines for the bodies of `if` statements?
    2. Why `Math.random() < 0.5`?

- `if (this.trace.score === -Infinity) { return this.cont(this.oldTrace, false); }` - What if `this.oldTrace.score` was `-Infinity`, too? Don't we want to accept then (perhaps unless we're doing rejuvenation, see [here](https://github.com/probmods/webppl/issues/64#issuecomment-133523127))?

On `src/inference/smc.js`:

- Is it worth having `TraceLite`? How much overhead would there be if we used `Trace` for SMC without rejuvenation?

- `ParticleFilter.prototype.lastParticle` - Rename this to `atLastParticle`? Similarly, rename `nextParticle` to `advanceParticleIndex` or something like that? Otherwise, one might expect the functions to return the last/next particle, like the `currentParticle` method.

- `ParticleFilter.prototype.allParticles` - when concatenating the lists, have the completed ones come first?

- `ParticleFilter.prototype.sync` - reorder the `if` cases to handle the shorter (and more common) one first?

- `assert(resampledParticles.length === this.numParticles);` - use `assert.equal`?

- Maybe particles should have a method `isComplete`, so that

        var p = _.partition(resampledParticles, function(p) { return p.trace.k && p.trace.store; });
    
    can be rewritten as 
    
        var p = _.partition(resampledParticles, function(p) { return p.isComplete(); });

null-a:
> Main comment

You're right, this needs some work!

**`SMC`/`ParticleFilter`**

Of the two pairs I think this needs the most attention. An important difference between these two methods is that `ParticleFilter` was intended to operate at the level of traces (though now returns particles) and `SMC` is a public method and is mostly concerned with option handling, building the marginal ERP etc. I don't think we should merge these because:

1. I think separating the details of performing inference from the details of interfacing with the user makes the code easier to understand.
2. Having low-level methods which operate on traces seems like a useful direction to pursue as these compose in useful ways. e.g I previously made `ParticleFilter` available as an initialization strategy for `MCMC`.

I agree the naming is terrible though. One suggestion is to rename `ParticleFilter` to `SMC`, and to rename `SMC` to something like `smc`.

````javascript
function SMC(s, k, a, wpplFn, options) {
  // ...
}

// prototypes here...

function smc() {
  return SMC().run();
}

return { SMC: smc };
````

We make a very similar move in nearly all of the other inference algorithms, and if nothing else we remove the misleading distinction between `SMC` and `ParticleFilter`. This might only be a temporary reprieve though, as we'd need to do something different if we want to make use of the method which operates on traces elsewhere.

Another idea is to use the name of the algorithm for the methods which operate on traces, and call the public interface something different. For example, `ParticleFilter` could become `SMC` and `SMC` could become something like `InferSMC`. This gives the methods which perform inference the obvious names, and once we have the `Infer` interface then `InferSMC` would be hidden behind something like `Infer(model, { method: 'SMC' })`.

**`MCMC`/`runMarkovChain`**

I'm much happier with this than with I am with `SMC`/`ParticleFilter`.

`runMarkovChain` seems to me to capture the essence of running a Markov chain and is reusable across `MCMC` and `SMC` with rejuvenation. Perhaps there's a better name than `runMarkovChain` though. My suggestion from above would see us rename `MCMC` to `InferMCMC` and optionally `runMarkovChain` to `MCMC` - it's not clear to me that this is an improvement.

I like your higher-order `iterateKernel` idea although if I understand it correctly using it here seems less straight-forward than using `runMarkovChain`. It seems as though `iterateKernel` would have to concern itself with tracking the acceptance ratio and accumulating samples. At present this is all done in `MCMC` which seems like the right place to me.

stuhlmueller:
> Another idea is to use the name of the algorithm for the methods which operate on traces, and call the public interface something different. For example, `ParticleFilter` could become `SMC` and `SMC` could become something like `InferSMC`. This gives the methods which perform inference the obvious names, and once we have the `Infer` interface then `InferSMC` would be hidden behind something like `Infer(model, { method: 'SMC' })`.

Sounds good.

> Having low-level methods which operate on traces seems like a useful direction to pursue as these compose in useful ways.

That kind of thing seems good to me, too. If it were accurate to think that the only distinction between `SMC` and `InferSMC` is that one operates on traces and is supposed to be used as a compositional piece in other inference algorithms, and the other interfaces with the user via `Infer`, I'd be happy. In general, it would be great if there were clear rules on (1) what kind of functionality goes into `InferX` vs `X` (in particular for X = `SMC`) and (2) how the interface (arguments, return value) differs between `InferX` and `X`.

> My suggestion from above would see us rename `MCMC` to `InferMCMC` and optionally `runMarkovChain` to `MCMC` - it's not clear to me that this is an improvement.

Yeah, the interface between (current) `MCMC` and `runMarkovChain` would have to be closer for this to be a good idea.

> `runMarkovChain` seems to me to capture the essence of running a Markov chain and is reusable across `MCMC` and `SMC` with rejuvenation.

Besides the name (which should be less confusable with `MCMC`), the interface feels a bit unwieldy to me relative to what the function accomplishes. We're using `runMarkovChain` in `MCMC` for three things:

1. Repeatedly apply MHKernel to a trace
2. At each step, extract the value from a trace if appropriate (depending on lag, burn)
3. At each step, keep track of acceptance counts (this part lives in `MCMC`)

I'd like the code to be more explicit about these pieces, which would hopefully make it easier to read.

One way of doing this is to use kernel combinators, inspired by the approach mit-church took ages ago -- check out [mcmc-core.ss](https://github.com/malmaud/RChurch/blob/master/inst/mit-church/church/mcmc/mcmc-core.ss) (although mit-church didn't actually use this approach for value extraction and tracking acceptance stats). Suppose we include an extra field in `Trace`, maybe named `info` or something like that. MHKernel could set `info.accepted` based on whether the current trace is the result of an accepted or rejected proposal. We could then build the overall Markov chain using:

    var kernel = compose(mhKernel, logAcceptance);
    var chain = sequence(
      initialize,
      repeat(burn, kernel),
      repeat(n,
        sequence(
          repeat(lag, kernel),
          compose(kernel, extractValue))));

This sort of specification of the Markov chain feels a lot cleaner to me.

null-a:
> Other comments

> It's a little unintuitive that calling var x = new Initialize(s, k, a, wpplFn) has the side effect of changing env.coroutine.

I think I might share a similar feeling. The `install` idea is nice but for me doesn't seem to help all that much. There may well be a better way we're not seeing but I'm not sure what it is, so I'm perhaps inclined to leave it as is for now.

> How would you feel about replacing method definitions ...

I agree that it makes the methods easier to scan, and I can't see any downside to writing it this way. If we're going to do this for all inference algorithms perhaps we should tackle it as a separate piece of work?

Why would we need `constructor: Initialize`?

>  the semantics of lag aren't clear

`lag = 1` means collect a sample every iteration. It's this way for consistency with `IncrementalMH` and the previous implementation. I think I recall sharing your confusion as I implemented it. Perhaps options like `{ every: n } == { lag: n }` and `{ skip: n-1 } == { lag: n }` would be clearer. (Though having all 3 would be overkill.)

> Should MAPEstimator be in this file

It's here as it's only used by `MCMC` at present and unlike `Histogram` I don't expect any of the other inference algorithms will make use of it. I'm indifferent to where it lives, so I'll happily move it into it's own file once I've figured out what the file name should be!

> we should standardize on calling the continuation k and the address a wherever possible

Agreed. I've changed `sample` and `factor` in `SMC` etc.

I intentionally called the continuation in `ParticleFilter.prototype.rejuvenateParticles` something other than `k` as it's a thunk rather than a continuation that takes a store and a value. (Perhaps similar to `nextK` in `cpsLoop`.) Do you agree that it's useful to have a convention of using a different name for these?

> Can we have the convention of always using new lines for the bodies of if statements?

Sure. I've removed the one-liners from this PR.

> Why Math.random() < 0.5?

Oops! That's a mistake, good catch.

> What if this.oldTrace.score was -Infinity, too?

When I wrote `MHKernel` I was assuming that `oldTrace.score` would never be `-Infinity` because I keep running `Initialize` in `MCMC` until I find a trace with positive probability, and MH won't transition from a trace with positive probability to one with zero probability.

Setting aside the case where we do `SMC` with a tiny particle set (e.g. `PFRjAsMH`), when might `oldTrace.score == -Infinity`?

> Is it worth having TraceLite

Good question. I think `SMC` has different asymptotics with `Trace` and `TraceLite`. Copying a `Trace` during resampling requires copying the array of choices in the `Trace` rather than just the most recent choice as with `TraceLite`, so I imagine we pick up an extra `n` term, where `n` is the number of factor statements. So overall `SMC` + `Trace` is maybe `O(mn^2)` vs. `O(mn)` for `TraceLite`, where `m` is the number of particles. (Unsure though.)

If that's right, maybe we could use something like a linked list to represent choices to have `SMC` + `Trace` be `O(mn)`. (Although this might give a constant factor slow down everywhere else.)

> Rename this to atLastParticle?

Done.

>  when concatenating the lists, have the completed ones come first?

Why? (I assume this helps but I don't see how - sorry.)

> reorder the if cases to handle the shorter (and more common) one first?
> assert(resampledParticles.length === this.numParticles); - use assert.strictEqual?
> Maybe particles should have a method isComplete, so that

All done!

stuhlmueller:
> The `install` idea is nice but for me doesn't seem to help all that much.

I think it helps some if `install` isn't called from the constructor, but rather from the place where we're currently calling `run`. We don't need to do this as part of the current PR though. I created an issue (#172) where we can further discuss this if it's controversial.

> Why would we need `constructor: Initialize`?

I was assuming that the constructor was still defined outside of the prototype (and so we would need to preserve it when we overwrite the prototype), but now that you mention it, I realize that it should just be:

    Initialize.prototype = {
      constructor: function() { ... }
      run: function() { ... },
      sample: function() { ... }
    }

I also created a separate issue for this (#173).

> Perhaps options like `{ every: n } == { lag: n }` and `{ skip: n-1 } == { lag: n }` would be clearer. (Though having all 3 would be overkill.)

Having either `lag` or `skip` with skip semantics seems fine to me.

> > Should MAPEstimator be in this file

> It's here as it's only used by `MCMC` at present and unlike `Histogram` I don't expect any of the other inference algorithms will make use of it.

Why do you think so? Doesn't it make sense to have all sampling-based algorithms support computing the MAP?

> I'm indifferent to where it lives, so I'll happily move it into it's own file once I've figured out what the file name should be!

Maybe `mapEstimator.js` :). Or put both this and `histogram` into `aggregation.js` or `aggregators.js`.

> I intentionally called the continuation in `ParticleFilter.prototype.rejuvenateParticles` something other than `k` as it's a thunk rather than a continuation that takes a store and a value. (Perhaps similar to `nextK` in `cpsLoop`.) Do you agree that it's useful to have a convention of using a different name for these?

Yes, calling those `cont` seems good.

> Setting aside the case where we do `SMC` with a tiny particle set (e.g. `PFRjAsMH`), when might `oldTrace.score == -Infinity`?

The case I was thinking of was the setting where you don't initialize to a non-zero probability, but rather use MH's local search to find a non-zero trace. If we have the contract that MH can only be initialized with a trace with non-zero probability (assert this), then the only other case I can think of is particle decay in SMC (where you want the currently implemented behavior).

> I think `SMC` has different asymptotics with `Trace` and `TraceLite`.

It would be good to verify this. If there's little difference even for very large numbers of factors and choices, we should just go with `Trace`. If the hypothesized asymptotic difference could make a significant difference in practice, I'd look into the linked list implementation you suggest. (We might want to look into [immutable datastructures](https://facebook.github.io/immutable-js/) more generally at some point. Created issue #174.) I think it's substantially better to only have one `Trace` datastructure, even if it comes at a small performance cost.

> >  when concatenating the lists, have the completed ones come first?

> Why? (I assume this helps but I don't see how - sorry.)

I don't think it makes a different computationally, it just felt like the more intuitive ordering to me (not a strong preference).

null-a:
> One way of doing this is to use kernel combinators,

Done.

I wasn't sure how `sequence` and `compose` differed in your example, in particular I'm not sure what type you expected the 2nd argument to `compose` to have?

In my version `compose(x, y) == sequence(x, y)` which I don't think is what you had in mind.

>> SMC could become something like InferSMC.

> Sounds good.

I've renamed `SMC` and `MCMC` to `MarginalSMC` and `MarginalMCMC`. This fits with Noah's original suggestion (#86) for the new inference interface and hints that the return type is a marginal distribution.

`ParticleFilter` is now `SMC`.

> In general, it would be great if there were clear rules on (1) what kind of functionality goes into InferX vs X (in particular for X = SMC) and (2) how the interface (arguments, return value) differs between InferX and X

I agree, although since we only have `MarginalSMC` and `SMC` it seems a little premature to try and pin down the general interface for such pairs. Perhaps we'll see a way to restructure `SMC` and dissolve the distinction, or perhaps it will be clearer what the right abstractions are when we port more algorithms to the new interface?

> Having either lag or skip with skip semantics seems fine to me.

`lag` now has `skip` semantics. I've not changed `IncrementalMH`.

> Doesn't it make sense to have all sampling-based algorithms support computing the MAP?

I was thinking of the particle filter in particular. In this case we don't benefit much from the strategy of only hanging onto the MAP sample as we have all the particles in memory at the end of inference anyway, so I thought that perhaps `ERP.prototype.MAP` or its successor (#140) could handle this.

Having thought some more I suspect you're right though.

> use MH's local search to find a non-zero trace

Ah, got it! I hadn't realized that `MH` and `HashMH` had different initialization strategies. Are you happy with us only having support for the 'repeatedly run the program from the start until we find a trace with non-zero probability' strategy and not the local search strategy?

>> I think SMC has different asymptotics with Trace and TraceLite.

> It would be good to verify this.

A crude benchmark suggests it does. This is the run-time in seconds of `SMC` on the model from `hmmIncremental` with a fixed number of particles (1K) and no rejuvenation.

Number of states | Trace | TraceLite
----|----|----
50  | 3.2  | 0.8
100 | 10.3 | 1.7
150 | 20.8 | 2.4
200 | 36.2 | 3.6

null-a:
> A crude benchmark suggests it does.

This chart is a bit more suggestive that there might be something non-linear happening with `Trace`.

plot

I should convince myself that this isn't a manifestation of #150.

stuhlmueller:
> In my version compose(x, y) == sequence(x, y) which I don't think is what you had in mind.

Oops, I didn't think this through. We only need one of `compose` and `sequence`. Then we can simplify the chain definition a bit:

    var chain = sequenceKernels(
        repeatKernel(options.burn, kernel),
        repeatKernel(options.samples,
            sequenceKernels(
                repeatKernel(options.lag + 1, kernel),
                collectSample)));

> I agree, although since we only have `MarginalSMC` and `SMC` it seems a little premature to try and pin down the general interface for such pairs. Perhaps we'll see a way to restructure `SMC` and dissolve the distinction, or perhaps it will be clearer what the right abstractions are when we port more algorithms to the new interface?

Yeah, I agree that it's not very important to have generalizable rules for this now. However, if there are two functions `MarginalSMC` and `SMC` in the final version of this pull request, we need to be able to tell which functionality should go where, and at the moment, I'm not clear on what the distinction is.

> `lag` now has `skip` semantics. I've not changed `IncrementalMH`.

I'd change it there, too. Inconsistency seems worse than surprising (but globally consistent) semantics.

> Are you happy with us only having support for the 'repeatedly run the program from the start until we find a trace with non-zero probability' strategy and not the local search strategy?

Yes. For the most part, it's not a good idea to write models that are very difficult to initialize anyway (they tend to be the kinds of programs where MCMC doesn't mix well either), and with the new modular framework, we can always add smarter initialization strategies later.

> This chart is a bit more suggestive that there might be something non-linear happening with `Trace`.

Interesting! In that case, it would be great to figure out whether we can make changes to `Trace` that fix this without resulting in much overhead for use in plain MCMC.

null-a:
> if there are two functions MarginalSMC and SMC in the final version of this pull request, we need to be able to tell which functionality should go where, and at the moment, I'm not clear on what the distinction is.

`MarginalSMC` *almost* exclusively deals with interfacing with the user. i.e. Setting default options and building the marginal ERP from the particles returned by `SMC`. The only wrinkle is that final rejuvenation happens in `SMC` which is annoying. Is this the part that troubles you too?

> it would be great to figure out whether we can make changes to Trace that fix this without resulting in much overhead for use in plain MCMC.

I tried the obvious thing of switching `Trace` to use `List` and `Map` from ImmutableJS. It helps close the gap to `TraceLite` for SMC without rejuvenation though it still might not scale in the same way:

plot

For moderately sized models it looks like this persistent version of `Trace` slows MCMC down by a factor of about 2.

stuhlmueller:
> The only wrinkle is that final rejuvenation happens in `SMC` which is annoying. Is this the part that troubles you too?

Yes. If all `MarginalSMC` did (beyond calling `SMC`) was to set some defaults and build the final ERP, I'd be happy with it.

stuhlmueller:
> I tried the obvious thing of switching `Trace` to use `List` and `Map` from ImmutableJS. It helps close the gap to `TraceLite` for SMC without rejuvenation though it still might not scale in the same way:

I'm not sure yet what we should do about this, but based on your graph, `PersistentTrace` still results in too much of a slowdown, in particular if the asymptotics are different. If so, it would be good to know what's causing this. One possibility is that it's due to operations in ImmutableJS being `O(log32 N)`, not `O(1)`, although `O(log32 N)` grows pretty slowly, too.

Another idea: What if we used `Trace`, but set a parameter in `SMC` (based on `this.rejuvSteps === 0`) to control whether to store random choices? Would that help?

null-a:
> If all MarginalSMC did (beyond calling SMC) was to set some defaults and build the final ERP, I'd be happy with it.

Resolving this factoring issues while retaining anything resembling compositional pieces appears non-trivial, so for now I've [reverted](https://github.com/null-a/webppl/commit/85954d1c9a9e2bbc3d61cd7fa6c20979f66e0d84) to the monolithic coroutine set-up. I think we should tackle doing anything better as a separate piece of work.

> in particular if the asymptotics are different ... it would be good to know what's causing this.

Looking at the numbers I'm not sure that even `TraceLite` is linear, but that might be a manifestation of #150. You're right that we pick up a `O(log32 N)` term in `PersistentTrace` though.

> What if we used Trace, but set a parameter in SMC (based on this.rejuvSteps === 0) to control whether to store random choices? Would that help?

This is exactly what `TraceLite` achieves, but using polymorphism rather than conditionals. I've [switched](https://github.com/null-a/webppl/commit/8ff6afdee9cd9ad598fdd2de5b9e2d34189ee71b) to your suggestion so you can see which you prefer.

null-a:
The previous version of `ParticleFilterRejuv` set `forwardChoiceScore = importanceScore` [at each random choice](https://github.com/probmods/webppl/blob/fca4775a6bb0b9a6baa64e1e73ee42fd2b49fb80/src/inference/smc.js#L82) encountered in the particle filter.

The refactored version doesn't do this, instead we wait until we compute the acceptance probability (during rejuvenation) to look-up the score, which I think would equivalent to setting `forwardChoiceScore = choiceScore` at each choice in the previous implementation.

So there's a change in behavior in this refactoring. I *think* the refactored version is correct - do you agree?

stuhlmueller:
> Resolving this factoring issues while retaining anything resembling compositional pieces appears non-trivial, so for now I've [reverted](https://github.com/null-a/webppl/commit/85954d1c9a9e2bbc3d61cd7fa6c20979f66e0d84) to the monolithic coroutine set-up. I think we should tackle doing anything better as a separate piece of work.

Fine with me. In that case, do we even need to call the functions `MarginalMCMC` and `MarginalSMC` for now? We could just use `MCMC` and `SMC` until we factor out the user-facing parts.

> I've [switched](https://github.com/null-a/webppl/commit/8ff6afdee9cd9ad598fdd2de5b9e2d34189ee71b) to your suggestion so you can see which you prefer.

I think your previous solution was a little better -- however, this one isn't quite what I had in mind. I was hoping that we could use your previous version of `Trace` as it was and just not add any choices in `SMC` if we don't do rejuvenation. That is, `Trace` still builds the `choices` and `addressMap` data structures, but they remain empty. This solution (if it works) appeals to me because almost no extra code is required, which makes it easier to understand and maintain.

stuhlmueller:
> So there's a change in behavior in this refactoring. I *think* the refactored version is correct - do you agree?

I think it's right, too. My understanding is that this only mattered for the first rejuvenation step, and only if no `reverseChoiceScore` was available (i.e. no custom proposer). In that case, there was a question whether to use prior score (`choiceScore`) or importance score for the probability of going back to the first trace. The previous implementation used the importance score, but now I don't think that's correct -- in this situation, the actual probability of making a proposal to go back to the first trace is the prior.

null-a:
> That is, Trace still builds the choices and addressMap data structures, but they remain empty.

You're right. We still have to copy these empty data structures but that costs virtually nothing and the scaling is correct. Sorry I didn't catch on to what you were suggesting sooner!

> We could just use MCMC and SMC until we factor out the user-facing parts.

Agreed.

null-a:
I don't have any more changes planned for this PR. I've had a another read through the code, pull request and the original issue (#64), and I think all the loose ends have been tied up. Can you think of anything else we need to do before we merge it?

stuhlmueller:
I'm pretty happy with the PR, too. Still, reading over the code again, I generated some more comments (some of which can be dealt with by creating issues):

- Suppose I call the (backwards-compatibility) function `MH` with `iterations = 100`, `burn = undefined`. Then it looks like `burn` will be set to 50, and I only get 50 samples back. This seems surprising—I'd expect to get 100 samples.
- There shouldn't be any `TODO`s in the source code, only in the issue tracker.
- Does the way we use `env.query` play nice with nested queries? I suspect it might not, in which case we should either resolve this (perhaps by treating it similarly to how we uninstall and reinstall coroutines at the moment), or at least make an issue for it.
- Also, it's a little surprising that we call `env.query.clear();` in `Initialize`, but don't use `query` there otherwise. This is perhaps what your `TODO` is referring to. I'm not sure where `env.query.clear();` should happen, so perhaps this should just be another issue for now.
- Having to set `incrementalize` in each coroutine feels wrong (e.g., `Initialize.prototype.incrementalize = env.defaultCoroutine.incrementalize;`). One way around this would be to create all coroutines by extending `env.defaultCoroutine`. This could happen when we implement #173.
- We should standardize on either always using the full filename in `require`s, or leaving out the `.js`. E.g. `var Query = require('../query.js').Query;` vs `var aggregation = require('../aggregation');`. I weakly prefer the style without `.js`, but would be happy with consistent use of the other style, too.
- `MCMC` still doesn't feel entirely readable to me, which is probably in large part because everything needs to be in CPS style. Perhaps it could be improved a little bit by assigning names to the anonymous functions in lines 23 (`run`?) and 43 (`finish`?).
- I think it would be clearer to pass the optional arguments to `MHKernel`—everything except `k` and `oldTrace`—as a single `options` argument. Things like `_.partial(MHKernel, _, _, _, _, options.particles === 1);` are difficult to interpret.
- `smc.js`, line 61: The comment is a little misleading (it's the other case where choices are not required).
- `smc.js`, line 197: Split into two lines?
- `smc.js`, line 234:
   
          if (particle.trace.score === -Infinity) {
            // Can happen with one particle as we don't resample to allow
            // ParticleFilterAsMH.
            throw 'Particle score is -Infinity';
          }
    
    I'm confused by the comment—if we throw an error here, then doesn't this sometimes make it impossible to use SMC with one particle?
- `withImportanceDist` - maybe we should move this to `erp.js` (since we moved the drift kernel ERPs there, too)
- In `util.js`, unify the argument naming of `cpsLoop` and `cpsIterate`?

null-a:
> This seems surprising—I'd expect to get 100 samples.

I agree. I didn't carry that behavior over into `MCMC` but that's how the current `MH` works. I'm happy to be less backwards compatible if you don't think it will trip anyone up.

Related: I've not added a function for backwards compatibility with `HashMH`. Would that be useful?

> There shouldn't be any TODOs in the source code

Thanks for the reminder! I've removed the `TODO` notes from this PR. Would it be worth me turning the remaining `TODO`/`FIXME` comments in the code base into issues? (Other than perhaps those in the variational branch.)

> Does the way we use env.query play nice with nested queries?

I've always assumed not. I've opened #199.

> it's a little surprising that we call env.query.clear(); in Initialize

I agree. The idea in the `TODO` is to change the contract between `MCMC` and the initialization routines. Instead of having e.g. `Initialize` return a trace with `score > -Inf` the idea would be to have it run the program just once, and leave `MCMC` to check the `score` and re-run if necessary. This would allow us to move `env.query.clear()` into `MCMC`. It looks something like [this](https://github.com/null-a/webppl/commit/2160fecc7a14394a23e02dbf62bc61ec4dfa231e#diff-369ffd72ce07b6a6f35eb56e47fd40a4R22).

However, having thought about this some more, I'm now not even sure that clearing the query is strictly necessary here, but this needs more thought. I've opened #198.

> Having to set incrementalize in each coroutine feels wrong

Indeed. I agree it's a separate issue, I've added a note to #173.

> We should standardize on either always using the full filename in requires, or leaving out the .js.

Since we're inconsistent across the whole code base and not just in this pull request I've opened #200.

> MCMC still doesn't feel entirely readable to me, which is probably in large part because everything needs to be in CPS style.

I agree, and CPS is definitely part of it.

I tried your suggestion [here](https://github.com/null-a/webppl/blob/83405ffc8d3eaaba4fb7d6c009cc953e3630ae1c/src/inference/mcmc.js#L11), I hope I understood you correctly. The thing I prefer about the anonymous function approach is that execution proceeds in the same order as you read the code - top to bottom. (I could swap `run` and `finish` around but the linter complains, incorrectly I suspect.)

This suggests keeping the names and recovering the readable ordering, something like [this](https://github.com/null-a/webppl/blob/e3c006715edb74bc7d7b308bb2126617eba23334/src/inference/mcmc.js#L11) perhaps? The only snag here is that this doesn't fit well with the way `Query` works. To make that work my hunch is that we'd need to view kernels and the other functions we're composing as functions of type `(Trace, Query) -> (Trace, Query)`.

Once we *do* have a way to deal with `Query`, perhaps we're not too far away from been able to write `MCMC` in WebPPL.

> I think it would be clearer to pass the optional arguments to MHKernel—everything except k and oldTrace—as a single options argument

Done. The only draw back is that [partially applying arguments](https://github.com/null-a/webppl/commit/b7480259ae5ee42d0a24fd50cba69ba796dcc9de#diff-f12e65f11f15a71a45a5877bdff3a5e5L17) inside of `options` is a bit messy.

I suppose we could have a `partiallyApplyOptions` function, although looking at this again I'm not sure `SMC` should even be concerned with `MHKernel` and `permissive` mode and all that. Perhaps if you want to do something like `ParticleFilterAsMH` we should expect you to call something like `SMC(model, { rejuvKernel: MHKernel({ permissive: true }), particles: 1 })`. Of course, this won't work at the moment as `MHKernel` takes a continuation and a trace rather than an options object. Perhaps we should delay the general question of how to parameterize kernels until #86.

>  line 61: The comment is a little misleading

Good catch, thanks!

> line 197: Split into two lines?

Done.

> if we throw an error here, then doesn't this sometimes make it impossible to use SMC with one particle?

I think this was left over from before I relaxed MH by adding permissive mode. In any case, I've now removed it.

> withImportanceDist - maybe we should move this to erp.js

> unify the argument naming of cpsLoop and cpsIterate

Done.

stuhlmueller:
> > This seems surprising—I'd expect to get 100 samples.
> 
> I agree. I didn't carry that behavior over into `MCMC` but that's how the current `MH` works. I'm happy to be less backwards compatible if you don't think it will trip anyone up.

I'd be less backwards compatible here. I don't think a change will cause issues for anyone, and keeping the old interpretation of `burn` will just result in people getting used to it.

> Related: I've not added a function for backwards compatibility with `HashMH`. Would that be useful?

I don't think it's necessary. For the few people who used this, it's straightforward to replace the calls with `MH` when the program fails to run.

> Would it be worth me turning the remaining `TODO`/`FIXME` comments in the code base into issues?

Yes, that would be good.

> Instead of having e.g. `Initialize` return a trace with `score > -Inf` the idea would be to have it run the program just once, and leave `MCMC` to check the `score` and re-run if necessary.

This seems better with respect to `env.query.clear()`, but the interface where `Initialize` returns a trace with non-zero probability is more natural otherwise, so should be kept.

> The thing I prefer about the anonymous function approach is that execution proceeds in the same order as you read the code - top to bottom. (I could swap `run` and `finish` around but the linter complains, incorrectly I suspect.)

How about something like [this](https://gist.github.com/stuhlmueller/1ee91b12b2f940509882)?

> To make that work my hunch is that we'd need to view kernels and the other functions we're composing as functions of type `(Trace, Query) -> (Trace, Query)`.

I think kernels are generally expected to be `Trace -> Trace`, but maybe the query object should be stored within the trace?

> Once we *do* have a way to deal with `Query`, perhaps we're not too far away from been able to write `MCMC` in WebPPL.

That would be great!

> I'm not sure `SMC` should even be concerned with `MHKernel` and `permissive` mode and all that. 

I tend to agree that it shouldn't.

> Perhaps if you want to do something like `ParticleFilterAsMH` we should expect you to call something like `SMC(model, { rejuvKernel: MHKernel({ permissive: true }), particles: 1 })`. Of course, this won't work at the moment as `MHKernel` takes a continuation and a trace rather than an options object. Perhaps we should delay the general question of how to parameterize kernels until #86.

I'd be okay with breaking this functionality until we address #86. I think it's only used in tests at the moment (but I expect that we might want to use it for actual applications in the future).

null-a:
> I'd be less backwards compatible here.

Great.

I think it now makes sense to view `MH` etc. as simple wrappers around `MCMC` which pass on their arguments without modification. I've made this change.

If we want to retain the behavior of setting a default for `burn` based on the number of samples then I'll need to add that to `MCMC`. At present `burn` always defaults to zero.

>> The thing I prefer about the anonymous function approach is that execution proceeds in the same order as you read the code - top to bottom.

> How about something like this?

Sorry, I did an awful job of explaining why I wasn't too keen on that approach. My concern is that you have to look closely to see that `initialize -> run -> finish` are chained together, which is what I tried to make explicit with `cpsSequence(initialize, run, finish)`. For me, this doesn't seem to be a problem when using anonymous functions. I think that's because I read that approach as a regular procedure, and think of the anonymous functions as simply (lexically) binding intermediate values. This seems easy to think about.

Having said all that, I've switched to your approach. If you think it's more readable then that's good enough for me! I'm quite possible too familiar with the code to make a good judgement here.

> I'd be okay with breaking this functionality until we address #86. 

I think we should either:

1) Leave it as is for now, then refactor `SMC` so it doesn't touch `MHKernel` etc. after #86.
2) Remove all of the stuff relating to single particle PF as MH.

In particular, I don't think we should break the functionality and then remove the tests, leaving bits of the implementation around to rot.

Since what we have isn't all that bad, I'd go with option 1.

stuhlmueller:
I just realized that the MCMC code can be simplified a little bit more—see [here](https://gist.github.com/stuhlmueller/1ee91b12b2f940509882).

stuhlmueller:
Merged. Thanks for the hard work, and for putting up with all my nitpicky comments!

null-a:
> Thanks for the hard work, and for putting up with all my nitpicky comments!

No problem! Your feedback has made this much better than it would have been otherwise - thanks!

PR #170, opened 2015-08-26
Handle empty return statements

### LINKED ISSUE #154: CPS doesn't handle 'empty return'
This program:
````
var foo = function() {
	if (flip(0.5))
		return;
	globalStore.count = globalStore.count + 1;
	foo();
};

MH(function() {
	globalStore.count = 0;
	foo();
	return globalStore.count;
}, 1000);
````
produces this error
````
/Users/dritchie/Git/webppl/src/transforms/cps.js:170
  switch (node.type) {
              ^
TypeError: Cannot read property 'type' of null
    at cps (/Users/dritchie/Git/webppl/src/transforms/cps.js:170:15)
    at Array.<anonymous> (/Users/dritchie/Git/webppl/src/transforms/cps.js:315:14)
    at Array.<anonymous> (/Users/dritchie/Git/webppl/src/syntax.js:30:27)
    at loop (/Users/dritchie/Git/webppl/src/syntax.js:50:24)
    at /Users/dritchie/Git/webppl/src/syntax.js:51:16
    at Array.<anonymous> (/Users/dritchie/Git/webppl/src/syntax.js:35:16)
    at loop (/Users/dritchie/Git/webppl/src/syntax.js:50:24)
    at /Users/dritchie/Git/webppl/src/syntax.js:51:16
    at Array.<anonymous> (/Users/dritchie/Git/webppl/src/syntax.js:35:16)
    at loop (/Users/dritchie/Git/webppl/src/syntax.js:50:24)
````
There's a workaround by using `return undefined;` instead of just `return;`, but IMO webppl should probably handle this better.

null-a:
Fixes #154.

### DIFF
--- a/src/transforms/cps.js
+++ b/src/transforms/cps.js
@@ -275,7 +275,7 @@ function cpsInnerStatement(node, e, fk) {
       });
     }),
     clause(Syntax.ReturnStatement, function(argument) {
-      return cps(argument, fk);
+      return cps(argument || build.identifier('undefined'), fk);
     }),
     clause(Syntax.VariableDeclaration, function(declarations) {
       return cpsDeclarations(declarations, 0, function(id) {
@@ -312,7 +312,7 @@ function cpsFinalStatement(node, k, fk) {
       });
     }),
     clause(Syntax.ReturnStatement, function(argument) {
-      return cps(argument, fk);
+      return cps(argument || build.identifier('undefined'), fk);
     }),
     clause(Syntax.VariableDeclaration, function(declarations) {
       return cpsDeclarations(declarations, 0, function(id) {
--- a/tests/test-transforms.js
+++ b/tests/test-transforms.js
@@ -1,5 +1,6 @@
 'use strict';
 
+require('../src/main');
 var _ = require('underscore');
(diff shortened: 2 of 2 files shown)


stuhlmueller:
+require('../src/main');

    Requiring main.js adds apply as a global as expected by some tests.

Does this mean that some transform tests didn't work previously?

null-a:
> Does this mean that some transform tests didn't work previously?

No. It was specifically `nodeunit tests/test-transforms.js` that didn't work. i.e. Running just the transform tests outside of `grunt`.

stuhlmueller:
Ok, thanks!

    var foo = function(){ return; return; }; foo()

Change the test to 

    var foo = function(){ return; return 2; }; foo()

so that we check that we actually return when we encounter the first return statement?

null-a:
The reason I had two empty `return` statements in the test was to exercise the changes I made in both `cpsInnerStatement` and `cpsFinalStatement`. I see now how that's confusing, so I've updated the test to try and clarify what I'm testing. Does that look better?

stuhlmueller:
Ah, I see, thanks! Looks good now!

PR #171, opened 2015-08-31
Infrastructure for testing deterministic webppl code

stuhlmueller:
I'm proposing to add this so that we can add tests for pull requests #157 and #162 and get them merged.


null-a:
Looks good!

I made a few tiny changes, including:

* I removed the `try`/`catch` around `webppl.run` in both inference & deterministic tests. I think I originally added this to the inference tests because node unit was reporting success/failure but not the detail of any exceptions thrown. I'm not sure why, but exceptions are now reported without this.
* I've tweaked the assertion error:

Before:

```
>> basic
>> Message: Expected basic: 4, actual: 5
```

After:

```
>> basic
>> Message: Expected result: 4, actual: 5
```

Providing you're happy with my tweaks, I think this is good to merge.

stuhlmueller:
Thanks for the tweaks!

Issue #175, opened 2015-09-07
For complete enumeration, default to depth-first search


stuhlmueller:
DFS tends to use less memory

PR #185, opened 2015-09-10
Default to depth first search for complete enumeration

null-a:
This addresses #175. Inference tests exercise both paths through `Enumerate`.


stuhlmueller:
I think it would be a little cleaner to move the new function out of the return object and to the top-level, like `enuFifo` and `enuFilo`. Call it `enuDefault` or something like that?

null-a:
Sure, done!

Issue #165, opened 2015-08-17
Make all arguments to ERP constructor named arguments?


stuhlmueller:
I think the ERP definitions would look nicer if we used

    var myERP = new ERP({
      sampler: function(){...},
      scorer: function(){...},
      support: function(){...}
    });

instead of

    var myERP = new ERP(
      function(){...},
      function(){...},
      {
        support: function(){...}
      }
    );

The former is more uniform and makes it clearer what the role of each function arguments is (although named functions help with the latter, too).

PR #186, opened 2015-09-10
Use named arguments for ERP constructor

null-a:
I agree with the suggestion in #165 that this looks a little nicer, it's also a little more straight forward I think.

I've also added a few extra tests to exercise some ERP which aren't exercised by existing inference tests.


stuhlmueller:
I like the look! 

Should `ERP` assert that `sample` and `score` are present?

null-a:
I considered it, but I think I decided not to because I imagined such a mistake would show up pretty quickly and it would be pretty obvious what had happened. (Obvious compared to say subtle inference bugs.)

I'm not at all against it though if you think it might be useful. Perhaps it would be most useful as documentation, in which case it might be an idea to combine the assertions with the interface documentation we have in the comments?

stuhlmueller:
If it doesn't slow things down, I don't see a reason not to add the assertions. Ideally, we'd have assertions here and elsewhere that are enabled in debug mode, but that get compiled away in the version of webppl that makes it to users.

I'd leave the comments at the top of the file as they are, and add assertions to `ERP` as local reminders, but if you think some other combination of comments and assertions would be clearer, feel free to choose another approach.

null-a:
>  I don't see a reason not to add the assertions

Agreed, done.

>  Ideally, we'd have assertions here and elsewhere that are enabled in debug mode

Indeed.

iffsid:
If we're changing the erps to have anonymous functions implementing `sample` and `score`, could we find an alternate way to get erp identity?
Debugging anything even slightly complex becomes a nightmare otherwise, when you can't tell things like exactly which erp MCMC proposals were made to, or what the last encountered erp before a crash was, etc.

stuhlmueller:
I wouldn't be opposed to associating names with ERPs, e.g.

    var gammaERP = new ERP({
      name: 'Gamma',
      sample: function(){...},
      score: function(){...}
    });

We could even automate this by iterating over the keys and values of the object exported by `erp.js`, and adding the name property based on the key if the corresponding value is an instance of `ERP`.

null-a:
> I wouldn't be opposed to associating names with ERPs

I'll take care of this.

Issue #187, opened 2015-09-10
Default parameters for ERP?


null-a:
I've been caught out a few times by the fact that `flip()` is equivalent to `flip(0)`rather than `flip(0.5)`.

Does it make sense to set default parameters for `bernoulliERP` and perhaps other ERP?

stuhlmueller:
I agree that `flip()` should be equivalent to `flip(0.5)`. I'm not sure we need ERP default parameters for that, in particular since there aren't that many ERPs where we would want this behavior. Would it be enough to just make the wrapper functions (such as `flip` and `gaussian` in `header.wppl`) provide default arguments to the ERP if insufficiently many arguments are given?

null-a:
Yes. I don't feel a strong expectation that `sample(bernoulliERP, [])` or perhaps `sample(bernoulliERP)` would be equivalent to `sample(bernoulliERP, [0.5])` so handling this in the wrappers seems reasonable to me.

null-a:
Closed by #195.

PR #189, opened 2015-09-14
Specify default params for sampling convenience functions.

null-a:
As discussed in #187.


null-a:
Some analysis tests are failing because `flip` now generates member expressions which aren't handled by `analyze`.

@stuhlmueller Are you happy for me to comment these tests out for now, or should we update `analyze`?

stuhlmueller:
Yes, comment out these tests for now -- I don't think anyone is using `analyze` at the moment.

stuhlmueller:
It would be good to have a test that checks that (at least one of the) header functions without parameters does the right thing. The implementation relies on `arguments`, which I implemented in a rather fragile way, so it's easy to imagine that the implementation breaks in the future. Now that I think about it, maybe it would be better to implement this by checking whether `theta` etc. are undefined.

null-a:
> Now that I think about it, maybe it would be better to implement this by checking whether theta etc. are undefined.

Done.

Seeing a bare `gaussian()` in a model made me question whether we even want default arguments for `gaussian`. (And to a lesser extent `uniform`.) Is it sufficiently clear what `gaussian()` means? An alternative would be to have something like a new `stdGaussian` function. This isn't any shorter but it might be convenient if you need to pass it to a higher-order function. (Though I don't think that happens often, so that point is perhaps moot.)

Thinking about *that* idea made me realize that `repeat(3, flip)` will always return `[false, true, true]` because `repeat` passes the current iteration to its second argument. I find this surprising.

If you're happy with the PR as it is, then so am I. If we're unsure, I suggest we just set a default for `flip` which I still feel strongly about.

iffsid:
> `repeat` passes the current iteration to its second argument

This was introduced in 2876516d89c7dcc9f61cf28987c7eb56c4591fd8.
`repeat` should repeat the given thunk `n` times. It shouldn't be adapting the function called according to the iteration.
It appears to have been introduced to replicate functionality provided by `mapN`, which handles this correctly. We should revert this.

stuhlmueller:
> If we're unsure, I suggest we just set a default for `flip` which I still feel strongly about.

Let's just do that for now, and also revert `repeat` as Sid says.

null-a:
Superseded by #195.

PR #194, opened 2015-09-15
Have ERP.prototype.print show the ERP name for non-serializable ERP

null-a:
We could do something better than this of course, but this is better than throwing an exception I think.


stuhlmueller:
Make the case ordering in `toJSON` and `print` analogous, or extract the test to a `serializeable` method? (This would make it clearer on first glance that the cases in the two functions dispatch on the same property.)

null-a:
Good call. Done.

PR #214, opened 2015-09-22
Test/fix that inference should leave store unmodified

### LINKED ISSUE #112: State of the store after inference
We appear to be inconsistent about the state in which we leave the store after performing inference in a model which modifies the store.

For example, with this model:

```
var model = function() {
  globalStore.foo = 1;
  flip(0.5);
  globalStore.foo += 1;
  return 1;
};
```

... we get:

```
MH(model, 10);
display(globalStore);
# => { foo: 2 }
```

```
ParticleFilterRejuv(model, 10, 10);
display(globalStore);
# => {}
```

In general the particle methods leave the store unchanged, the others don't. (IncrementalMH is an exception I think.)

Should we be consistent about this, and if so what is the correct behavior?

(Without a principled way to choose between the various states of the store that we might encounter across multiple executions of the model, leaving the store unmodified seems reasonable to me. It seems likely there's a better argument than this though.)

`Enumerate` leaves the store in the state it was in halfway through an execution:

```
Enumerate(model, 10, 10);
display(globalStore);
# => { foo: 1 }
```

I imagine this needs changing even if we leave everything else as it?

null-a:
Fixes #112.


stuhlmueller:
Thanks! I think we may be able to simplify the test code slightly, but it's a separate task, so I created a new issue: #215.

PR #220, opened 2015-09-23
Link to docs rather than README.

null-a:
The intention here is to link directly to the section we want to refer people to.

I wonder whether long-term we should consider merging the contributing guide with the workflow section in the docs - they seem pretty similar,


stuhlmueller:
> I wonder whether long-term we should consider merging the contributing guide with the workflow section in the docs

We could consolidate it somewhat, but we'll still need a `CONTRIBUTING ` file, because it receives special treatment from github (it's linked to prominently when people create issues and pull requests).

PR #221, opened 2015-09-24
Grab version/release from package.json.

null-a:
I noticed that version & release in `docs/conf.py` were out of date so I've switched to grabbing these from `package.json`. This works for me locally if I do `make html` from within `webppl/docs`. `json` is part of the  Python standard library so I assume it'll be ok on RTD too, but I'm not sure.

The version number shows up in the HTML title tag of the docs, not sure if it's anywhere else.


stuhlmueller:
Neat. I expect that it'll work on RTD, too. Revert my changes to `docs/development/npm.rst` as part of this PR, since the version adjustment won't be necessary anymore?

null-a:
> Revert my changes to docs/development/npm.rst

Thanks - I'd not spotted that!

Issue #140, opened 2015-07-03
Improve MAP Estimates


null-a:
The way we compute the MAP in `makeMarginalERP` doesn't work very well when the marginal comes from a sampling algorithm. This is because we pick the sample with the highest count as the MAP value, but in large enough spaces all counts are 1.

We can do better than this by looking at each sample's score (where available) - `HashMH` and `IncrementalMH` already do this.

There are a couple of things we can improve:

1. `HashMH` and `IncrementalMH` overwrite the `MAP` *function* on the marginal ERP with an *object*, which makes the interface inconsistent.
2. More generally, we could try to refactor the MAP computation to reduce duplication etc. (Related: #64.)

iffsid:
The reliance on counts is indeed unpleasant.
One way around this is to instrument the histogram creation with objects that not only record the value and count (called `prob` in the object), but also the score (appropriately incremented) for each trace/particle as it is added to the histogram.

The reason I put in the computations as functions is that the computation of _erp-specific_ values, like the MAP estimate or entropy, i.e., those that don't change w.r.t anything but the erp itself, can be easily memoised. 
The MAP and entropy methods implemented as prototypes for the ERP class/type compute the requisite values once and then modify the local method to just return that value.

mhtess:
Possibly related. Would it be possible to implement a Highest Posterior Density interval estimator for ERPs? This would be tremendously useful for bayesian data analysis. 

Ideally it would take the arguments of the interval, defaulting to 95%.  e.g.. `myERP.HPD([0.025,0.975])`

ngoodman:
sounds like a great thing for you to implement @mhtessler! ;)


-N

null-a:
> We can do better than this by looking at each sample's score

I've realised that this isn't always the case, as doing so doesn't account for marginalization.

For example:

```javascript
function() {
  var a = flip();
  var b = flip();
  return a + b;
}
```

With this model all samples have the same score, so the MAP estimate (from `HashMH` say) will always be the first sample taken. This means the distribution on MAP estimates is the same as the marginal of the model which seems odd. (You may as well take a single sample and use that instead.) In this case then, the approach taken in `ERP.prototype.MAP` has some utility.

So both approaches are useful, but the best approach doesn't coincide with whether or not an MCMC method is used. We perhaps need a straight-forward interface to present both options to the user.

[...]

PR #222, opened 2015-09-29
Fix IncrementalMH justSample option

### LINKED ISSUE #160: IncrementalMH with justSample option throws error
As [reported](https://groups.google.com/forum/#!msg/webppl-dev/B0RNLrS_cws/4FQx2ZATBAAJ) by @mhtess.

It looks like this is caused by attempting to create an empty marginal ERP which now triggers an assertion failure. ([Relevant change](https://github.com/probmods/webppl/commit/b5253da0f4cff94b5be518d5e44d2476c057ba41).)

If we want to continue to not allow empty marginal ERPs we could perhaps make an ERP from the MAP estimate to be somewhat faithful to the inference interface. Thoughts?

We should also add inference tests for some of these options.

HashMH will also have this bug.

null-a:
Since b5253da `makeMarginalERP ` hasn't accepted an empty histogram, so IncrementalMH with these options now returns a marginal distribution which is a delta distribution on the MAP estimate. Fixes #160.

### DIFF
--- a/src/inference/incrementalmh.js
+++ b/src/inference/incrementalmh.js
@@ -934,11 +934,14 @@ module.exports = function(env) {
         }
       }
     } else {
-      var dist;
-      if (this.returnHist)
-        dist = erp.makeMarginalERP(util.logHist(this.returnHist));
-      else
-        dist = erp.makeMarginalERP({});
+      var hist;
+      if (!this.returnSamps && !this.onlyMAP) {
+        hist = this.returnHist;
+      } else {
+        hist = {};
+        hist[JSON.stringify(this.MAP.value)] = { prob: 1, val: this.MAP.value };
+      }
+      var dist = erp.makeMarginalERP(util.logHist(hist));
       if (this.returnSamps) {
         if (this.onlyMAP)
           this.returnSamps.push(this.MAP);
--- a/src/inference/incrementalmh.js
+++ b/src/inference/incrementalmh.js
@@ -935,11 +935,11 @@ module.exports = function(env) {
       }
     } else {
       var hist;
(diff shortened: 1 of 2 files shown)


stuhlmueller:
if (!this.returnSamps && !this.onlyMAP) {

Replace with `if (this.returnSamps || this.onlyMAP) {` and switch the case order? It's easier to think about.

For `MAP` tests, use a non-deterministic model (such as `simple.wppl`) so that we can verify that it actually only returns the MAP value?

null-a:
> Replace with if (this.returnSamps || this.onlyMAP) { and switch the case order? It's easier to think about.

Ha, I very nearly did that! I only elected to go this way so that the common case was in the consequent. I'm happy to switch it around though.

> For MAP tests, use a non-deterministic model (such as simple.wppl) so that we can verify that it actually only returns the MAP value?

My thinking here was that I'd worry about that correctness of MAP estimates when I look at #140. Not only because we don't really compute the MAP of the marginal, but also because we overwrite the MAP *function* on the marginal ERP with a *property* making it messy to test across all algorithms. I thought there was some value in exercising the code though, hence it's inclusion.

If you'd prefer I added anyway, then of course I'm happy to do so!

stuhlmueller:
> Ha, I very nearly did that! I only elected to go this way so that the common case was in the consequent. I'm happy to switch it around though.

In this case making the predicate easy to understand makes a bigger difference in readability to me, but it's a judgment call.

> My thinking here was that I'd worry about that correctness of MAP estimates when I look at #140.

Fine with me.

null-a:
> Replace with if (this.returnSamps || this.onlyMAP) { and switch the case order? It's easier to think about.

Done.

PR #223, opened 2015-09-29
Use seedable PRNG for inference tests.

### LINKED ISSUE #143: Record or fix RNG seed for travis tests

null-a:
Each inference test is run with `Math.random` seeded with a particular value. This value changes across runs of the test suite, but within a run is the same for all inference tests. The random seed is logged to the console when the tests run.

If one (or more) tests fail you can replicate the failure without running the entire test suite by running something like:

````
RANDOM_SEED=1033080963 nodeunit tests/test-inference.js -f 'MH - simple'
````

The only downside I can think of with this approach is that we're not testing inference with V8's `Math.random` at all, so there's a chance that we could miss something. I'm not sure this is very likely to happen, but if we were concerned we could set things up so we do an extra run of the test suite on Travis without the seedable RNG.

Closes #143.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "grunt-contrib-nodeunit": "^0.4.1",
     "grunt-gjslint": "^0.1.6",
     "nodeunit": "^0.9.1",
+    "seedrandom": "^2.4.2",
     "through2": "^2.0.0"
   },
   "scripts": {
--- a/src/erp.js
+++ b/src/erp.js
@@ -128,7 +128,7 @@ var deserializeERP = function(JSONString) {
 
 var uniformERP = new ERP({
   sample: function(params) {
-    var u = Math.random();
+    var u = util.random();
     return (1 - u) * params[0] + u * params[1];
   },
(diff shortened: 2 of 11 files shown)


stuhlmueller:
Two thoughts:

1. Having the ability to set the random seed would be useful outside of tests, too. It would be nice to be able to run `RANDOM_SEED=1033080963 webppl myscript.wppl` and get reproducible results.
2. I'm not feeling good about setting `Math.random`, even temporarily.

This makes me wonder whether we should export our own `random` function from `utils`, with the export conditional on `process.env.RANDOM_SEED` and defaulting to `Math.random` if no seed is given. We would then use this function wherever we currently use `Math.random`. If we want to set the seed per test, we'd have to export a `setRandomSeed` function (or something like that) as well.

null-a:
> It would be nice to be able to run RANDOM_SEED=1033080963 webppl myscript.wppl and get reproducible results.

Done.

stuhlmueller:
Looks great, thanks!

PR #226, opened 2015-09-30
Remove unused utility functions.

null-a:
These were used in `ParticleFilter` prior to #169 but are now unused. I don't expect they'll be used from WebPPL programs, though of course I can't be sure.

The functions `getOpt` and `copyObj` are also unused in the WebPPL code base, but I've not removed those as I'm not sure if they are intended to used internally or in WebPPL programs.

Perhaps we should split out these two kinds of functions at some point.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -79,26 +79,6 @@ function copyObj(obj) {
   return newobj;
 }
 
-// More efficient version of (indexOf o map p)
-var indexOfPred = function(l, p, start) {
-  var start = start || 0;
-  for (var i = start; i < l.length; i++) {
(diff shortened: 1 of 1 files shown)


stuhlmueller:
> The functions `getOpt` and `copyObj` are also unused in the WebPPL code base, but I've not removed those as I'm not sure if they are intended to used internally or in WebPPL programs.

Going forward, I think it's okay to aggressively delete code that isn't used in core. This will encourage moving utility functions to packages and/or writing tests that demonstrate the use of such functions if they are supposed to be used in webppl programs.

Issue #201, opened 2015-09-18
Remove all TODO/FIXME comments


null-a:
> There shouldn't be any TODOs in the source code, only in the issue tracker.

Fix, remove, or raise as issues.

PR #229, opened 2015-10-01
Fix edge cases in multinomialSample.

null-a:
A minimal fix to allow the removal of the `FIXME` note. (#201)

### DIFF
--- a/src/erp.js
+++ b/src/erp.js
@@ -504,11 +504,11 @@ function multinomialSample(theta) {
   var probAccum = 0;
   for (var i = 0; i < k; i++) {
     probAccum += theta[i];
-    if (probAccum >= x) {
+    if (x < probAccum) {
       return i;
-    } //FIXME: if x=0 returns i=0, but this isn't right if theta[0]==0...
+    }
   }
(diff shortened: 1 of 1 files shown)


stuhlmueller:
Isn't there now a symmetric edge case when `x == thetaSum` (in which case we return `k - 1`, but this isn't right if `theta[k-1] == 0`)?

In any case, these edge cases happen with probability 0 and are probably not worth worrying about. However, this did remind me that our multinomial sampler could be improved in other ways. I created issue #236.

null-a:
> Isn't there now a symmetric edge case

I don't think so. It's not symmetric because `Math.random` returns a value in `[0, 1)`. I think the only way `x == thetaSum` will hold is through floating point rounding.

> these edge cases happen with probability 0 

Well, *nearly* zero :)

stuhlmueller:
> It's not symmetric because `Math.random` returns a value in `[0, 1)`.

Nice! I didn't know about that.

> Well, *nearly* zero :)

True. Still waiting for my computer that runs on actual real numbers...

PR #235, opened 2015-10-01
Document RANDOM_SEED.

### LINKED ISSUE #227: Add RANDOM_SEED to documentation

null-a:
Doing this made me wonder whether we should have a `--random-seed` option for `webppl`, or whether we should stick to the environment variable for consistency.

Closes #227.

### DIFF
--- a/docs/development/workflow.rst
+++ b/docs/development/workflow.rst
@@ -12,6 +12,17 @@ To only run the tests, do::
 
     npm test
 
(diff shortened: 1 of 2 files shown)


stuhlmueller:
It would be a little better to have `--random-seed` instead, because it's easier to forget that an environment variable is set. Now issue #237. (If we do stick with the environment variable, the `webppl` script should print a message that it's using a random seed.)

Issue #153, opened 2015-07-28
Macros in wppl programs


ngoodman:
It would be useful to have macros in webppl. It should be able to do this without interfering much with the system by running webppl files through sweet.js before everything else happens. (Perhaps only when a macro is detected?)

If we have macros like this, are there other things that can be simplified by relying on them? E.g. AD, desugarring assignment, ..

null-a:
I've been trying to figure out how best to get sweet.js + webppl working in the browser. (As you probably know, a straight-forward application of our current browserify approach doesn't work.) So far I have two possible approaches, neither of which are ideal.

**Load sweet.js with RequireJS**

This is how the library is loaded into the online sweet.js editor. I got this working with webppl by loading sweet.js with RequireJS and modifying webppl so that sweet.js is passed into `run` and `compile`. Downsides with this approach at present include:

1. We have two different mechanisms for getting code into the browser.
2. It ends up loading `escodegen` twice I think.
3. We have to thread the `sweet` module through the compilation code as it comes from different places depending on whether we're in node or the browser.
4. A single `webppl.js` file is replaced by several files.

If you want to try this:

1. Checkout [the branch](https://github.com/null-a/webppl/commit/cfd8d2ab024697b12b34f25db734f3fc55278b3a).
2. `cd compiled`
3. `make clean && make`
4. `cd ..`
5. `python -m SimpleHTTPServer`
6. Visit [http://localhost:8000/compiled/index.html](http://localhost:8000/compiled/index.html)
7. Call `run(<webppl_code>)` in the debug console to test.

**Patch sweet.js so it can be browserified**

It turns out that there are two problems in applying browserify to sweet.js. One comes from the use of `fs` and `require` in ways which can't be transformed by the current tools, and the other comes from a bug in `brfs`. As a proof of concept I worked around this by patching sweet.js and by using an old version of `brfs`. It's not ideal, but it works. Problems include:

1. Having to patch `sweet.js` is brittle.
2. We don't want to be stuck on an old version of `brfs`. (Though I suspect we could fix the bug.)

If you want to try this:

1. Check out [the branch](https://github.com/null-a/webppl/commit/fb1be1c9ec4e1e923d64f6fcd8f62948e0abae73).
2. `npm install` (Should install `brfs@0.0.8`)
3. `cd compiled`
4. `make clean && make`
5. Open `compiled/index.html` in a browser. (`open index.html`)
6. Call `run(<webppl_code>)` in the debug console to test.

[...]

Issue #139, opened 2015-07-02
recursively require headers from webppl package dependencies


hawkrobe:
The package/module system implemented in #133 only handles the header files specified in the package.json file under the ``webppl``` key. 

Ideally, as the (currently non-existent) webppl package ecosystem grows, though, we'd like to be able to specify *other* webppl packages as 'dependencies' in the package.json, and be assured that whatever headers are associated with *those* packages will be available as well. 

This should be possible using tools like [resolve-dep](https://www.npmjs.com/package/resolve-dep), or just creative us of ```require``` statements and package.json introspection.

PR #241, opened 2015-10-06
Add sweet.js macros.

### LINKED ISSUE #209: Package wppl files concatenated in differing orders between browser/node
This is problematic when there are dependencies between packages.

The current implementation for the browser is perhaps the most natural as packages are loaded in the order they are specified on the command line, and individual files within a package are loaded in the order they appear in `package.json`.

null-a:
This is a cleaned up version of the macros branch we've been discussing in #153. I don't have any more changes planned so this is ready to be reviewed.

I'm still using an old version of brfs and a patched version of sweet.js in order to get `browserify` working.

Getting brfs updated is more involved than I'd hoped, it's certainly not a trivial bug-fix. I've opened an issue to try and find out whether a PR making the change we need is likely to be accepted, but I've not heard back yet. Perhaps we're best merging this as is so we can see if we're happy with our approach before putting lots of effort into updating brfs and sweet.js?

Perhaps the fork of sweet.js should live under the probmods account?

This also closes #209.

### DIFF
--- a/compiled/Makefile
+++ b/compiled/Makefile
@@ -13,7 +13,7 @@ $(MINIFIED): $(BROWSERIFIED)
 	uglifyjs $< -b ascii_only=true,beautify=false > $@
 
 $(BROWSERIFIED): $(SRCFILES)
-	browserify -t brfs $(MAINFILE) > $@
+	browserify -g brfs $(MAINFILE) > $@
 
 clean:
 	rm -f $(MINIFIED) $(BROWSERIFIED)
--- a/docs/analysis/documentation.txt
+++ b/docs/analysis/documentation.txt
@@ -1,6 +1,6 @@
 HOW TO USE IT
 
-The top level of the control-flow analysis is the 'analyze' function in the 'analysis/main' module. This function takes a named, CPS'd AST (which 'prepare' of the 'main' module produces given a literal program) and produces a relation over abstract program states along with some information about the relation itself. All this is encoded as a JS object naming this information. These parts are
+The top level of the control-flow analysis is the 'analyze' function in the 'analysis/main' module. This function takes a named, CPS'd AST (which 'prepare' of the 'analysis/main' module produces given a literal program) and produces a relation over abstract program states along with some information about the relation itself. All this is encoded as a JS object naming this information. These parts are
 
 - 'seen': the set of abstract states over which the relation is defined
 - 'calls': a map from an abstract entry point to a set of abstract call points
--- a/docs/development/workflow.rst
+++ b/docs/development/workflow.rst
@@ -39,12 +39,12 @@ many of them automatically using::
 To compile webppl for use in browser, run::
 
     npm install -g browserify
-    browserify -t brfs src/browser.js > compiled/webppl.js
+    browserify -g brfs src/browser.js > compiled/webppl.js
 
 Packages can also be used in the browser. For example, to include the
 ``webppl-viz`` package use::
 
-    browserify -t [./src/bundle.js --require webppl-viz] -t brfs src/browser.js > compiled/webppl.js
+    browserify -t [./src/bundle.js --require webppl-viz] -g brfs src/browser.js > compiled/webppl.js
 
 Multiple ``--require`` arguments can be used to include multiple
 packages.
--- a/docs/packages.rst
+++ b/docs/packages.rst
@@ -2,7 +2,7 @@ WebPPL Packages
 ===============
 
 WebPPL packages are regular Node.js packages optionally extended to
-include WebPPL code and headers.
+include WebPPL code, macros and headers.
 
 To make a package available in your program use the ``--require``
 argument::
@@ -21,7 +21,7 @@ Packages can be loaded from other locations by passing a path::
 
     webppl myFile.wppl --require ../myPackage
 
-Packages can extend WebPPL in three ways:
+Packages can extend WebPPL in several ways:
 
 WebPPL code
 -----------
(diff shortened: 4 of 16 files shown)


null-a:
@mhtess Once this change reaches you, you'll need to reverse the order in which the `wppl` files are specified in your `naturalUtils` package. (Something [like this](https://github.com/null-a/real-kinds-priors/commit/ae6c4e7121e31c4b9765a50fb53f41e0969f16a0).) This is because this PR changes the order in which `wppl` files are loaded, and in your package there are dependencies between the `wppl` files.

stuhlmueller:
> Perhaps the fork of sweet.js should live under the probmods account?

Yes, feel free to move it there.

null-a:
> Yes, feel free to move it there.

Thanks. Done!

stuhlmueller:
Looks pretty good overall. A few comments:

- I found it difficult to understand what the type of `extra`/`extras` is. My thinking went something like this:

    1. `webppl.prepareExtras(packages)` takes an array of packages as arguments, so it probably returns an array of extras, whatever those are.
    2. Let's look at what it actually returns. It's an array of ASTs and an array of macros. So I guess an extra is roughly an object `{ ast, macros }`.
    3. Let's check this idea. `parseExtra` takes an `extra` as an argument. Looking at its code, it assumes that the extra is of the form `{ wppl, macros }`, so that's different from what I thought.
    4. Let's look at other places where `extras` is used. In `compile`, `extras` is (the same type as) the return value of `prepareExtras()`, so maybe it's `{ ast, macros }` after all?

- I wonder if we should have a more unified treatment of header and packages. Once we have dependencies between packages (#139), we could have a `core` package that is an implicit dependence for every other package and that contains the code that's currently in `header.wppl` and `headerMacros.sjs`.

- It's confusing that we have `prepare` and `prepareExtras` which do completely different things. Since `prepare` is only used in the analysis (which itself hasn't been used by anyone in a long time), can we move it there?

- Can we move `cachingRequired` to `transforms/caching`?

- In ` src/browser.js`, it looks like the `underscore` import isn't actually used.

null-a:
> I found it difficult to understand what the type of extra/extras is in different places.

You're right. I was using "extra" to refer to at least 2 things which is awful, thanks for pointing it out! I've refactored this and hopefully it's a little clearer now. Even though I've removed the overloaded use of the term "extra" I'm not overly happy with the new names I've introduced. Please let me know if any better names come to mind.

> I wonder if we should have a more unified treatment of header and packages

I realised this too. This PR takes a *small* step in that direction by treating the header and packages more uniformly, but I agree we should go further eventually.

> It's confusing that we have `prepare` and `prepareExtras`

I've renamed `prepareExtras` which removes the confusion.

I had mixed feelings about moving this. I think there's a case for leaving it where it is as it's usually the case that changing one requires changing (or at least thinking about) the other, so having them nearby seems useful. With that in mind I moved it, but refactored `compile` so that `prepare` could be written in terms of it.

> Can we move cachingRequired to transforms/caching?

Done.

> In src/browser.js, it looks like the underscore import isn't actually used.

Good catch!

stuhlmueller:
> Even though I've removed the overloaded use of the term "extra" I'm not overly happy with the new names I've introduced. Please let me know if any better names come to mind.

Thanks for the adjustments, this is definitely an improvement. The key question regarding naming seems to be: what should we call a pair of code and associated macros? I can't think of a clearly better name than `pair` ("unexpanded code", "bundle", and "spec" are some ideas I had). One could try to avoid the concept entirely by rewriting the pipeline so that it operates more directly on `code` and `macros` (e.g., using `parse` instead of `parseAllPairs`), but I don't know how feasible this is. It's probably not worth worrying about too much, since we'll likely refactor some more when we expand the module system.

> With that in mind I moved it, but refactored `compile` so that `prepare` could be written in terms of it.

Great.

In `browser.js`, would it be possible to move `var extra = webppl.parsePackageCode(packages, verbose);` out of the `run` and `compile` functions and to the top-level so that it is only run once? It seems that this would save some computation.

Would it make sense to get rid of `loadMacros` and to directly call `sweet.loadModule` in the `pkg.js` `load` function?

Typo: convinience

null-a:
> In browser.js, would it be possible to move var extra = webppl.parsePackageCode(packages, verbose); out of the run and compile

Well spotted! The only reason I pass the result of `parsePackageCode` to `run` and `compile` (rather than passing the packages directly) is that I intend to make exactly this move. Unfortunately this doesn't work at present. I think the problem is that `compile` is mutating some of its input which leads to the first call to `run` working but subsequent calls to fail because some parts of the pipeline have been applied twice. I'll look more closely at this at some point, I'm not sure it needs to hold this up though.

> Would it make sense to get rid of loadMacros and to directly call sweet.loadModule in the pkg.js load function?

Good idea, but I don't think so.

`pkg.load` reads the package contents from disk and it only happens in node. The analogous stage in the browser might be the browserify plugin which reads the package contents and stuffs them into `browser.js`. `loadMacros` operates on the contents of a file of macros, and its current position allows it to operate uniformly across node and the browser as the two steps I just mentioned have happened by this point. If we pushed `loadMacros` into `pkg.load` we'd have to add an analogous call into `browser.js` which would make things more complicated I think.

null-a:
> I'll look more closely at this at some point

It turns out that the particular problem I was running into was the fact that the `naming` transform does mutate its input:

````
> ast = parse('(function() { f() })')
{ type: 'Program',
  body: [ { type: 'ExpressionStatement', expression: [Object] } ],
  sourceType: 'script' }
> gen(naming(ast))
'(function (_address0) {\n    f(_address0.concat(\'_0\'));\n});'
> gen(naming(ast))
'(function (_address0) {\n    f(_address0.concat(\'_0\'), _address0.concat(\'_0\'));\n});'
> 
````

`naming` is written using `estraverse.replace`. There doesn't appear to be a straight-forward way to make this non-mutating. I guess the other transforms which use this will be similarly problematic.

I tried the obvious thing of cloning the AST (using lodash) but this took longer than the call to `parsePackageContents` which we're trying to avoid.

stuhlmueller:
> `naming` is written using `estraverse.replace`. There doesn't appear to be a straight-forward way to make this non-mutating. I guess the other transforms which use this will be similarly problematic.

Ok, then let's leave it as it is for now. At some point we'll need to revisit this -- the fact that compile mutates some of its arguments is unexpected behavior. I created issue #247.

Issue #238, opened 2015-10-01
Document how to upgrade dippl.org to latest webppl version


stuhlmueller:
...and make sure it gets done whenever I update the npm package.

null-a:
Should we also update webppl.org when we update the npm package? Keeping the homepage and docs in sync. might [avoid confusion](https://groups.google.com/d/msg/webppl-dev/mY5RtkvrT6Y/5l4EGyOeBAAJ) in the future.

longouyang:
That sounds good, though [updating webppl.org](https://github.com/probmods/webppl/blob/gh-pages/README.md) is just complicated enough that it'd be a bit annoying to automate. Also, sometimes we need to do it manually because there's a concurrent commit in webppl-editor (e.g., back when we changed the trampolining interface).

PR #242, opened 2015-10-06
Add basic browser tests.

null-a:
This makes it straight-forward to run a few very basic browser tests. These are intentionally *very* simple and are only intended to ensure that `broswerify` has worked and that the methods we expose via `webppl` exist and work for simple cases.

It's also serves as a quick way to build WebPPL and load it into a browser so things can be tried out at the console.

Something like this was useful while working on macros. It might also be useful when updating dippl.org. (#238)

To use it do:

````
make clean
make test
````

You can specify which browser the tests should be run in like so:

````
BROWSER="Google Chrome" make test
````

### DIFF
--- a/compiled/Makefile
+++ b/compiled/Makefile
@@ -6,6 +6,7 @@ SRCFILES = $(wildcard ../src/*.js) \
 MAINFILE = ../src/browser.js
 MINIFIED = webppl.min.js
 BROWSERIFIED = webppl.js
+TESTS = ../tests/browser/index.html
 
 all: $(MINIFIED)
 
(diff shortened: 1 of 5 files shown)


stuhlmueller:
Add instructions on how to use this to the docs as well?

It would be great (but not very pressing) to go further in this direction, and to run all tests on travis-ci within the browser as well. I added issue #243 for this.

null-a:
I've briefly mentioned this in the docs. It seemed sensible to switch the docs to describe the use of `make` rather than `browserify` directly as I think it's the easiest way to compile, minimize and run the these tests. I think this is fine as long as everyone who might need to do this has `make` - do they? (I'm thinking about people using windows.)

There's a line in the docs which is now touch by both this branch and the macros branch. Whichever is merged second will need to be merged with dev by hand I think. Maybe merge this first, then I'll take care of the macros branch?

stuhlmueller:
> I think this is fine as long as everyone who might need to do this has `make` - do they?

I think this is safe to assume.

Issue #176, opened 2015-09-08
Rejuvenation for models with structural change


stuhlmueller:
See discussion in issue #20

null-a:
I have [a version of rejuvenation](https://github.com/null-a/webppl/commit/911f03ea10e1399763fa11ab4b0d6cb53c7668d6) which seems to work on our `varFactors` tests. Here's a wordy attempt to explain what it does:

My understanding (loosely speaking) is that SMC is valid for any sequence of distributions which has the distribution of interest as a marginal, and not only for the typical state space model decomposition. The sequence of distributions we use in `ParticleFilter` seems a little odd from a state space perspective (as particles can be "out of sync") but it seems fine with respect to the criteria for valid SMC.

One might have a suspicion that something fishy could be happening when some particles finish before others. In such cases you can imagine padding the shorter paths through the program with additional `factor(0)` statements to ensure all particles encounter the same number of factor statements. I think of our implementation as doing this implicitly. Note that this suggests that particles which have reached the exit should be included in resampling, as indeed they are in `ParticleFilter`.

My understanding of the correctness of rejuvenation is that at each step of SMC, the rejuvenation kernel should have as its stationary distribution the distribution which the particles are approximating. With that in mind, all this change does is tweak the rejuvenation kernel for a given step `n` to have as its stationary distribution the distribution used by `ParticleFilter` at step `n`.

More specifically, assuming programs have been padded as mentioned, then `ParticleFilter` simply runs all particles to the first factor statement (the stack address of which may vary across particles) then resamples, then advances them all to the second factor statement etc., until all particles are done. So the distribution at step `n` is the distribution on traces you get by running particles up to the `n`<sup>th</sup> factor statement. To have this be the stationary distribution of the rejuvenation kernel, all we need to do is switch from truncating the program at the factor statement with a given stack address (as we do now), to truncating it at the `n`<sup>th</sup> factor statement encountered by the particle. That's what this change does.

So, the hope is that this version of rejuvenation will work wherever `ParticleFilter` works, and perhaps `ParticleFilter` works generally. (At present, I'm not very confident that this is the case.)

stuhlmueller:
> In such cases you can imagine padding the shorter paths through the program with additional `factor(0)` statements to ensure all particles encounter the same number of factor statements.

That's a good way to think about it.

[...]

PR #245, opened 2015-10-09
Make rejuvenation work for the varFactors tests.

null-a:
As discussed in #176.

### DIFF
--- a/src/inference/mhkernel.js
+++ b/src/inference/mhkernel.js
@@ -10,6 +10,7 @@ module.exports = function(env) {
   function MHKernel(k, oldTrace, options) {
     var options = util.mergeDefaults(options, {
       proposalBoundary: 0,
+      exitFactor: 0,
       permissive: false
     });
 
@@ -20,8 +21,8 @@ module.exports = function(env) {
     this.k = k;
     this.oldTrace = oldTrace;
     this.reused = {};
-    this.exitAddress = options.exitAddress;
     this.proposalBoundary = options.proposalBoundary;
+    this.exitFactor = options.exitFactor;
 
     this.coroutine = env.coroutine;
     env.coroutine = this;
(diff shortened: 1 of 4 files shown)


stuhlmueller:
Do you think we should make the naming of `proposalBoundary` and `exitFactor` more analogous? They seem to fill similar roles in that they control the initial and final bounds of the MH re-evaluation region. Maybe `proposalBoundary` should be `startERPIndex`, and `exitFactor` `exitFactorIndex`? On that note, `exitFactor` is currently 1-indexed, right? I wonder if it should be 0-indexed, with default `-1` to indicate no limit.

Can rejuvenation lead to completed particles reverting to non-complete status? If `exitFactor` is `k` and we switch from a branch with `< k` factors to a branch with `> k` factors, then this should happen, right? (Assuming that this is the only thing going on in the program.)

null-a:
> Do you think we should make the naming of proposalBoundary and exitFactor more analogous? 

At first I thought this sounded like a good idea, but now I'm not so sure. I'm thinking that:

1. It seems we lose valuable information by dropping the work "proposal". Someone looking at the code might ask "what is it that starts at `startERPIndex`?"
2. Without the "proposal" clue, the analogy might suggest to them that since we stop execution of the program at `exitFactorIndex`, then we must start execution at `startERPIndex`. (Which we don't, of course.)

> exitFactor is currently 1-indexed, right?

That's right. On `Trace` I count how many factor statements an execution has encountered which seems pretty natural to me, and `exitFactor` just follows that. I'm happy to pull in a [change](https://github.com/null-a/webppl/commit/60fe1c924512a325f18d9caa0778349fb502c441) which switches this around if you like. (I've held off while we decide about `startERPIndex` etc.)

> Can rejuvenation lead to completed particles reverting to non-complete status?

Yes, that' the idea.

stuhlmueller:
> Without the "proposal" clue, the analogy might suggest to them that since we stop execution of the program at `exitFactorIndex`, then we must start execution at `startERPIndex`.

It seems to me that this is not too far from the truth -- `MHKernel` only ever re-executes code that comes after `startERPIndex`/`proposalBoundary`. If you want to be entirely explicit, you could call it `minProposalERPIndex`.

> I'm happy to pull in a [change](https://github.com/null-a/webppl/commit/60fe1c924512a325f18d9caa0778349fb502c441) which switches this around if you like.

If you end up renaming `proposalBoundary` to `...Index`, switch it; if not, the previous version is better.

stuhlmueller:
Should I merge this as is? (I'm happy to.)

null-a:
> Should I merge this as is? (I'm happy to.)

Sure.

Issue #259, opened 2015-10-23
Merge variational inference


dritchie:
Over in my fork of webppl, I have an inference routine that implements various variational inference algorithms (ELBO maximization, EUBO minimization using posterior samples, and variational particle filtering a.k.a. NASMC):

https://github.com/dritchie/webppl/blob/vpf/src/inference/variational.js

According to recent conversation with Noah, one of the next high-priority items for webppl is to get HMC implemented and merged (#81), so I figured it'd be a good time to bring this up, since both HMC and variational inference require gradients. Also, this code could be pretty useful for implementing 'Bayesian adaptive caching.'

AFAIK, there are just a couple of roadblocks to overcome/decisions to be made before this is possible.
 - The ad library I'm using is structured a little bit differently than Sid's, and it also introduces support for tensor-valued variables. We've talked about how to resolve these differences, and it seems totally doable.
 - My code sweet-js transforms files via `make`, whereas Sid's uses an augmentation to `require`. I'm totally happy to convert to established practice here.

Any other issues anybody can see?

null-a:
> My code sweet-js transforms files via make, whereas Sid's uses an augmentation to require. 

Getting this working in the browser seems more straight-forward using a build step or similar (perhaps using grunt #248) rather than `require-transform`. I have a [branch](https://github.com/probmods/webppl/compare/dev...null-a:gradients?expand=1) where I'm experimenting with merging HMC and while it's far from complete, it is running in the browser. It differs slightly from your approach:

* Rather than splitting out the "scorers" into a separate file I experimented with a [transform](https://github.com/null-a/webppl/blob/a4bf51111ccfab3b11ab2999c9938c676c750978/src/transforms/scoreFunctions.js) which selectively expands AD macros in only the relevant functions.
* The transformed version of the ERP code isn't checked into Git. Instead the transform is performed by an NPM post install script. With this approach we'd have to manually run the script if `erp.ad.js` is changed during development, or perhaps automate it with [`grunt watch`](https://github.com/gruntjs/grunt-contrib-watch).

I don't claim that either of these are better than your solutions, I just wanted to point them out.

> We've talked about how to resolve these differences

How can we make this happen? Two suggestions I have are:

1. I could attempt to integrate AD into WebPPL in a way which is usable by both variational and HMC. (Chatting to @dritchie and @iffsid to figure out the details e.g. tensor support.) We might not get this exactly right, but hopefully once it's done work could proceed largely independently on merging the variational and HMC branches.
[...]

Issue #81, opened 2015-04-08
Implement HMC


ngoodman:
Following the ideas in [Nonstandard Interpretations of Probabilistic Programs for Efficient Inference](http://www.stanford.edu/~ngoodman/papers/WGSS-NIPS11.pdf) we can implement Hamiltonian Monte Carlo (HMC) using automatic differentiation (AD). 

HMC is useful in models with many continuous parameters. This seems especially helpful for bayesian data analysis, where we are often dealing with parameters per subject and condition (i.e. thousands of continuous parameters).

I believe @iffsid already has AD working using sweet.js, so it would perhaps be straightforward to do HMC?

The Stan project has some good tricks for making HMC work in practice. Worth looking at if we start using HMC a lot.

With HMC implemented, evaluate combinations such as HMC rejuvenation in SMC.

iffsid:
The `ad` branch (current b24e282ba05ef60a0e2a60037b9b412ebe3cb363 ) has both HMC and HMC rejuvenation in SMC (HSMC) implemented.
HMC can be run with a repetition-set of kernels (e.g. `['mh', 'leapfrog']`, etc.) or just as `['mh']` or just as `['leapfrog']`. Same applies to the rejuvenation in SMC.

A simple example for HSMC

``` javascript
var model = function() {
  var mu = gaussian(0, 2);
  factor(gaussianERP.score([mu, 1], 5.5));
  factor(gaussianERP.score([mu, 1], 6.5));
  return mu;
};

var hsmcOpts = {numParticles: 1000,
                kernelOpts: {stepSize: 0.1,
                             steps: 5,
                             iterations: 20,
                             proposers: ['leapfrog'],
                             aggregator: 'score'},
                aggregator: 'score'};

var erp = HSMC(model, hsmcOpts);
console.log("logZ", erp.normalizationConstant)
ad.untapify(expectation(erp))
// => 5.3 (approximately)
```

The codebase is fairly clean, with a decent amount of refactoring of `trace` and `particle` data-structures that helps keep working with stateful `ad` sane.
Also, some work along the lines of #86 was put in to simplify the interface and usability.
The `aggregator` option is used to decide if the histogram update should happen on `'count'` or `'score'` -- the latter for use when the support of the returned distribution is `Real`

I'm not closing the issue yet because the test-framework needs work to `untapify` the results where necessary so that the results match the expected results. Currently most things are a `tape`, unless explicitly asked to be `untapified`; a requirement when nesting models with `ad`.

[...]

Issue #174, opened 2015-09-02
Use linked list as primary data structure


stuhlmueller:
~~This would make copying trivially O(1).~~

~~See [ImmutableJS](https://facebook.github.io/immutable-js/).~~

~~This came up in #169.~~

null-a:
I suspect header functions `map`/`reduce` and friends scale quadratically(?) in the length of the input array, presumably because of the `ar.slice(1)` at each step of the recursion. (This [example](https://github.com/null-a/webppl/blob/610914a2f3d7ceb952589895c3e323cb243f2ddc/examples/headers.wppl) shows something non-linear I think.)

Perhaps we need a persistent list data structure available in WebPPL?

(I think this might be a practical concern in BDA type models. For example it causes the data munging of a few thousand data points to take ~20 secs in the model [discussed here](https://groups.google.com/d/msg/webppl-dev/lEzC_1dAJY0/JyIbxFbvDQAJ).)

stuhlmueller:
Yes, the scaling of `map` etc is quite bad.  On my computer, mapping `function(x){return x + 1;}` over a list of 6655 entries (all `1`s) takes 389ms, with 13310 entries it takes 1258ms, and with 26620 entries, the process runs out of memory:

    FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
    Abort trap: 6

I'm not sure yet how we should resolve this problem globally (e.g., whether to replace the header functions with alternatives), but maybe we can start by creating a `webppl-immutable` package that provides a cons-like immutable list data structure and corresponding alternatives to `map` etc. This could solve the issue for concrete applications, and we'd be more informed when we think about what to do about core webppl.

null-a:
> maybe we can start by creating a webppl-immutable package that provides a cons-like immutable list data structure and corresponding alternatives to map etc

I've started a [linked list implementation](https://github.com/null-a/webppl/commit/7a6227960e57907cfd164c84eff9fd6178e7e6cc).

On your test it gives the following:

 | Array (ms) | List (ms)
----|----|----
6655 | 258 | 23
13310 | 852 | 36
26620 | out of mem | 88

I've also tried it out on @mhtess's [model](https://groups.google.com/forum/#!msg/webppl-dev/lEzC_1dAJY0/JyIbxFbvDQAJ) and the results are promising:

 | Array | List
----|----|----
Memory use at 1st iteration (MB) | 1240 | 150
Run time (min:secs) | 1:54 | 0:26

[...]

Issue #136, opened 2015-07-02
General architecture for conditionally enabling/disabling source transforms


dritchie:
Some inference algorithms require source transforms that others do not; having all of them enabled all the time can add unnecessary overhead (see #135). It would be good to have a general-purpose solution for detecting when a program-to-be-compiled uses an inference algorithm, and then enabling just the transforms that that algorithm needs. This is non-trivial, especially because the transforms aren't necessarily independent / re-orderable.

PR #265, opened 2015-11-09
HMC

null-a:
This is a cleaned up version of my [gradients](https://github.com/probmods/webppl/compare/dev...null-a:gradients) branch. There are a few loose ends to tie up (see below) but I don't expect them to change things much, so now might be a good time to start reviewing my changes.

This PR includes:

* ad.js integration
* A revised `MCMC` interface which allows per-kernel options to be specified
* A HMC kernel which can be used with `MCMC` and `SMC`

Using HMC with the new interface looks like this:

````js
MCMC(model, { samples: 100, kernel: 'HMC' })
SMC(model, { particles: 100, rejuvSteps: 5, rejuvKernel: 'HMC' })
MCMC(model, { samples: 100, kernel: { sequence: { kernels: ['HMC', { MH: discreteOnly: true }] } } })
````

Before use, the ERP score code needs to be transformed, this is now done like so:

````bash
./scripts/transformERP
````

This should happen automatically after an `npm install`. If `erp.js` is missing when `webppl` is run a descriptive error message is shown.

**Still to do**

* ~~Switch to the latest ad.js for tensor support. (See #259.) Test Dirichlet. (See #275.) Switch to version of `isTape` exported by `ad.js` (in `aggregation.js`) once available.~~ Postponed, will tackle as a separate piece of work.
* [x] Avoid unnecessarily creating tapes.
* [x] Remove `TODO` notes. I've intentionally left these in for now as they might be useful while the PR is been reviewed.
* [x] Add more continuous and mixed continuous/discrete tests, we only have 1 or 2 right now. Suggestions for good tests welcome!
* [x] Add a `HMC` function or some other convenient way of saying `kernel: { sequence: { kernels: ['HMC', { MH: discreteOnly: true }] } }`. Thoughts?
* [x] Update inference docs.
* [x] Update `MAP` and `entropy` methods on `ERP`. (Not needed if we drop constraint stuff.)
* ~~Maybe: Add `erp.js` to repo.~~
* Maybe: Reject proposal rather than throw error on structure change, if we think doing so is correct. Discussed in #81.

**Speed**

I ran a few quick benchmarks and in the model I tested the time spent doing inference increased by ~20% using `MH` and by ~40% using `SMC` after these changes. I've not spent any time trying to optimize that away. The inference tests are probably quite a bit more than 40% slower because of the increase in the time taken to compile programs.

**Future work**

As discussed in #81, there are several things not included here that we might like in the future:

* Constraint handling by transformation. (Do we need to handle hard `factor` statements?)
* Step size adaption.
* Support for nesting models.


stuhlmueller:
This looks good overall -- and like it was a lot of work! Thanks!

Are there any examples where the current implementation of HMC does visibly better than plain MH? Some model with multiple continuous variables that are strongly correlated under the posterior so that all need to be updated jointly? It would be good to include at least one such example in the tests.

In general, the main thing we should convince ourselves of is that the algorithm is correctly implemented. I'd add tests that include various sorts of dependencies between discrete and continuous erps and factors. For example:

- Direct dependencies:

        var x = sample(continuousERP)
        var y = sample(continuousERP, [x])
      
        var x = sample(discreteERP)
        var y = sample(continuousERP, [x])
      
        var x = sample(continuousERP)
        var y = sample(discreteERP, [x])

- Indirect dependencies: Each of the three cases above, but with `x` transformed by a compound function that calls out to a primitive function:

        var f = function(x) {
          return Math.log(x + x);
        }

- Single factor depending on multiple ERPs: Each of the six cases above, but including a factor with a score that depends on a function of `x` and `y` (probably need to write a different function for each combination of discrete/continuous).

- Multiple factors depending on a single ERP

- More complicated programs:

        var x = sample(continuousERP)
        var y = sample(continuousERP, [f(x)])
        var w = sample(discreteERP, [g(x, y)])
        factor(h(x, y, w))
        var z = sample(continuousERP, [i(x, y, w)])
        factor(j(y, z, w))

Other comments:

- The installation instructions in `docs/development/install.rst` now include a mention of `./scripts/transformERP`, but you say it happens automatically with `npm install` -- are the extra install instructions unnecessary?

- I'd rename `scoreFunctions.js` to something that makes it clear that it's about automatic differentiation (maybe `adscorers.js`). Also, to make it more like the other transforms, don't export the transform function directly, but rather as property of an object?

- Having a list of `helpers` in `scoreFunctions.js` seems a bit fragile. Could we have a way to mark the corresponding functions directly in `erp.ad.js`?

- As you point out, `untapify` in `src/aggregation.js` should probably be integrated into `untapify` in `ad.js`. If not, the condition should be factored out as `isNonTapeObject` or something like that to make it clearer what's going on.

- Is there a reference for the constraint handling mechanism you implemented, or did you make it up? It looks sensible, but it would be good to have some assurance that it's doing the right thing. At least we'd want some tests that exercise it.

- Add a comment to `erp.ad.js` explaining why `sum` needs to be defined there even though we already have one in `util.js`?

- `hmckernel.js`: Add a reference to the top of the file that points to the main source for the algorithm? (The Radford Neal handbook chapter?)

- `hmckernel.js`: Add a comment that explains what `ad.yGradientR` in line 153 does?

- What's the role that `thisPositionStepCont` plays? Add a comment to clarify?

- What's the motivation for the interface change to `Initialize` and kernels? In particular, I'm not convinced that kernels should take `runWppl` as an argument. Maybe it should be part of `options`, or a method on the `trace` object?

- Related to this interface change (but not entirely explained by it): I found `src/inference/kernels.js` difficult to understand, in particular the top half. I think it would have helped a little to rename `MH`, `HMC`, and `sequence` to `MHKernel`, `HMCKernel`, and `SequenceKernel` respectively.

- In various places (`hmckernel.js`, `initialize.js`, `smc.js`, `mhkernel.js`, `trace.js`), `ad` is accessed without being explicitly imported. I assume this works because, by the time these modules are imported, `ad` has already been made available globally? If so, I would feel better if it were imported explicitly. If you want reference to a single instance, do this via `env`?

- `mhkernel.js`:65 - I'd avoid initializing multiple variables in one line, or at least limit it to the case where all are uninitialized. I was confused for a moment, thinking we were destructuring the return value of `this.oldTrace.findChoice`.

- In 'smc.js', I wonder if it would be useful to keep the particle `logWeight` (~ estimate of normalization constant) tapified. Isn't the gradient of this quantity one of the ingredients we need for variational inference? (If not, how is it different?)

- Now that the MCMC/SMC interface is getting more complex, we should start documenting it.

- Regarding a convenient way to say `kernel: { sequence: { kernels: ['HMC', { MH: discreteOnly: true }] } }`: I think your proposed `HMC` function would be reasonable. We want to make it easier to use HMC+MH than pure HMC, so that users don't accidentally use the latter and then get confused about why the results for mixed continuous-discrete models aren't right.

- Typos: hmckernel.js: structual->structural; mhkernel.js: indicies -> indices

null-a:
> are the extra install instructions unnecessary?

You're right. I'd still like to include something about this, would you be in favor of including a section on this under workflow? Something like [this](https://github.com/null-a/webppl/commit/cec2d95a4f9605690e3371026ffef68f4315d28e).

> Having a list of helpers in scoreFunctions.js seems a bit fragile.

Good point. I've changed this to transform functions with names ending in `AD`. There are more elegant ways, but this is nice and simple. It also makes it easier see at a glance whether a score function only calls ADified helpers.

> As you point out, untapify in src/aggregation.js should probably be integrated into untapify in ad.js.

I've started talking to Sid about this.

> Is there a reference for the constraint handling mechanism you implemented

It's from Radford Neal's handbook chapter. Section 5.1, p36.

I don't think the current implementation is very general though. It only checks the constraint which comes from the support of the variable which is been updated which I think can come apart when constraints depend on other variables.

It's conceivable that simply rejecting proposals that wind up in areas of zero probability is correct, but I don't know that it is.

In #81 the suggestion was to transform constrained variables into unbounded spaces which sounds reasonable. Assuming we tackle this as a separate piece of work, the immediate decision is whether to stick with what we have, take it out, or improve it. Of those, the first two probably make most sense.

> Add a comment to erp.ad.js explaining why sum needs to be defined there even though we already have one in util.js?

This is now `sumAD`. Is that enough of a hint, or do you still want a comment?

> hmckernel.js: Add a comment that explains what ad.yGradientR in line 153 does?

What do you have in mind? I happy to do it, I'm just not sure what jumps out to you about this compared to any other part of the algorithm. Does `// Compute gradient of score w.r.t. continuous variables` cover it?

> What's the role that thisPositionStepCont plays? Add a comment to clarify?

This is indeed a little awkward. I've added a [comment](https://github.com/null-a/webppl/commit/26e654f55aa84febda085fdc68c769bbfc5f36c3) to explain.

> What's the motivation for the interface change to Initialize and kernels? In particular, I'm not convinced that kernels should take runWppl as an argument. Maybe it should be part of options, or a method on the trace object?

The idea is that it's a convenient way to give `HMC` access to `wpplFn` and the appropriate store and address needed to run the program from the beginning. `Initialize` already took these as three separate parameters, it now just takes a single `runWppl` argument instead. This has the nice side-effect of allowing us to clone the store (before running the program) in just one place.

I did considered including it on `Trace` but decided against it as:

1. You have to include it in the logic for copying/slicing traces
2. You have to pass it along to each `new Trace()` in `HMC`
3. It doesn't work for `Initialize` as it doesn't take a `Trace`

It could go in `options`, but the problem is that it doesn't play so well with the current approach of gradually building up the kernel by partially applying arguments to the main kernel function. It's not impossible of course (we can do it by wrapping the original function with a function that merges option objects appropriately) but I'm not sure it's worth it in this case. Another way you could go is to question the current approach of partially applying arguments -- I'm happy to talk about that if you like.

> I think it would have helped a little to rename MH, HMC, and sequence to MHKernel, HMCKernel, and SequenceKernel respectively.

Done. Also added a comment to `parseOptions` which might help.

> In various places (hmckernel.js, initialize.js, smc.js, mhkernel.js, trace.js), ad is accessed without being explicitly imported.

Indeed, this works for the reason you mention. The reason I didn't do the obvious thing and `require('ad.js')` in each file is that doing so requires specifying which mode `ad.js` should use each time `require` is called. I was happier calling methods on a global `ad` variable (which has to be there for `ad.js` anyway) than repeatedly having to say "reverse mode please" everywhere.

I'm sure this could be cleaner, but does accessing it via `env.ad` buy us much since we'll still have a top-level `ad` hanging around?

> In smc.js, I wonder if it would be useful to keep the particle logWeight (~ estimate of normalization constant) tapified. Isn't the gradient of this quantity one of the ingredients we need for variational inference?

Not sure. Won't we need to change lots of things for VI? Can we worry about it when we need it?

> Now that the MCMC/SMC interface is getting more complex, we should start documenting it.

Sure. I've made a start on this [here](https://github.com/null-a/webppl/commits/docs), see what you think. I'll consider this a separate piece of work. i.e. #131.

> I think your proposed HMC function would be reasonable.

I've thought about this and if people already use `MCMC` routinely then I don't think adding an `HMC` method is enough. Perhaps we should have the common case of HMC + MH be available using `MCMC(model, { kernel: 'HMC' })` and come up with another name for the "HMC only" kernel. Alternatively we could have something like `MCMC(model, { kernel: 'HMC+MH' })` be short hand for the thing you usually want.

Everything else you mention under "Other comments" has been done.

null-a:
I think I've now added most of the tests you suggested. I hope the following makes sense:

Var Types | Direct Dependence | Indirect Dependence | Direct + Factor | Indirect + Factor
-------------|---------|-----------|--------------------|----------------------
CC | [bivariateGaussian](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/bivariateGaussian.wppl) | [indirectDependency](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/indirectDependency.wppl) | [bivariateGaussianFactor](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/bivariateGaussianFactor.wppl)* | [gaussianMeanVar](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/gaussianMeanVar.wppl)
DC | [mixed2](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/mixed2.wppl) | [mixed4](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/mixed4.wppl) | [mixed2Factor](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/mixed2Factor.wppl) | [mixed4Factor](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/mixed4Factor.wppl)
| CD | [mixed1](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/mixed1.wppl) | [mixed3](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/mixed3.wppl) | [mixed1Factor](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/mixed1Factor.wppl) | [mixed3Factor](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/mixed3Factor.wppl)

e.g. DC = `var x = discrete(); var y = continuous(x);`

*Note -- in bivariateGaussianFactor the `factor` depends only on `x`, not `x` and `y`.

The "Multiple factors depending on a single ERP" case might be covered by [gaussianMean](https://github.com/null-a/webppl/blob/hmc/tests/test-data/stochastic/models/gaussianMean.wppl)?

As far as tests go, I think that just leaves your "More complicated programs" suggestion.

stuhlmueller:
> would you be in favor of including a section on this under workflow? Something like [this](https://github.com/null-a/webppl/commit/cec2d95a4f9605690e3371026ffef68f4315d28e).

Looks good.

> Assuming we tackle this as a separate piece of work, the immediate decision is whether to stick with what we have, take it out, or improve it. Of those, the first two probably make most sense.

Yes, I'd stick with what we have for now.

> This is now `sumAD`. Is that enough of a hint, or do you still want a comment?

That's fine.

> > hmckernel.js: Add a comment that explains what ad.yGradientR in line 153 does?

> What do you have in mind? I happy to do it, I'm just not sure what jumps out to you about this compared to any other part of the algorithm. Does `// Compute gradient of score w.r.t. continuous variables` cover it?

The reason it jumped out at me is because it operates by mutation, and it's not immediately clear to the reader what exactly is being mutated and how. Your comment is sufficient.

> > What's the motivation for the interface change to Initialize and kernels?
> I did considered including it on `Trace` 

Are there any cases where what's currently stored in the trace (the random choices for each address) and the `wpplFn` come apart -- where we wouldn't want to store them together?

My impression is that it makes a lot of sense to combine them, since the random choices are only meaningful in the context of a particular program

stuhlmueller:
> would you be in favor of including a section on this under workflow? Something like [this](https://github.com/null-a/webppl/commit/cec2d95a4f9605690e3371026ffef68f4315d28e).

Looks good.

> In #81 the suggestion was to transform constrained variables into unbounded spaces which sounds reasonable. Assuming we tackle this as a separate piece of work, the immediate decision is whether to stick with what we have, take it out, or improve it. Of those, the first two probably make most sense.

Yes, I'd stick with what we have for now.

> This is now `sumAD`. Is that enough of a hint, or do you still want a comment?

That's fine.

> > hmckernel.js: Add a comment that explains what ad.yGradientR in line 153 does?

> What do you have in mind? I happy to do it, I'm just not sure what jumps out to you about this compared to any other part of the algorithm. Does `// Compute gradient of score w.r.t. continuous variables` cover it?

The reason it jumped out at me is because it operates by mutation, and it's not immediately clear what is being mutated and how. Your comment should be sufficient.

> > What's the motivation for the interface change to Initialize and kernels?
> I did considered including it on `Trace` 

Are there any cases where the random choice values stored in the trace and the `wpplFn` come apart -- that is, where we wouldn't want to store them together? My impression is that it makes a lot of sense to combine them, since the random choices are only meaningful in the context of a particular program. 

> but decided against it as:
> 
> 1. You have to include it in the logic for copying/slicing traces

This seems easy -- just copy over the reference?

> 2. You have to pass it along to each `new Trace()` in `HMC`

True. I only see one such call at the moment (in `positionStep`) -- are there more?

> 3. It doesn't work for `Initialize` as it doesn't take a `Trace`

Maybe it should? Initialize could take a trace without choices, and turn it into one with the choices filled in?

The main consideration for me is that one should be able to think about of as maps from traces to traces. The current interface makes this possible -- `k` is there because we're in CPS, and it's natural to have `options` to parameterize the kernel -- but including `runWppl` makes this more difficult, in particular since the role and functionality of this argument is not immediately clear.

> I'm sure this could be cleaner, but does accessing it via `env.ad` buy us much since we'll still have a top-level `ad` hanging around?

I think it's worth establishing the principle that we don't use any externally defined global variables in the js part of the webppl codebase. (I think this is the case right now?) Then, in the future, we can use a smarter linter that will complain about undefined variables.

> Won't we need to change lots of things for VI? Can we worry about it when we need it?

Yes, you're right.

> > Now that the MCMC/SMC interface is getting more complex, we should start documenting it.
> 
> Sure. I've made a start on this [here](https://github.com/null-a/webppl/commits/docs), see what you think. I'll consider this a separate piece of work. i.e. #131.

Nice. This will be very useful.

> Perhaps we should have the common case of HMC + MH be available using `MCMC(model, { kernel: 'HMC' })` and come up with another name for the "HMC only" kernel.

Fine with me.

null-a:
Thanks for the feedback! I've pushed a few commits which address the easy bits.

**Kernel Interface/`runWppl` Argument**

> The main consideration for me is that one should be able to think about of as maps from traces to traces. 

That's how I think of them, and this change hasn't prevented me from doing so! And once `runWppl` has been [partially applied](https://github.com/null-a/webppl/blob/6ede011a7ec6d0e316e7619cd32c0729a922f54e/src/inference/mcmc.js#L21), what we have is exactly a (CPS) map from trace to trace.

It seems to me that the crux of the matter is how should `wpplFn` (and the `store`, `address`) be passed to the kernels. I mentioned why I thought the approach I picked was better than threading these things through inference as part of `Trace`, and it still seems fine to me at the moment. (That may well change of course!) To check my thinking and to maybe better illustrate what I was describing I've [coded up](https://github.com/null-a/webppl/commit/a880b5ed0b7275f534ea32499fce7aa19f97d432) the "threading through via `Trace`" approach.

(I've not done the passing empty `Trace`s to `Initialize` bit. Instead I've put `Initialize` back how it was before HMC.)

Have a look and let me know which you prefer - we'll go with that.

**Global `ad`**

> I think it's worth establishing the principle that we don't use any externally defined global variables in the js part of the webppl codebase. (I think this is the case right now?) Then, in the future, we can use a smarter linter that will complain about undefined variables.

I'm not sure that gets to the heart of it. Presumably a smarter linter would have a list of variables to ignore (e.g. `env`) so we could just add `ad` to that list.

To be clear, I don't much like having a global `ad` myself, but switching to using `env.ad` instead doesn't seem fundamentally different to me.

Perhaps the core issue here is that webppl programs share their global scope with the JS part of the code. If we could pass `eval` a JS object to use as the global environment we'd be able to make `ad` available globally in webppl programs without doing so in the JS part. I seem to remember that I once read the `eval` docs (for another reason) and thought this was possible, but then couldn't make it work. I could take another look though if this *is* the issue.

(Note: Things might change if we end up AD transforming inference code, but I've not thought that through.)

> externally defined global variables in the js part of the webppl codebase

To clarify, the `ad` global is coming from [here](https://github.com/null-a/webppl/blob/6ede011a7ec6d0e316e7619cd32c0729a922f54e/src/header.js#L125)

stuhlmueller:
**Global `ad`**

> Presumably a smarter linter would have a list of variables to ignore (e.g. `env`) so we could just add `ad` to that list.

Can't this list be empty? I know that this isn't the case for `.wppl` code at the moment, but can't we get by without any locally undefined variables in our `.js` files?

> To be clear, I don't much like having a global `ad` myself, but switching to using `env.ad` instead doesn't seem fundamentally different to me.

It seems fundamentally different to me in that `env` isn't an undefined variable. Instead, it's explicitly passed into modules (e.g. [here](https://github.com/null-a/webppl/blob/hmc/src/inference/smc.js#L12)), so it wouldn't need to be part of the list of variables the linter would need to ignore.

> Perhaps the core issue here is that webppl programs share their global scope with the JS part of the code.

That's an issue, too, but fixing this seems a lot more ambitious than ensuring that any of our `.js` files can be imported as a node module without implicit dependencies that need to be loaded first to avoid "undefined variable" errors.

How about the following alternative to `env.ad`: you create a helper file (maybe `src/ad.js`) that imports `var ad = require('ad.js')({ mode: 'r' });` and re-exports `ad`; then, wherever you want to use `ad`, you require this file. This way, we have explicit imports without re-specification of the mode everywhere.

null-a:
>> switching to using env.ad instead doesn't seem fundamentally different to me

> It seems fundamentally different to me in that env isn't an undefined variable. Instead, it's explicitly passed into modules

Ah, of course! You're right. That was an oversight on my part, thanks for clearing it up!

>  create a helper file (maybe src/ad.js) that imports var ad = require('ad.js')({ mode: 'r' }); and re-exports ad

I've considered this too, and I think it's the way to go. It's more transparent, and it also works in places where we don't currently have access to `env`. (e.g. `erp.js`.)

stuhlmueller:
**Kernel Interface/`runWppl` Argument**

> It seems to me that the crux of the matter is how should `wpplFn` (and the `store`, `address`) be passed to the kernels.

Agreed. I have strong intuitions that a kernel (in our usage) should only have one main argument, the trace. To me, this is a strong enough argument to outweigh the considerations that have been given in favor of making `runWppl` an additional arugment to kernels -- to recap, those were (and let me know if I'm missing one):

1. "it's a convenient way to give `HMC` access to `wpplFn` and the appropriate store and address needed to run the program from the beginning."
2. "`Initialize` already took these as three separate parameters, it now just takes a single `runWppl` argument instead."
3. "This has the nice side-effect of allowing us to clone the store (before running the program) in just one place."

I agree that it's convenient (1). I suspect that 2 and 3 could be handled with similar elegance using the "threading through via `Trace`" approach (though I'm not certain). In particular, maybe store cloning could be handled automatically as a consequence of trace copying?

Thanks for coding up the "threading through via `Trace`" version! Is there a reason to store `wpplFn`, `s` and `a` separately, or could we just as well store the combined `runWppl` function? The latter would feel very natural to me. The trace already stores the continuation at each of the random choices. I think it's reasonable to also store the initial program continuation.

null-a:
> The trace already stores the continuation at each of the random choices. I think it's reasonable to also store the initial program continuation.

Yeah, definitely. That's always been in my mind, and having thought about it more I suspect consistency here is more important than (what I see as) the convenience of doing it the other way, so I agree we should take the "threading via `Trace`" approach.

A concrete example of this consistency is that we now have [a method](https://github.com/null-a/webppl/blob/cf7ba99f496d7e0fb1d0266d7ffa837d88621d04/src/trace.js#L43) on `Trace` which continues execution from the right place, whether that's from the beginning of the program or from a factor statement at which `saveContinuation` was called.

>  maybe store cloning could be handled automatically as a consequence of trace copying?

This was already happening, but the method I mention above does move another spot where cloning was happening into `Trace`.

> Is there a reason to store wpplFn, s and a separately

The idea is that rather than expecting callers of `new Trace()` to create the special `runWppl` thunk, that logic can be pushed into `Trace`. The end result is the method I linked to above.

Thanks for all your help so far!

stuhlmueller:
> A concrete example of this consistency is that we now have [a method](https://github.com/null-a/webppl/blob/cf7ba99f496d7e0fb1d0266d7ffa837d88621d04/src/trace.js#L43) on `Trace` which continues execution from the right place, whether that's from the beginning of the program or from a factor statement at which `saveContinuation` was called.

Nice!

stuhlmueller:
> Are there any examples where the current implementation of HMC does visibly better than plain MH? Some model with multiple continuous variables that are strongly correlated under the posterior so that all need to be updated jointly?

Has this already been dealt with as part of one of the tests? Would be good to include such an example.

What is left to do before we can merge this? The points in the to do list of the initial comment?

null-a:
>> Are there any examples where the current implementation of HMC does visibly better than plain MH? 
> Has this already been dealt with as part of one of the tests?

It's possible that this is the case for one of the tests but I've not spent any time carefully comparing `MH` to `HMC` in terms of performance.

> Would be good to include such an example.

It would, although I don't have such an example at present. The Bayesian linear regression challenge problem might work for this, but at the moment I'm not convinced that `HMC` is that big of a win here. `MH` is surprisingly fast, and there are a few things which might not be helping `HMC`:

* Tuning step size and simulation length is *really* fiddly. In particular, I think the posterior on the precision really narrow, and it seems hard to find a step size which works with this but still explores well. I can make `HMC` do quite a bit better by transforming the problem by re-writing the program by hand, but that's not something you need to do with this problem in Stan. Maybe NUTS helps, or maybe it's something else that I'm missing.
* The prior on precision is constrained to be positive, and I'm not sure how well our current approach fairs compared to transforming to an unconstrained space. (I think it's probably OK in this case, though I have some uncertainty here.)

We'd maybe have a better chance of finding good examples if more people can try `HMC` out. Could we merge this without such an example? If we're nervous about correctness, could we do something equivalent to calling it a beta version?

Somewhat related: I have a mostly working implementation of the adaptive step size algorithm mentioned in #81. It's already been suggested that we don't need this right away and it may well be easier to review separately, but I *could* include it here if you prefer.

> What is left to do before we can merge this?

As well as the things in the list:

* We should chat about performance. Is the general slow down from AD acceptable, or do we want to do some work to try to improve that?
* I'm no longer happy with my approach of transforming `erp.ad.js` via a script when `npm install` is run. The script seems to work, but there's a chance someone will have an odd set-up and it will break. We can remove that risk altogether by committing `erp.js` to the repo, and I think the only costs are 1) a little duplication 2) `erp.js` and `erp.ad.js` may get out of sync. (We can easily fix the latter.) What do you think?

> The points in the to do list of the initial comment?

I'll make sure the list it up-to-date.

null-a:
> I have a mostly working implementation of the adaptive step size algorithm

To clarify, the main reason this isn't part of this PR is that I'm not very happy with the implementation, and I didn't want fixing it to hold up this PR.

stuhlmueller:
We shouldn't spend a lot of time looking for an example where HMC does better, but I think having one would be sufficiently good that we should look a little. Such an example would provide additional assurance that the implementation is correct. More importantly, seeing an example where HMC does better than MCMC (given the same runtime) would tell us that we're on the right track performance-wise — that is, that our implementation of HMC doesn't incur costs somewhere that undo any algorithmic gains (if only in a cherry-picked ideal case model).

How about the following kind of model, where we sample a number of uniform variables (20, say) and constrain their sum relatively tightly to be equal to some constant?

~~~~{javascript}
var bin = function(x){
  return Math.floor(x*10)/10;
}

var model = function(){  
  var xs = repeat(20, function(){
    return uniform(0, 1);
  });  
  var targetSum = xs.length / 2;
  var error = sum(xs) - targetSum;
  factor(-100*error*error);  
  return bin(xs[0])
}

print(MH(model, 10000))
~~~~

For this model, single-site MH should have problems, since (less importantly) it will need to do 20x as many proposals to make a move that's comparable to what HMC can achieve in one proposal and (more importantly) all variables are strongly correlated under the posterior, so that the intermediate moves that MH would make are unlikely to be accepted.

> I have a mostly working implementation of the adaptive step size algorithm mentioned in #81. It's already been suggested that we don't need this right away and it may well be easier to review separately, but I *could* include it here if you prefer.

I agree that it's better to do it separately.

> We should chat about performance. Is the general slow down from AD acceptable, or do we want to do some work to try to improve that?

What's the cause of the slowdown for `MH` and `SMC`?

> I'm no longer happy with my approach of transforming erp.ad.js via a script when npm install is run. The script seems to work, but there's a chance someone will have an odd set-up and it will break.

Are you thinking of concrete ways in which it could break, or is this more of a general "unknown unknowns" concern? Is there some way to make it more robust and/or make it so that if it's broken the user will be notified with instructions on how to compile it themselves? I think it's substantially preferable not to check in compiled code.

null-a:
> How about the following kind of model

`MH` doesn't do too bad with that model, but if you tighten the constraint further (`-100 => -5000` in this case) you see results like:

````js
// MCMC(model, { samples: 150000, kernel: 'MH', verbose: 1 });
Acceptance ratio: 0.03150666666666667
ERP:
    0.9 : 0.4918133333333333
    0.7 : 0.2974266666666667
    0.6 : 0.15076666666666674
    0.8 : 0.05999333333333335
run: 31226ms
````

````js
// MCMC(model, { samples: 500, burn: 50, kernel: { HMC: { stepSize: 0.002, steps: 200 } }, verbose: 1 });
Acceptance ratio: 0.9181818181818182
ERP:
    0.8 : 0.11800000000000006
    0.6 : 0.11200000000000009
    0.7 : 0.11000000000000008
    0 : 0.10800000000000004
    0.5 : 0.10400000000000004
    0.3 : 0.1
    0.9 : 0.09599999999999997
    0.4 : 0.08999999999999996
    0.2 : 0.08600000000000003
    0.1 : 0.07600000000000001
run: 28981ms
````
Are you imagining putting something like this in `/examples` or including it as a test case? If the latter, what shall we test? I don't know what the true posterior is exactly, but perhaps we can just check the marginal ERP against a histogram with a fairly large tolerance, with the intention of ensuring that the whole support (each bin) of the posterior is explored?

> What's the cause of the slowdown for MH and SMC?

AD, I assume. If that's right and we want to improve the situation, then I'd guess we could look at:

* AD in inference code, in particular look at avoiding creating tapes when not required for inference. (To avoid wrapping every single `tapify` call in a conditional, maybe `ad.tapify` could be the identity function when ad is not needed. Perhaps related: #136) There's also a bunch of ad math operators scattered throughout the code base, though I'm not sure avoiding those will make a huge difference so long as they aren't passed tapes.
* ERP scoring code. Can we use un-ADified versions of scorers when not doing e.g. HMC.
* Program code. Can we skip expanding AD macros when appropriate.

> Are you thinking of concrete ways in which it could break, or is this more of a general "unknown unknowns" concern? 

Avoiding the the unknown unknowns is the main motivation. It's hard to assess the risk you mention, but I'd be a little surprised if it works first time across all platforms and all possible `npm` configurations etc. (Having said that, I don't know how well the rest of webppl would fair in those situations either.) One thing in our favor is the fact that we can perhaps rely on anyone who is likely to try webppl been sufficiently motivated to open a ticket if they do run into problems.

stuhlmueller:
> `MH` doesn't do too bad with that model, but if you tighten the constraint further (`-100 => -5000` in this case) you see results like:

That's great! That's exactly what I was hoping would happen.

I'm surprised you chose to take 500 samples for HMC and 200 steps, as opposed to fewer (and maybe bigger) steps and a larger number of samples. Is that what worked best?

> Are you imagining putting something like this in `/examples` or including it as a test case?

I don't have strong opinions on this — I mainly wanted to see such an example before we merge the PR. That said, if we can reduce the runtime to about 10 seconds and still get reasonable results, it would be nice to include it as an inference test, with (as you propose) a lenient check on the distribution (maybe verifying that the probability of each element in the support is in `[.05, .2]`).

> > What's the cause of the slowdown for MH and SMC?
> 
> AD, I assume. If that's right and we want to improve the situation, then I'd guess we could look at:

I think it's worth checking how far we would get performance-wise if all we did was replace `ad.tapify` with the identity function. If this results in substantial gains, we should think about whether we can do this substitution conditional on `"HMC"` (not) appearing in the program code, similar to how we choose whether to apply caching at the moment.

> One thing in our favor is the fact that we can perhaps rely on anyone who is likely to try webppl been sufficiently motivated to open a ticket if they do run into problems.

I'd tend to rely on that for now; if we hear about problems, we should reconsider.

null-a:
> I'm surprised you chose to take 500 samples for HMC and 200 steps, as opposed to fewer (and maybe bigger) steps and a larger number of samples.

Make the steps much bigger and it doesn't work at all. This seems a general phenomenon with tight constraints, and it seems to make sense to me that this would be the case. If that's surprising to you, perhaps there's something to investigate.

I certainly wouldn't claim 200 steps is optimal. I tried a few different settings and just by eye-balling the marginal they all looked pretty similar, I just happened to try this last I think.

> I think it's worth checking how far we would get performance-wise if all we did was replace ad.tapify with the identity function. If this results in substantial gains, we should think about whether we can do this substitution conditional on "HMC" (not) appearing in the program code, similar to how we choose whether to apply caching at the moment.

I agree.

stuhlmueller:
> Make the steps much bigger and it doesn't work at all. This seems a general phenomenon with tight constraints, and it seems to make sense to me that this would be the case.

Yes. I was more surprised by the tradeoff between number of HMC steps and number of samples, since estimating a distribution on 10 elements from only 500 samples seemed likely to have a fair amount of variance. A priori, I would have expected (5000 samples, 20 steps each) to give lower-variance estimates.

null-a:
> A priori, I would have expected (5000 samples, 20 steps each) to give lower-variance estimates.

You may very well be right! Of course, there's a reason we don't (usually) go all the way and take just one step, and I really have no idea at all whether the best setting is closer to 20 or 200. FWIW, in Radford Neil's handbook chapter it says the following about setting L, the number of leap frog steps:

> Setting the trajectory length by trial and error therefore seems necessary. For a problem thought to be fairly difficult, a trajectory with L = 100 might be a suitable starting point.

Which if nothing else suggests that a large number of steps is *sometimes* a good idea. (Again, no claim that it's good in this case.)

stuhlmueller:
Interesting, thanks! (It would be interesting to explore these issues more systematically at some point – compute the true posterior, then run many simulations at many different parameter settings and visualize the error for each setting.)

null-a:
> I think it's worth checking how far we would get performance-wise if all we did was replace ad.tapify with the identity function.

I ran a few tests to check where we the extra run time comes from. As a very rough sketch I'd say it breaks down something like:

````
Creating tapes  | 40%
ADified scorers | 40%
Other           | 20%
````

Another reason it seems important to avoid always creating tapes is so that it remains possible to do inference with MH over models which pass numbers to external JS libraries. (I suppose it would be possible to manually `untapify`, but clearly we'd want to avoid that.)

> we should think about whether we can do this substitution conditional on "HMC" (not) appearing in the program code, similar to how we choose whether to apply caching at the moment.

I originally thought this too, but then I started to wonder whether doing this globally is the best approach. My main line of reasoning here is that if you have instances of both MH and HMC in one program, it seems unnecessary to create tapes during MH, just because you happen to do HMC elsewhere.

I have a working implementation of a more local method of deciding whether to create tapes [here](https://github.com/null-a/webppl/commit/12fe72cd6562cd781eae54a11b416e02576a3ede). The basic idea is to figure out whether tapes are needed by looking at which kernel we're using in the `MCMC` and `SMC` methods. What do you think of this approach?

stuhlmueller:
> Another reason it seems important to avoid always creating tapes is so that it remains possible to do inference with MH over models which pass numbers to external JS libraries.

Good point!

> I have a working implementation of a more local method of deciding whether to create tapes [here](https://github.com/null-a/webppl/commit/12fe72cd6562cd781eae54a11b416e02576a3ede). The basic idea is to figure out whether tapes are needed by looking at which kernel we're using in the `MCMC` and `SMC` methods. What do you think of this approach?

This looks like a much better idea. I didn't realize all `tapify` calls are happening within the MCMC/SMC functions such that this works.

I wonder if `this.ad` should be `this.adRequired` everywhere. That would make it clearer that it's a Boolean flag.

Do you have thoughts on how to get rid of the overhead from AD-ified scorers? Maybe the ERP transform could keep around the original scorers, and dispatch calls to score functions to the right type (AD-ified or not) based on whether the current coroutine has an `adRequired` property that evaluates to true?

null-a:
> this.ad should be this.adRequired everywhere

Definitely.

> Do you have thoughts on how to get rid of the overhead from AD-ified scorers?

I had a half-baked idea that we might be able to have an `adRequired` flag on `Trace`, and that we might be able to use that to delegate to the appropriate score function. Your way seems much better though, so I'll take a look at that.

null-a:
> Maybe the ERP transform could keep around the original scorers, and dispatch calls to score functions of the right type

Just to clarify, are you suggesting the `score` methods each become something like:

````js
function score(params, x) {
    return env.coroutine.adRequired ? originalScore(params,x) : adScore(params, x);
}
````

(I think that would work.)

stuhlmueller:
Yes, that's what I was thinking.

null-a:
> Yes, that's what I was thinking.

Great, thanks. I have a version working [here](https://github.com/null-a/webppl/commit/f2070c38c24216d7305d50eca931d78593bf11b2). I've not yet checked how much of the performance drop this recovers. I'll do that next.

null-a:
> I've not yet checked how much of the performance drop this recovers. I'll do that next.

The following shows the fractional slow down incurred by some of the models in `examples`. "HMC" refers to this PR in its current state, "HMC both scorers" refers to HMC with the additional change we've discussed in the last few comments. (i.e. Maintaining two score functions for each ERP.)

Model | HMC | HMC both scorers
---------|---------|------------------------
linearRegression | 0.27 | 0.1
lda (20k) | 0.18 | 0.09
logisticRegression | 0.19 | 0.1
hmmIncremental (n=10, full enum) | 0.01 | 0.02
ldaCollapsed (10k) | 0.81 | 0.74

(The reason ldaCollapsed is so much slower is that the `normalize` helper it uses is now written in webppl rather than JS. So there'll be some extra CPS overhead, and perhaps it's also affected by #174. I'll ignore this here.)

For the first 3 models listed it looks like this change helps a bit, although things weren't *that* much slower to begin with. Seeing these results, it's not 100% clear to me that it's worth the extra effort of keeping both scorers around. My main concern is the fact that the [code](https://github.com/null-a/webppl/blob/f2070c38c24216d7305d50eca931d78593bf11b2/src/transforms/adscorers.js) to transform the scorers is now pretty complicated and rather tightly coupled to the current structure of `erp.ad.js`, which could make future changes to `erp.ad.js` painful.

If we do think it's worth keeping both scorers around, perhaps I could revisit Sid/Daniel's original approach of splitting the scorers out into a separate file. (With the hope that transforming that file will be simpler, and perhaps delegation to the correct scorer can be added in the `ERP` constructor or something.)

It's also a drag having to thread `env` though all the chains of `require`s which end with `require('erp')`, though that's not such a big deal.

Any thoughts?

stuhlmueller:
Thanks for running these experiments! I agree that the additional complexity together with the relatively small improvement in speed doesn't make it worth keeping both scorers around, but it's good that we checked.

Suppose we take the performance hit and decide to merge the PR essentially as is. Then what else is left to do? Based on the original comment, it looks like we want to switch to the latest ad.js for tensor support. What needs to happen there? Let's aim to merge the PR this week, and let's create issues for any (non-critical) remaining issues.

null-a:
Updating to a newer version of ad.js is the only thing I have in mind to do. The main reason for doing so would be to gain the ability to take derivatives w.r.t. tensor valued variables. ([This hack](https://github.com/null-a/webppl/blob/65fb8a4606f320f0069f39d5629072eab6c3c62c/src/erp.ad.js#L241) avoids creating tapes for Dirichlet/mv Gaussian at present.)

I'm not sure how far away ad.js is from been ready for this, I'd need to check. Elsewhere I've played with integrating Daniel's implementation of ad, so that might be an option if getting this working soon is important. It's not a trivial change though, since (as far as I can tell) it also requires reimplementing some basic matrix/vector operations we currently get from numeric.js. (Some of which I've done, though it would be a stretch to get it finished this week.) There was some talk of having both these ad libraries share a common interface, an idea which might be relevant here.

On a different note, can I just check, are you OK with the change ([described here](https://github.com/null-a/webppl/blob/65fb8a4606f320f0069f39d5629072eab6c3c62c/src/erp.ad.js#L19)) to `ERP.prototype.support`? (It's used for constraint handling in HMC.)

stuhlmueller:
OK, keep me updated on the `ad.js` upgrade status.

If it looks like this is going to take a while, would it make sense to keep the PR as is, but detect when HMC is used with multivariate distributions and throw an error that says that it's not supported yet?

> On a different note, can I just check, are you OK with the change ([described here](https://github.com/null-a/webppl/blob/65fb8a4606f320f0069f39d5629072eab6c3c62c/src/erp.ad.js#L19)) to `ERP.prototype.support`? (It's used for constraint handling in HMC.)

Fine with me.

null-a:
I do think this will take a little while (at least a week, probably longer) so if you're happy to do so I think we should merge this. I've made a few final tweaks (including throwing an error if multivariate ERP are used with HMC as you suggested) so I think it's ready.

If we do merge it, I'll open a new issue for updating ad.js and also include the following on it:

* Move the version of `untapify` in `aggregation.js` into the ad package.
* Don't hardcode the path to the `ad.js` macros file.

stuhlmueller:
Sounds good. I'm going to merge this now—thanks for making this happen! :-)

Issue #202, opened 2015-09-18
Caching transform is applied to wppl files in packages but not header.wppl


null-a:
This came up when I tried to extract my linked list code (#174) into a package. When the linked list code is in `header.wppl` it works great, but once I put it in a package I run into 'out of memory' issues. I've tracked down the cause of this difference to the fact that the caching transform is applied to `*.wppl` files loaded from packages, but not `header.wppl` in core WebPPL.

Would it make sense for package authors to have the option to opt-out of the transform so that package code can be equivalent to code added directly to `header.wppl`?

stuhlmueller:
Yes, we could have a flag for this, and this may be a good short-term solution to this problem.

Longer-term, it is potentially worrisome that moving code between user-specified wppl and `header.wppl` results in such difference in performance for Incremental MH. What goes into the header is determined based on whether it will be useful across many programs. Now this decision is entangled with the decision what to cache, which may be correlated with the other criterion, but will likely come apart in some cases.

null-a:
I have an implementation of this [here](https://github.com/null-a/webppl/commit/e377c5e1a3b795309cf49e379957ebd53f72649a).

> Longer-term, it is potentially worrisome that moving code between user-specified wppl and header.wppl results in such difference in performance for Incremental MH.

Indeed. My best guess as to what's happening would be that all of the intermediate results (of these recursive functions which operate on lists) are been cached. (I have low confidence in this guess however.) If that was the case, then perhaps a sufficiently smart caching strategy could notice this.

stuhlmueller:
> I have an implementation of this [here](https://github.com/null-a/webppl/commit/e377c5e1a3b795309cf49e379957ebd53f72649a).

Nice, we should merge this.

PR #266, opened 2015-11-10
Support 'no caching' directive in header files

null-a:
This adds support for a `no caching` directive which allows header files to opt out of the caching transform. It was mentioned [here](https://github.com/probmods/webppl/issues/202#issuecomment-148183395) that we might want to merge this.

### DIFF
--- a/docs/packages.rst
+++ b/docs/packages.rst
@@ -36,6 +36,18 @@ You can automatically prepend WebPPL files to your code by added a
       }
     }
 
+The use of some inference algorithms causes a caching transform to be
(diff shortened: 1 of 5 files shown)


stuhlmueller:
Looks good. Thanks for writing thorough tests!

Can you add a line to the docs (maybe to "WebPPL code" under "WebPPL Packages") that documents this functionality?

null-a:
> Can you add a line to the docs

Sure, done! Look ok?

stuhlmueller:
Great, thanks!

Issue #131, opened 2015-06-30
Document language, core inference algorithms, and ERPs


stuhlmueller:
Possibly in part automatically: #87

null-a:
Keep in mind that we plan to change the interface to inference algorithms. (#86)

stuhlmueller:
One option is to write the docs in markdown, and then auto-build and host using [readthedocs](https://readthedocs.org/). This is [what Julia does](http://julia.readthedocs.org/en/release-0.3/).

stuhlmueller:
As part of this process, I'd also move some of the information that's currently in the README to the documentation.

null-a:
Core inference algorithms are now documented. Documenting ERPs is issue #87.

null-a:
@ngoodman [made a start](https://github.com/probmods/webppl/commit/722ca05dbd4167f331271d10282d5eeb50052ef9) on this.

chrisranderson:
Working on getting everything in `header.wppl` into the docs. I'll start with at least listing the function, then try to get explanations, then go for examples after that (if you'd like them).

null-a:
> @ngoodman made a start on this.

Looking at this, I get the impression that there are only a few bits we are short of for an initial pass of this. I'm thinking I could do the following:

**Introduction to sample, factor, marginalization**

We probably need to explain how sample, factor, distributions, inference fit together *somewhere*, but this already exists in [dippl](http://dippl.org/chapters/02-webppl.html) and [agent models](http://agentmodels.org/chapters/2-webppl).

The "Learning" section in "Quick Start" already points to the dippl version of this. We can perhaps make this a little more obvious by moving the installation instructions out of the quick start page. (As suggested by @stuhlmueller [here](https://groups.google.com/d/msg/webppl-dev/MMtFEhWYge0/tdT0eZkfBAAJ).)

Having skimmed through these, I think the agent models intro may be the most suitable of the two to link to eventually, but at the moment it still talks about `ERP`.

**Language reference**

Add a "language reference" page with the full list of JS syntax used in webppl. (Already exists in dippl.)

It seems useful (?) to have this around in a place that's easy to point people at. It probably doesn't need to be the first thing people see though.

**Reference docs for `sample`, `factor` etc.**

Perhaps we should have reference docs for `sample`, `factor`, etc. These would be similar to the current header function docs.

[...]

PR #271, opened 2015-12-02
Inference Documentation

null-a:
Here's a first attempt at some documentation for the inference algorithms. (#131)

I've not written docs for the async. particle filter or PMCMC as I'm not sure of their status (Are they used much? Are they sticking around?), but I'm happy to add them if that will be useful.

(It seems cumbersome to have everyone clone my repo and build the docs in order to preview the HTML for this change. I wonder if there's a slicker workflow for this kind of change.)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -8,6 +8,7 @@ Contents:
    
    quickstart
    usage
+   inference
    debugging
    packages
    development/index
--- a/docs/inference.rst
+++ b/docs/inference.rst
@@ -1,0 +1,245 @@
+Inference
+=========
+
+Enumeration
+-----------
+
+.. js:function:: Enumerate(thunk[, maxExecutions])
+
+   :param function thunk: Program to perform inference in.
+   :param number maxExecutions: Maximum number of executions to perform.
+   :returns: Marginal ERP
+
+   This method performs inference by enumeration. If ``maxExecutions``
+   is not specified exhaustive enumeration is performed. Otherwise,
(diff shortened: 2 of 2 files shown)


stuhlmueller:
When I build it, the styling doesn't quite look right. This is what I see:

screenshot 2015-12-02 08 15 15

Contrast to [this](http://spritejs.readthedocs.org/en/latest/api.html#redrawCallback). Do you see this, too? If so, do you know why there are extra `code` tags in our case?

Other than that, I have very minor content changes to make, but I'll do that directly on dev when it's merged.

> I've not written docs for the async. particle filter or PMCMC as I'm not sure of their status (Are they used much? Are they sticking around?), but I'm happy to add them if that will be useful.

I don't think we need to write docs for these. We should probably move these into packages. This reminds me that we should establish a standard for -- and provide instructions on -- writing tests for webppl packages (now #273).

> (It seems cumbersome to have everyone clone my repo and build the docs in order to preview the HTML for this change. I wonder if there's a slicker workflow for this kind of change.)

I'd be interested, too. I'm using the cumbersome workflow for now.

null-a:
> Do you see this, too?

No, I see the same styling that you linked to. Very long shot: `make clean` before `make html`? (I saw something with a build once and this cleared it up.)

stuhlmueller:
I can't get the right styling to work on my computer right now (`make clean` didn't help), but this shouldn't be an obstacle to merging this.

PR #272, opened 2015-12-02
Use grunt rather than make to compile for browser.

### LINKED ISSUE #248: Unify Makefile and Gruntfile
We probably shouldn't have two different build systems

null-a:
This adds a grunt task to run browserify and the browser tests, and removes the `Makefile` which did something similar. We now have:

* `grunt compile`
* `grunt test-browser`
* `grunt clean`

The compile task takes packages as args, so it's now easier to include packages in the browser bundle. For example you can now do:

````
grunt compile:path/to/webppl-viz:path/to/webppl-linked-list
````

Closes #248.


stuhlmueller:
Looks good!

How difficult would it be to have a basic test for the `grunt compile` pipeline? It would already be helpful to know that none of the commands involved returned an error code.

null-a:
Elsewhere, I have phantomjs + browser tests working on travis (#243), and as part of that `grunt compile` is run, so that should catch any show stoppers. Does that sound like it will do the trick? It's probably easiest for me to submit a follow-up PR for that as it also changes `Gruntfile.js`.

stuhlmueller:
That will work. Browser tests on travis will be really good to have in general.

null-a:
> That will work. Browser tests on travis will be really good to have in general.

Great, will submit a PR for that once this is done.

PR #276, opened 2015-12-03
Run browser tests in phantomjs on Travis.

### LINKED ISSUE #243: Test browserified webppl on travis-ci using PhantomJS

null-a:
This PR:

* Adds a new task `grunt test-phantomjs` which runs our (very simple) browser tests in phantomjs.
* Adds the config required to run this task on Travis.

The Travis VMs include v1.9.8 of phantomjs but WebPPL and the tests won't run under that, hence the manual install of v2.0.0.

We could get a lot of the benefit of this set-up (e.g. ensuring `grunt compile` works) by simply `require`ing `webppl.min.js` in node and running the tests from there. Maybe we should keep that in mind in case maintaining this phantomjs set-up becomes painful.

Closes #243.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,23 @@ node_js:
   - "4"
   - "5"
 
+env:
+  - TASK=default
+
+matrix:
+  include:
+    - node_js: "5"
+      env: TASK=travis-phantomjs
+
 install:
   - npm install
   - npm install -g nodeunit grunt-cli
+  - if [ $TASK = travis-phantomjs ]; then npm install -g browserify uglifyjs; fi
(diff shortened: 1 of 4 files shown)


stuhlmueller:
How about we switch to the [Ubuntu Trusty Travis environment](https://docs.travis-ci.com/user/trusty-ci-environment/) which  [has support](https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-153289245) for PhantomJS 2, and switch back to the current container-based one once they have support for it there as well?

null-a:
Good call. I looked at that thread but must have missed that comment, thanks.

Looks like it works, and may be faster.

Issue #269, opened 2015-11-14
beta sometimes returns NaN


longouyang:
( I'm on the dev branch -- 55e342a0393421ce1a87eda36211b8c7d8b057be -- although this bug is present on the current webppl.org version as well; I suspect that it's been with us since the very beginning )

Here's a fairly simple model:
```js
repeat(10, function() { return beta(0.001,0.001) });
```

With some troubling output:
```js
[ 1,
  1,
  9.235726936878123e-267,
  NaN,
  0,
  0,
  1,
  NaN,
  2.26873135676291e-25,
  NaN ]
```

For context, the beta sampler is:
```js
function betaSample(params) {
  var a = params[0];
  var b = params[1];
  var x = gammaSample([a, 1]);
  return x / (x + gammaSample([b, 1]));
}
```

The cause of the NaN error is that `x` and `x + gammaSample([b,1])` can both be 0 (0/0 gives NaN in Javascript). A simple patch would be:

```js
function betaSample(params) {
  var a = params[0];
  var b = params[1];
  var x = 0, y = 0;
  // if we hit x == 0 and y == 0, resample
  while (x == 0 && x + y == 0) {
    x = gammaSample([a, 1]);
    y = gammaSample([b, 1]);
  }
  return x / (x + y);
}
```

(in my informal testing, you can't just treat 0/0 as 1; it doesn't sample things in the right probabilities).
I'm hesitant to apply this fix because the better solution is to prevent `gammaSample` from returning 0, which I will soon work on.

More generally, I'm kind of alarmed at the lack of documentation in erp.js (e.g., which papers / websites do our implementations come from?) as well as the paucity of statistical sanity checks in the unit tests.

null-a:
> the better solution is to prevent gammaSample from returning 0, which I will soon work on.

Great. You've probably seen it, but for reference that is issue #132.

stuhlmueller:
A similar issue sometimes afflicts dirichlet:

    dirichlet([.0001, .0001])
    # [null,null]

This came up in [this mailing list discussion](https://groups.google.com/forum/#!topic/webppl-dev/jTi4INaFDfQ) where MH presented a model that has essentially this structure:

    MH(function(){
      var alpha = dirichlet([.1,.1])
      return dirichlet(alpha)
    }, 10)

longouyang:
Yes, although simply fixing the bug in gamma does not completely solve the dirichlet bug because we can still underflow when we normalize the sampled gamma values:

```js
[...]

PR #277, opened 2015-12-04
Gamma: fix underflow, add option to sample log values, add tests

### LINKED ISSUE #132: Underflow from common prior: Gamma(0.001,0.001)
A common prior used in Bayesian Data Analysis is Gamma(0.001,0.001). Here is it from Lee & Wagenmakers:

>The practice of assigning Gamma􏰂(0.001, 0.001) priors on precision parameters is theoretically motivated by scale invariance arguments, meaning that priors are chosen so that changing the measurement scale of the data does not affect inference. The invariant prior on precision λ corresponds to a uniform distribution on log σ, that is, p (σ2) ∝ 1/σ2, or a Gamma􏰂a → 0, b → 0􏰃 distribution. This invariant prior distribution, however, is improper (i.e., the area under the curve is unbounded), which means it is not really a distribution, but the limit of a sequence of distributions (see Jaynes, 2003). WinBUGS requires the use of proper distributions, and the Gamma(􏰂0.001, 0.001)􏰃 prior is intended as a proper approximation to the theoretically motivated improper prior. This raises the issue of whether inference is sensitive to the essentially arbitrary value 0.001, and it is sometimes the case that using other small values such as 0.01 or 0.1 leads to more stable sampling in WinBUGS. (Lee & Wagenmakers, *Bayesian Cognitive Modeling*, 2014, Chapter 4, Box 4.1)

In WebPPL,  `var lambda = gamma(0.001,0.001)` returns `0` within just a few samples.

longouyang:
This would close #132 

I fixed the gamma underflow issues and added an optional parameter to return log samples.

This looks correct, as I added a number of statistical tests (mean, mode, variance, kurtosis, skew, support) for parameter values across various orders of magnitude; these tests pass.

~~*The test harness could be a bit more flexible. Right now, I'm setting a single relative error tolerance for all statistics of a particular ERP parameter value, e.g., for `gamma(1,5)`, the same relative error tolerance is used for testing mean, for variance, for skew, and so on. This is too rigid, as sampling tends to get the lower moments more accurately. Right now I've resorted to setting a generous tolerance across the board. We should eventually patch this up but I wanted to get your feedback on the other parts of the fix.*~~
(more flexible error tolerances implemented)


stuhlmueller:
I haven't looked at this in detail yet (will wait for Paul's review), but here's a quick suggestion: if you haven't done so yet, run the tests many times — say 100x — to make sure that the tolerances are broad enough.

null-a:
I like it, lots of good stuff here. Here are my main comments. I have a few other minor comments, but I'll post a follow-up with those later.

*ERP testing framework*

I think this is definitely an improvement over what we're doing now. It looks like it's easy to add other samplers, so I think we should move over our existing sampler tests at some point. (For me, this could be after this PR is merged, and I'd be happy to do it myself if you like.)

One thing that I think needs looking at is the approach of taking the samples for all of the tests before any of the test cases run. I think this runs into a few problems:

* As far as I can tell, the random seed stuff doesn't really work because the samples are taken before the seed is set.
* I think we have all of the samples for all of the tests in memory at once, which I imagine will break down eventually.
* If you use `node unit test-erp.js -f ...` to run an individual test, you still   have to wait for all the samples to be taken.

I'm guessing that the reason it's this way is so that we can have individual test cases for each statistic, but I think it will be OK to lump these together as in the inference tests. If we do this, it might
also turn out that the tests can be written in a straight-forward way without needing to use `cache`?

Another thing that I think needs looking at is `util.warn`. This doesn't look like it will work as expected in the browser to me. (`global` is undefined, and `jsPrint` isn't available generally?)

*Underflow fix*

It seems to me that the `MIN_VALUE` trick is an improvement over what we have now. I don't know whether e.g. `gamma(0.001, 0.001)` is now a sensible prior to use, but returning `MIN_VALUE` seems better than returning `0` and the warning certainly seems useful. Andreas mentioned that he had reservations about this (in #132) so I guess he'll have something to add.

Sampling in log space seems to work well, although it's not entirely clear to me that this is enough to fix the original problem. In using this as a prior (e.g. on the precision of a Gaussian) won't the
sampled value still underflow once exponentiated, or can this be avoided?

Assuming it is sufficient, I think my preference would be to have a separate ERP for this rather than a `giveLog` option. (Primarily because I think this will be clearer/more explicit in webppl code.) Andreas suggested calling this `logGamma` in #132, although I wonder whether following log-normal it should be `expGamma/ExpGammaERP` instead?

longouyang:
> As far as I can tell, the random seed stuff doesn't really work because the samples are taken before the seed is set.

Oh -- right you are!

> I think we have all of the samples for all of the tests in memory at once, which I imagine will break down eventually.
> If you use node unit test-erp.js -f ... to run an individual test, you still have to wait for all the samples to be taken.
> I'm guessing that the reason it's this way is so that we can have individual test cases for each statistic, but I think it will be OK to lump these together as in the inference tests. If we do this, it might
> also turn out that the tests can be written in a straight-forward way without needing to use cache?

From a quick look, we might be able to use [nodeunit groups](https://github.com/caolan/nodeunit#groups-setup-and-teardown) to do this lumping and still retain the performance win of sample reuse.

> Another thing that I think needs looking at is util.warn. This doesn't look like it will work as expected in the browser to me. (global is undefined, and jsPrint isn't available generally?)

`global` is defined in the browser (see [node docs](https://nodejs.org/api/globals.html#globals_global)), though `global.jsPrint` is only defined on dippl (fwiw, when I browserify the current PR and use it in a copy of dippl, `util.warn` does work). It looks like, in the browser, I should prefer `jsPrint` but if it's undefined fallback to `console.warn`.

> Sampling in log space seems to work well, although it's not entirely clear to me that this is enough to fix the original problem. In using this as a prior (e.g. on the precision of a Gaussian) won't the
> sampled value still underflow once exponentiated, or can this be avoided?

I view returning log samples is only a partial solution. It gives the user some ingredients for working around underflow but isn't a full solution.

> Assuming it is sufficient, I think my preference would be to have a separate ERP for this rather than a giveLog option. (Primarily because I think this will be clearer/more explicit in webppl code.) Andreas suggested calling this logGamma in #132, although I wonder whether following log-normal it should be expGamma/ExpGammaERP instead?

Yes, I think expGamma is a better name.

null-a:
> From a quick look, we might be able to use nodeunit groups to do this lumping and still retain the performance win of sample reuse.

I would just have a test case for each combination of ERP and params, with a call to `test.ok` for each statistic. It's simple, you get to re-use samples, and the error messages will be fine if there's a failure. Any other way that works seems fine too!

> For node scripts, global is defined in the browser

Ah, yeah you're right about `global`. (Although it's browserify that defines it.) 

I'm not sure webppl should "know" about methods in dippl, that seems a little brittle to me.

I wonder if there's also a more general piece of work to do here, as there's presumably other information that's written to the console that would be useful to see in the main browser window. If so, I'd be tempted to just have `util.warn` call `console.warn` until we figure out the general thing.

>  It gives the user some ingredients for working around underflow but isn't a full solution.

Fair enough. I guess that for MHT fixing the underflow was only an instrumental goal, and what he really cares about is been able to use those priors, so if we close #132 I think we should probably open a new ticket for that.

longouyang:
@andreas Good idea.

I ran the tests 500 times:

```sh
cd webppl
for i in {1..500}; do echo $i; RANDOM_SEED=$i nodeunit tests/test-erp.js > test$i.txt; done
```
(as Paul noted, setting the seed doesn't work right now, but it should once we refactor)

Then looked at which cases tend to fail:

```sh
grep --context=3 AssertionError *.txt | grep gamma | sed -e 's/.*gamma//g' | freq | sort -rn
# freq is my copy of matt might's frequency script
# ht http://matt.might.net/articles/console-hacks-exploiting-frequency/
```

The failure cases are:

```
19 (0.0001,10000,false): variance
14 (0.0001,10000,false): mean
11 (100000,30,false): skew
3 (0.01,100,true): variance
3 (0.01,100,true): mean
2 (0.1,10,true): variance
2 (0.1,10,true): mean
```

I then tweaked the settings for the first three cases (bump 0.0001,10000,false variance to 0.5, mean to 0.2; bump 100000,30,false skew to 0.6) reran 500 tests (for the three cases only), and tests passed.

Also, I think that we might be able to avoid a lot of manual tolerance-setting. The statistics of the sample distributions are analytically tractable, e.g., the mean of the sample mean is just mu, the variance of the sample mean is sigma^2/n, the variance of the sample variance is mu_4/n - sigma^4(n-3)/(n*n(-1)). I think we can, for each statistic function, accept if the computed statistic is within m standard deviations of the expected statistic.

Btw, what are your thoughts on breaking out expGamma into its own erp?

stuhlmueller:
> I think we can, for each statistic function, accept if the computed statistic is within m standard deviations of the expected statistic.

Yes, we should do this, and set the tolerances such that any given test fails at most with probability 1/10000.

> Btw, what are your thoughts on breaking out expGamma into its own erp?

It's better than having a parameter, since the resulting distribution is quite different from a normal `gamma` distribution. I'm not sure how one would use this ERP though—when you say "It gives the user some ingredients for working around underflow", what do you have in mind? We should only add `expGamma` if we have at least one concrete story for how it would help.

> I'm not sure webppl should "know" about methods in dippl, that seems a little brittle to me.

Agreed, it shouldn't.

longouyang:
Alright, I've:

- Refactored with nodeunit groups such that similar tests are lumped together and the random seed works
- Made util.warn forget what it knows about dippl
- Implemented the automatic tolerances. For the vanilla samples, these work for everything except the mode. For the log samples, it doesn't work at all (yet?) because I need to know the moments of the distribution, and I couldn't find these for expGamma. This seems like another (small) reason to not do expGamma.

I don't have a compelling story for how to use expGamma (though I do have two decidedly non-compelling stories), so maybe we ought to drop it? At any rate, I believe that we will want an internal copy of it for sampling other distributions (e.g., it would certainly help in the fix to #269)

null-a:
Good stuff.

> so maybe we ought to drop it

FWIW, I agree that it makes sense to delay adding it until we have the other bits that will make it useful.

Here are those other minor comments I mentioned I had earlier:

* It would be nice if we could unify your `mean` and `variance` etc. with the `expectation` and `std` functions in `util`. At first glance this looks like a bit of a drag, in which case maybe I could tackle it later.

* I wonder if `suppressWarning` should be set in `setUp` and unset in `tearDown` to avoid any surprises in the future?

* Similarly (but less importantly) with `stackTraceLimit`. It feels like this should either be set and unset for specific tests, else we should apply it to all tests, not just those that happen to run after `test-erp.js`.

* I wonder if the `// for shape > 1` comment could be an assertion?

* I don't understand what's happening with the `// appease linter` lines. If I remove them `grunt lint` still works?

* We should move the duplicated `getRandomSeedFromEnv` function and any associated logic into `test/helpers.js` or similar.

* I'm intrigued by the comment about running out of memory with big arrays. My understanding is that v8 will back JS "arrays" with something like a C array where possible, and when I replace `Float64Array` with `Array` in `test-erp.js` I don't see an increase in memory use. For my own understanding, can you say something more about the problems you ran into?

* I have a suspicion that as written, the arrays of samples won't ever be garbage collected. This might not be an immediate problem, but it might become one as we add more tests.

* I don't think you're using your `require('fs')`.

longouyang:
> It would be nice if we could unify your mean and variance etc. with the expectation and std functions in util. At first glance this looks like a bit of a drag, in which case maybe I could tackle it later.

Though my mean and variance functions do caching through mutation, which may not be appropriate everywhere (particularly if array elements are added/removed/changed after the statistics computation).

> I wonder if suppressWarning should be set in setUp and unset in tearDown to avoid any surprises in the future?
> Similarly (but less importantly) with stackTraceLimit. It feels like this should either be set and unset for specific tests, else we should apply it to all tests, not just those that happen to run after test-erp.js.

Both done

> I wonder if the // for shape > 1 comment could be an assertion?

Done

> I don't understand what's happening with the // appease linter lines. If I remove them grunt lint still works?

The linter complains if I uncomment those lines (unused variable declarations) but I'll just remove the lines entirely -- less confusing.

> We should move the duplicated getRandomSeedFromEnv function and any associated logic into test/helpers.js or similar.

Done

> I'm intrigued by the comment about running out of memory with big arrays. My understanding is that v8 will back JS "arrays" with something like a C array where possible, and when I replace Float64Array with Array in test-erp.js I don't see an increase in memory use. For my own understanding, can you say something more about the problems you ran into?

I ran into memory allocation errors with larger arrays. Try this MWE:

```js
var repeat = function(constructor, n, f) {
  // used typedarray because node can run out of memory easily with lots of big arrays
  var a = new constructor(n);
  for (var i = 0; i < n; i++) {
    a[i] = f()
  }
  return a;
}

repeat(Array,        1e8, Math.random); // fails with memory allocation error
repeat(Float64Array, 1e8, Math.random); // works
```

> I have a suspicion that as written, the arrays of samples won't ever be garbage collected. This might not be an immediate problem, but it might become one as we add more tests.

I added some logic in the group-level `tearDown` to null out `samples` once all the tests are done. This should make gc work, right?

> I don't think you're using your require('fs').

Fixed

null-a:
Great.

 > I ran into memory allocation errors with larger arrays.

Gotcha. That's helpful, thanks.

I've had another look and had a few additional thoughts:

**Bug?**

The "in support" test doesn't look like it works as expected. (The `for` loop condition looks wrong.)

**Possible improvement**

I wonder if saying something like "rounded to nearest value in support" rather than "bumped up" in the underflow warning would be useful?

**Style**

These are things that *I* would look at, but it's less obvious that they are improvements. What you have works well, so if you don't want to do these and Andreas is happy then that's fine by me. I'm also happy to address them myself after this is merged if you/Andreas prefer.

>> It would be nice if we could unify your mean and variance etc. with the expectation and std functions in util

> Though my mean and variance functions do caching through mutation, which may not be appropriate everywhere

Right, but perhaps we could at least write the versions of `mean`, `variance`, etc. in `test-erp.js` in terms of `expectation`. I wonder if something this would work:

````js
var variance = cache("variance", function(a) {
  return expectation(a, function(x) { Math.pow(x - mean(a), 2); });
});
````
> This should make gc work, right?

Yes, that will work. The fact that we have to worry about this at all would lead me to try and write this a different way. The only thing that comes to mind is my original suggestion of having one test case for each combination of ERP + params.

Finally, I think I'd write [these](https://github.com/probmods/webppl/pull/277/files#diff-a3be47e75d2badb0422d03addc0ffdd0R357) `if`s differently. If nothing else I'd use `else if`.

**Comment by longouyang** [comment]
> The "in support" test doesn't look like it works as expected. (The for loop condition looks wrong.)

Ah -- good catch!

> I wonder if saying something like "rounded to nearest value in support" rather than "bumped up" in the underflow warning would be useful?

Changed to "rounded to nearest representable support value"

> Right, but perhaps we could at least write the versions of mean, variance, etc. in test-erp.js in terms of expectation.

I have a slight preference for not doing this, as the current versions seem more performant to me (e.g., they don't have to pass around big `Float64Array`s). If you/Andreas feel more strongly about this, then we should switch.

> Yes, that will work. The fact that we have to worry about this at all would lead me to try and write this a different way. The only thing that comes to mind is my original suggestion of having one test case for each combination of ERP + params.

I don't quite see how sample reuse works here.

> Finally, I think I'd write these ifs differently. If nothing else I'd use else if.

Done

(Misc: I've hidden the flag for generating log samples from end users)

**Comment by stuhlmueller** [comment]
1. Can you say more about how you plan to use the exp-gamma sampler? The `giveLog` option is a complication for the sampler and tests, so I want to be sure that we need it.

2. There's a lot going on in `test-erp.js`. I think we should consider splitting it up.

    First, very general functions such as `product`, `repeat`, `mean`, `variance`, and `sd` should probably not be defined in `test-erp.js` but rather in something like `util.js`. If the relevant existing functions in `util.js` (such as `std` and `expectation`) are not performant enough, we should try to replace them with more performant alternatives. We can also consider separating out statistics-related functions from `util.js` into a separate file. 

    Second, I'd also consider moving all gamma-related definitions to its own file. When we add tests for other ERPs to `test-erp.js`, this will become necessary to prevent the file from becoming bloated, so we might as well do it now.

    Overall, there is a fairly clear conceptual separation between (1) infrastructure for running ERP tests, (2) test definitions, (3) statistics functions, and (4) general-purpose helper functions. I think our module structure should mirror this separation. I haven't thought a lot about what the best structure is, but to give an example of what I'd consider an improvement over the current setup: `test-erp.js` only contains (1); (2) goes into `test-data/erps/gamma.js`; (3) goes into `src/statistics/gamma.js`; (4) goes into `src/util.js`.

**Comment by null-a** [comment]
> the current versions seem more performant to me

My guess would be that it won't make a great of difference.

I've opened #282 which might be useful.

> I don't quite see how sample reuse works here.

I was imagining something as simple as:

````js
function(test) {
  var samples = generateSamples();
  test.strictEqual(mean(samples), expectedMean);
  test.strictEqual(std(samples), expectedStd);
  test.done();
}
````
Of course it's more complicated than this because the function looks at the definitions to figure out which statistics to check etc., but would that structure work? Something [similar](https://github.com/probmods/webppl/blob/dev/tests/test-inference.js#L306) happens in the inference tests.

> We can also consider separating out statistics-related functions from util.js into a separate file.

Sounds good. This might already be part of your suggestion, but it might be nice to group all the math functions (`loggamma`, `digamma`, etc.) together in a single module at some point. HMC is going to complicate this though (because it uses macros to replace math operators with AD versions) so it might be best to leave the math functions in `erp.js` where they are for now.

**Comment by longouyang** [comment]
> Can you say more about how you plan to use the exp-gamma sampler? The giveLog option is a complication for the sampler and tests, so I want to be sure that we need it.

Currently we use the gamma sampler for beta and dirichlet. We can improve quality of those samplers by switching to the expGamma sampler.

Current beta, paraphrasing:
```js
var x = gamma(a,1);
var y = gamma(b,1);
return x/(x+y)
```
As noted in #269, underflow can cause x = 0, x + y = 0, leading to a return value of NaN. We can fix this by doing a little math to use the logged values:
```
x / (x+y)
= 1 / [ (x+y)/x ]
= 1 / [ 1 + y/x ]
= 1 / [ 1 + exp( log(y/x) ]
= 1 / [ 1 + exp( log_y - log_x) ]
```

Current dirichlet, paraphrasing:

```js
var theta_sum = 0;
var theta = new Array(n);
for(var i = 0; i < n; i++) {
  theta[i] = gamma(alpha[i], 1)
  theta_sum += theta[i]
}
for(var j = 0; j < n; j++) {
  theta[j] /= theta_sum;
}
```

We can use log values like this:

```js
var log_theta = new Array(n);
for(var i = 0; i < n; i++) {
  log_theta[i] = expGamma(alpha[i], 1)
}
var log_theta_sum = util.logsumexp(log_theta);

for(var j = 0; j < n; j++) {
  log_theta[j] -= theta_sum;
}

var theta = new Array(n);

for(var k = 0; k < n; k++) {
  theta[k] = Math.exp(log_theta[k])
}
```

(It'd help to get a double check of my math)

longouyang:
Next todos:

- [ ] Rewrite my mean, variance, etc functions in terms of existing utils
- [ ] Split up test-erp.js into (1) infrastructure for running ERP tests, (2) test definitions, (3) statistics functions, and (4) general-purpose helper functions

stuhlmueller:
The plan for improving the beta and dirichlet samplers looks good to me. This definitely justifies including the expGamma sampler.

I don't know if this has been clear from previous messages, but I do think it would be clearer to break out `expGamma` into its own ERP. This would allow us to split up the test definitions so that we avoid the `if (giveLog)` conditionals in various places.

null-a:
> This definitely justifies including the expGamma sampler

>  I do think it would be clearer to break out expGamma into its own ERP. This would allow us to split up the test definitions

It sounds like we think that an expGamma sampling routine is useful, but we also think that an expGamma ERP is not. So maybe `test-erp.js` should in effect be `test-samplers.js` where we can happily test an internal expGamma sampling routine without adding a corresponding ERP. (Though the fact that the sampling routines aren't exported other than as properties of ERP would need thought, with the obvious solution been to split out the core sampling routines from `erp.js`.)

stuhlmueller:
I'd be happy with this solution. I also don't think it would be a problem to have an `expGamma` ERP even if we only use its sampling routine.

longouyang:
> I was imagining something as simple as:
> 
> ```js
> function(test) {
>   var samples = generateSamples();
>   test.strictEqual(mean(samples), expectedMean);
>   test.strictEqual(std(samples), expectedStd);
>   test.done();
> }
> ```

@null-a If we write it this way, we lose the ability to run very fine grained tests with nodeunit (e.g., `RANDOM_SEED=1234 nodeunit tests/test-erp.js gamma - 1,1,true - mean`). I have found this to be useful while developing; do you think the tradeoff is worth it?

null-a:
> do you think the tradeoff is worth it?

I would guess so. I'm thinking that:

* We have enough control to avoid wasting time collecting lots of irrelevant samples. This seems to be the thing we want most from focussed tests.
* If there is a failure, we will still be able to get detailed enough errors to know which (sub)test failed. Inference tests work this way, and it doesn't seem to be a problem.

Of course, if your experience actually using this suggests otherwise, I'm more than happy to do whatever you think is best. We can always correct course later if need be.

longouyang:
Okay - I've organized the code as Andreas suggested and rewritten `mean`, `var`, etc in terms of `expectation` (currently using a local copy of `expectation` that Paul wrote in #282, we'll want to rewire things after merging that PR). So I believe this PR is merge-able now.

stuhlmueller:
Thanks Long! Just one more thing: the test results (e.g. [here](https://travis-ci.org/probmods/webppl/jobs/100024205)) are a bit verbose — can you change the settings so that we don't print out expected/actual/tolerance, at least as part of the CI tests?

null-a:
Looks good!

One question: I'm not sure I understand the situation with `gammaERP`. Its `sample` method no longer takes the `giveLog` argument but its `score` method does. Am I missing something?

(Also, the `gamma` helper in the header still passes `giveLog=false`, and the comment above `gammaSample` still mentions `giveLog`.)

> currently using a local copy of expectation that Paul wrote in #282; we'll want to rewire things after merging that PR

Great, I'll take care of that.

longouyang:
Ah, thanks Paul -- I've removed all remaining `giveLog` stuff (and moved the giveLog version scorer to `expGammaScore` for now).

Andreas -- I've silenced the expected/actual/tolerance stuff (now it only prints if the test fails)

PR #279, opened 2015-12-11
Enumeration: exit early if we fail to satisfy a hard factor

longouyang:
I discussed a model that relies on this behavior today with Andreas and he mentioned that this should be a PR; we currently do it for rejection sampling so it makes sense to do it for enumeration too.

Should we, as with rejection, hide this behind the `incremental/maxScore` options?


null-a:
> so it makes sense to do it for enumeration too.

Great, I think so too!

> Should we, as with rejection, hide this behind the incremental/maxScore options?

I don't think we need an option as this is always correct, right?

Rejection always bails if the score drops to `-Infinity`. The `incremental` option gives us extra information, which allows us to be more aggressive and bail out when the score drops below some threshold.

A couple of thoughts:

* Is it possible to turn the model you have in mind into a simple test case? If not, perhaps we could have a test in which we `assert.ok(false)` after a `factor(-Infinity)` statement. (I did something similar [here](https://github.com/probmods/webppl/blob/dev/tests/test-data/stochastic/models/variableSupport.wppl).)
* `==` should be `===` I think. I'm not sure whether the type conversion of the former would ever trip us up here, but using `===` unless you explicitly want the conversion is a good way to go. All other similar checks that we already have use `===`.

I'm happy to take care of these if you like.

null-a:
That test looks like it will pass even without this change, right? We really want a test which fails without this change.

longouyang:
D'oh. Okay -- I've made the test fail without the change.

I think I got a little confused because I didn't understand variableSupport.wppl:

```js
var model = function() {
  var allTrue = repeat(poisson(10) + 1, trueF);
  var val = uniformDraw(allTrue);
  // Simulate error arising from execution continuing with zero probability
  // trace.
  assert.ok(val === true);
  return val;
};
```

Is the comment just out of date? It seems like samples from allTrue are always true, so the assertion would never have a chance of failing.

null-a:
> I think I got a little confused because I didn't understand variableSupport.wppl:

Ah, sorry for the confusion! There was a bug in MH which would have led to the assertion failing like so:

1. `poisson(10)` is resampled and decreases, so the array `allTrue` gets shorter.
2. The index sampled by `uniformDraw` is reused. If the array shrinks enough, then `val` becomes `undefined ` because we index off the end of the array.

Hope that helps.

longouyang:
Ah, got it.

This PR now has stricter equality checking and the improved test. Is there anything else to be done?

PR #280, opened 2015-12-11
[grunt] default: lint before nodeunit

longouyang:
So that Travis tests that only fail linting fail faster.


null-a:
This seems reasonable to me. Since there are conflicts with this PR which need resolving, I've created (and will merge) #281 which makes exactly this change.

PR #282, opened 2015-12-15
Updates to expectation and std util funcs.

null-a:
The motivation for this change is the desire to be able to use `util.expectation` in #277. The main things in the way of that are the fact that `expectation` ignores its `func` arg when operating on an arrays and it won't work with `Float64Array`s. Both of those are fixed here. I've also made a few other tweaks while I was doing this.

* hists in inference tests are now represented as lists of pairs to avoid doing arithmetic
  with the keys of objects which are strings not numbers.
* Remove unnecessary normalizeHist from `test-inference.js`. (Marginals are already normalized.)

### DIFF
--- a/src/erp.js
+++ b/src/erp.js
@@ -620,6 +620,7 @@ function makeMarginalERP(marginal) {
   });
 
   dist.MAP = function() {return mapEst};
+  dist.hist = marginal;
   return dist;
 }
 
--- a/src/inference/mh-diagnostics/diagnostics.js
+++ b/src/inference/mh-diagnostics/diagnostics.js
@@ -1,4 +1,4 @@
-var util = require('../../util');
+var stats = require('../../statistics');
 var fs = require('fs');
 var os = require('os');
 
@@ -15,8 +15,8 @@ function geweke(traces, first, last, intervals) {
   for (var i = 0; i < traces.length / 2; i = i + Math.floor((traces.length / 2) / (intervals - 1))) {
     var firstSlice = traces.slice(i, i + Math.floor(first * (end - i)));
     var lastSlice = traces.slice(Math.floor(end - last * (end - i)), traces.length);
-    var mu = (util.mean(firstSlice) - util.mean(lastSlice));
-    var zscore = mu / Math.sqrt(Math.pow(util.std(firstSlice), 2) + Math.pow(util.std(lastSlice), 2));
+    var mu = (stats.mean(firstSlice) - stats.mean(lastSlice));
+    var zscore = mu / Math.sqrt(stats.variance(firstSlice) + stats.variance(lastSlice));
     zscores.push([i, zscore]);
   }
   return zscores;
--- a/src/util.js
+++ b/src/util.js
@@ -134,22 +134,6 @@ function cpsIterate(n, initial, func, cont) {
       function() { return cont(val); });
 }
 
-function expectation(a, func) {
-  assert.ok(a.length > 0);
-  var f = func || _.identity;
-  return _.reduce(a, function(acc, x) {
-    return acc + f(x);
-  }, 0) / a.length;
-}
-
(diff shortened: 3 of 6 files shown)


stuhlmueller:
I think it's a good idea to have a separate `expectation`/`std` for arrays and distributions. I also like not doing arithmetic on string keys :-).

The representation of hists in terms of lists of pairs seems to address the same problem that is addressed by the data structure assumed by [`makeMarginalERP`](https://github.com/probmods/webppl/blob/11cd4ec252cd9112e911307a31c4adadf088ca3f/src/erp.js#L530), namely an object with stringified values as keys and `{val: ..., prob: ...}` objects as values. (Note that the comment above the function definition is inaccurate.) Indeed, [`logHist`](https://github.com/probmods/webppl/blob/dev/src/util.js#L70) in `util.js` assumes that representation. 

Would it make sense to standardize all the hist functions in `util.js` so that they operate on this kind of data structure? We could even create a prototype for this object and turn `expectation`, `std`, comparison (currently `histsApproximatelyEqual`), and conversion to log into methods.

null-a:
> Would it make sense to standardize all the hist functions in util.js so that they operate on this kind of data structure?

I considered it, but decided what we had was tolerable. In part because `histExpectation` and `histsApproxEqual` etc. are only called from `test-inference.js`, so this isn't something which is complicating the whole code base. Of course, I'm happy to look at it if we think it's worth while.  A couple of thoughts relevant if we do:

* `getHist` in `test-inference.js` reconstructs the histogram from the marginal ERP. We could simplify this by making the histogram available as a property on the ERP. (I don't think this costs us anything as the marginal ERP closes over the hist already.)
* We already have a prototype for the histogram data structure: [`Histogram`](https://github.com/probmods/webppl/blob/875812b2eb538098b248b5eec214925465c70301/src/aggregation.js#L6). While we could add `expectation` etc. to this, I wonder whether inference tests should instead be calling the (non-existent) `expectation` method directly on the marginal ERP. (In which case `expectation` in `header.wppl` could delegate to `ERP.proto.expectation`, much as `MAP` and `entropy` already do.)

stuhlmueller:
> We already have a prototype for the histogram data structure: [`Histogram`](https://github.com/probmods/webppl/blob/875812b2eb538098b248b5eec214925465c70301/src/aggregation.js#L6).

So (with this PR) we'd have three data structures for histograms! :-)

- Lists of pairs
- Objects with stringified values as keys and `{val: ..., prob: ...}` objects as values
- `Histogram` in `aggregation.js`

These should either be consolidated, or it needs to be clearer (perhaps from changed names) how their functionality differs. (Or some of both.)

> I wonder whether inference tests should instead be calling the (non-existent) `expectation` method directly on the marginal ERP.

This could be good.

> (In which case `expectation` in `header.wppl` could delegate to `ERP.proto.expectation`, much as `MAP` and `entropy` already do.)

Wouldn't there be an issue with cps'ed webppl functions (which would be passed in for `header.wppl`) vs normal functions (which would be passed in for the tests)?

null-a:
> Wouldn't there be an issue with cps'ed webppl functions

Oops, yes, of course!

> we'd have three data structures for histograms
> These should either be consolidated

I've added a `hist` property to marginal ERPs which means we avoid having to recreate the histogram from the marginal ERP in inference tests. This is simpler, and removes the list of pairs representation.

The other 2 are really the same data structure, its just that the `Histogram` prototype isn't used everywhere yet. I've opened an issue (#284) for that.

Also, I've removed the duplicated `expectation` function as discussed in #277. `mean`, `sd`, etc. all now live in `statistics.js` rather than `util.js`. There is only one place outside of `test-samplers.js` where these are used, and that's been updated accordingly.

I think this is good to merge.

stuhlmueller:
Looks good, thanks!

Issue #109, opened 2015-06-01
makeMarginalERP should take log probabilities


stuhlmueller:
If we require exponentiation before the construction of marginal erps, low probabilities will get rounded to 0.

stuhlmueller:
This is a problem in practice whenever unnormalized scores are low. For example:

    Enumerate(function(){
      var x = flip(.5);
      factor(-1000);
      return x
    })

    => ERP with NaN scores!

Using logsumexp in `makeMarginalERP` would fix this.

PR #287, opened 2016-01-15
Use Histogram everywhere

### LINKED ISSUE #284: Use the Histogram prototype in all inference algorithms
[`Histogram`](https://github.com/probmods/webppl/blob/875812b2eb538098b248b5eec214925465c70301/src/aggregation.js#L6) is only used in `MCMC` and `SMC` but could be used for other inference algorithms.

Using it in sampling algorithms will be straight-forward I think. A little work will be needed to get it working in `Enumerate` since:

*  At present all values added to `Histogram`are given equal weight.
* `Enumerate` does some [extra work](https://github.com/probmods/webppl/blob/5988c5e0a39ee4666953f6ed1407e01f4e30a408/src/inference/enumerate.js#L128) to avoid underflow. (History: #109.)

null-a:
The main change here is a switch to using `Histogram` (and other 'aggregators') in all inference algorithms, which closes #284.

This slightly changes the MCMC interface when `justSample` or `onlyMAP` are used as described [here](https://github.com/probmods/webppl/issues/284#issuecomment-171010017).

This also includes a few additional changes which seemed reasonable to tackle at the same time. The main benefits of these are:

* `makeMarginalERP` no longer adds a `MAP` method to the returned ERP as there is already [a similar method](https://github.com/probmods/webppl/blob/2a5775e6251c9e233a30e2ddd39c0d44b4d1579b/src/erp.ad.js#L43) on the ERP prototype. The only reason I can think of to keep the `makeMarginalERP` specific [version](https://github.com/probmods/webppl/blob/2a5775e6251c9e233a30e2ddd39c0d44b4d1579b/src/erp.ad.js#L624) around would be if it were noticeably more performant, but I don't think that's likely to be the case.
* Normalization now happens in the aggregators. Previously all marginals were normalized in log space. With this change, we can continue to do this for the results of `Enumerate` (#109), without having to do so for sample counts. (Where it's unnecessary, right?) There's also yields an opportunity to accumulate the normalizing constant as the histogram is built, though I've not done that yet.
* As a result, `makeMarginalERP` is now quite a bit simpler.

### DIFF
--- a/src/aggregation.js
+++ b/src/aggregation.js
@@ -13,17 +13,17 @@ Histogram.prototype.add = function(value) {
   var value = untapify(value);
   var k = util.serialize(value);
   if (this.hist[k] === undefined) {
-    this.hist[k] = { prob: 0, val: value };
+    this.hist[k] = { count: 0, val: value };
   }
-  this.hist[k].prob += 1;
+  this.hist[k].count += 1;
 };
 
 function normalizeHist(hist) {
-  var norm = _.reduce(hist, function(acc, obj) {
-    return acc + obj.prob;
+  var totalCount = _.reduce(hist, function(acc, obj) {
+    return acc + obj.count;
   }, 0);
   return _.mapObject(hist, function(obj) {
-    return { val: obj.val, prob: obj.prob / norm };
+    return { val: obj.val, prob: obj.count / totalCount };
   });
 }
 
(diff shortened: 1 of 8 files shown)


stuhlmueller:
Neat, a PR with more deletions than additions! :-)

- Does it matter for HMC that we moved the normalization arithmetic that previously happened in `erp.ad.js` (in `makeMarginalERP`) to `aggregation.js`? For example, what if we're doing HMC over Enumeration, as in [this example](https://gist.github.com/stuhlmueller/4bf0c2e5e0f30aef15a8)? I assume that we don't support this case right now, and so don't think any changes are warranted, but this may be worth keeping in mind for the future.
- I'm a little worried that someone will pass an unnormalized distribution into `makeMarginalERP`. I wonder if we should add a line to check that it's normalized, such as `assert.ok(_(marginal).map(function(v){return v.prob}).sum() < epsilon)`.

null-a:
> I assume that we don't support this case right now

That's right. It's a good point though, and one I'd not considered. I think it's still worth going with this PR for now, we can worry about nested inference later I think. Perhaps it will turn out that this change helps us by factoring out just that code which needs to be transformed for ad.

 > I'm a little worried that someone will pass an unnormalized distribution into `makeMarginalERP`.

I'm not sure this is very likely to happen. `makeMarginalERP` is only called from `aggregation.js`, and if you remove normalization from there the inference tests fail spectacularly. I suppose someone could add a new inference algorithm and not use `aggregation.js`, but as long they add the standard inference tests (e.g. `simple.wppl`) I think the problem will be noticed.

Nevertheless, I've added the assertion, I can always revert it if need be.

null-a:
The CI failure was unrelated to this PR. It was an HMC test which ran OK but the result was outside of the test tolerance.

stuhlmueller:
> I'm not sure this is very likely to happen. `makeMarginalERP` is only called from `aggregation.js`, and if you remove normalization from there the inference tests fail spectacularly.

I think some of my worry was because I vaguely remember having used `makeMarginalERP` before (maybe as part of the coarse-to-fine project), and I don't think I made sure that what I passed in was normalized. I agree that it's not a big issue.

> The CI failure was unrelated to this PR. It was an HMC test which ran OK but the result was outside of the test tolerance.

Let's adjust the tolerance then (in another PR).

PR #288, opened 2016-01-15
Back cache helper with LRU cache.

null-a:
This uses a LRU cache to back `cache`. It uses the `lru-cache` package which is implemented with a hashmap & doubly-linked list, so I think the asymptotics of `cache` will be unchanged.

This was the main change which helped @mhtess with [this issue](https://groups.google.com/d/msg/webppl-dev/QCO3sDSekzI/Y4Dag1KNBQAJ). The underlying problem there was that Node was running out of memory. This change helps because the model has an instance of the RSA model using `cache` nested within MCMC, where one of the args to one of the cached functions was real-valued (or similar), causing the store used by `cache` to continue growing throughout inference. After this change, memory usage was constant (or close to), and inference wasn't any slower.

There may have been other ways to fix this particular problem, but limiting the amount of memory used by `cache` seems like a reasonable thing to in general.

I've set the default `maxSize` of the store to 10K. MHT was using 1K in the results I mention above, but I bumped that up a little as it felt too conservative.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "esutils": "^2.0.2",
     "git-rev-2": "^0.1.0",
     "immutable": "^3.7.5",
+    "lru-cache": "^4.0.0",
     "minimist": "^1.2.0",
     "numeric": "^1.2.6",
     "priorityqueuejs": "~1.0.0",
--- a/src/headerUtils.js
+++ b/src/headerUtils.js
@@ -15,7 +15,7 @@ module.exports = function(env) {
   // caching is across all uses of f, even in different execuation
   // paths.
   function cache(s, k, a, f, maxSize) {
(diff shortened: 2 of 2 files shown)


null-a:
I forgot to mention: I've tried this out in the browser and it seems to work fine.

stuhlmueller:
Looks very useful. I wonder if the default cache should be unlimited, though. We could print out a warning message if the cache grows very large. What do you think?

null-a:
> I wonder if the default cache should be unlimited

Sure, done! The warning is issued once the cache has 10K entries which would probably have helped MHT, and doesn't seem unreasonable in general, though it may well need tweaking over time of course.

stuhlmueller:
Thanks!

Issue #70, opened 2015-03-25
Refactor and document how to include webppl boxes and scripts in web pages


stuhlmueller:

null-a:
There's possibility this feature could be extracted as a package/module. (#123)

longouyang:
I've been working on refactoring the webppl boxes code in a package: [webppl-editor](http://github.com/probmods/webppl-editor)

The existing results rendering code is not very responsive because there's never time to render to the DOM. So even if the call to build graphs / print statements is made at the beginning of the webppl code, it doesn't show up for the user until after webppl execution.

So I decided to place model execution in a Web Worker (as well as write code editors as React components). This appears to work well and feels more responsive than the existing editor. Can you guys play around with [this demo](http://web.stanford.edu/~louyang/wp-editor/index.html), peruse the code, and offer feedback?

A list of changes:

- Because webppl happens on a different thread, you can now click run on multiple code boxes (it's currently all queued through a single worker though)
- For graphics, I switched from highcharts to vega (to be consistent with Erin's webppl-viz package, which Noah would like to integrate)
- You can now right click rendered graphics and download them to disk.

Todos:

[ ] Switch to the browser.js versions of webppl run and compile methods (so that packages included in webppl get bundled)
[ ] Get code folding working
[ ] Get bracket matching working
[ ] Get vanilla JS execution working
[ ] Translate the dippl Draw functions so that the computer vision examples work with the new editor setup (this might be hard?)
[ ] More robust stringifying, error catching (so that, e.g., just typing `flip` and clicking run returns something like `Function flip() { ... }` rather than `DataCloneError`
[ ] Write a `density()` visualizer?
[ ] Implement an inference progress interface? This would further improve the user friendliness of the editor.
[ ] At some point I had dynamic stuff working through vega (e.g., showing an exact count when you hover over a histogram bar) but then it mysteriously stopped; try to get this working again
[ ] Write the literate code editor? (i.e., the widget that lives at webppl.org) Though maybe this could just be bespoke code for webppl.org

stuhlmueller:
I like the web workers idea and resulting responsiveness!

I'd try to get rid of the jQuery dependence. React works best if the DOM state is a pure function of the state of the React components, but this is no longer true if you manipulate it using jQuery. This can easily lead to difficult-to-debug bugs down the road, and I'm not convinced it's necessary here. For example, the `run` button should just be its own component, with updates happening as a consequence of its `props` changing.

longouyang:
[...]

PR #289, opened 2016-01-16
Factor out runTrampoline; use "pausy" version in browser

longouyang:
In the browser, we can mimic the responsiveness of running webppl in a Web Worker by using a trampoline that takes occasional breaks (here, every 100ms) to allow DOM updates to happen.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -44,26 +44,13 @@ function webpplNaming(code) {
   return escodegen.generate(newProgramAst);
 }
 
-var runTrampoline = function(t) {
-  var lastPauseTime = Date.now();
-  while (t) {
-    var currTime = Date.now();
-    if (currTime - lastPauseTime > 100) {
-      return setTimeout(function() { runTrampoline(t) }, 0);
-    } else {
-      t = t();
-    }
-  }
-}
-
-global.runTrampoline = runTrampoline;
-
 global.webppl = {
   run: run,
   compile: compile,
   cps: webpplCPS,
   naming: webpplNaming,
-  analyze: analyze
+  analyze: analyze,
+  runTrampoline: require('./transforms/trampoline').runner
 };
 
 console.log('webppl loaded.');
--- a/src/browser.js
+++ b/src/browser.js
@@ -44,6 +44,20 @@ function webpplNaming(code) {
   return escodegen.generate(newProgramAst);
 }
 
+var runTrampoline = function(t) {
+  var lastPauseTime = Date.now();
+  while (t) {
+    var currTime = Date.now();
+    if (currTime - lastPauseTime > 100) {
+      return setTimeout(function() { runTrampoline(t) }, 0);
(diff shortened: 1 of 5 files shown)


stuhlmueller:
How does performance differ between the two trampolines?

longouyang:
I described some measurements in issue #70 but I'll reproduce them here:

I measured total computation time for a very simple model:
```
util.seedRNG('a')
repeat(5e4, flip)
```

Time spent in:
- Old trampoline: M = 1357 ms, SD = 18 (N = 5)
- Pausy trampoline: M = 1340 ms, SD = 54 (N = 5)

I also measured a real model, the final dippl computer vision example..

Timings:
- Old trampoline: M = 3865 ms, SD = 204 (N = 6)
- Pausy trampoline: M = 3846 ms, SD = 154 (N = 6)

stuhlmueller:
I'm surprised, I would have expected a substantial difference. This is great news! (The pessimistic reading is that other factors slow down our computations enough that we don't notice the additional slowdown.)

Is there some way to do this PR without introducing a global `runTrampoline` variable? For example, `webppl.run`/`compile` could take the `runTrampoline` function as an additional optional argument, defaulting to the standard trampoline, but called with the new trampoline function from `browser.js`. `trampoline.js` would then be responsible for generating code that includes the code of the trampoline function that is passed in. I'm not sure this works out, but I think we should try to avoid introducing globals as much as possible.

longouyang:
The extra stuff in the pausy runner isn't very expensive.. we just check the time at every step and if the time since the last pause exceeds 100ms, we do a `setTimeout` (of delay 0) on continuing `runTrampoline`, giving the browser a chance to update the DOM.

On my machine, `Date.now()` can run 13 million times per second (see [this jsperf](https://jsperf.com/new-date-vs-date-now-vs-performance-now/37)), so the measurements aren't so surprising.

In the browser, `runTrampoline` needs to be -- in some way or another -- global because we need it as an escape hatch for libraries (e.g., the `loadImage` function that dippl defines for the computer vision examples). The browser version already defines a global called `webppl`, though, so maybe we can attach `runTrampoline` there?

stuhlmueller:
> The browser version already defines a global called `webppl`, though, so maybe we can attach `runTrampoline` there?

Yes, that would be a good solution.

longouyang:
Okay, I removed all uses of global variables for providing `runTrampoline` and, in the browser, attached it to `webppl`

stuhlmueller:
This looks good to me now.

Note for posterity: If we ever want browser tests that run longer than 100ms, we'll have to change our test architecture. With the current one, the `setTimeout` will prevent `QUnit` from seeing the assertions in time.

longouyang:
For browser tests, we could always switch to the vanilla trampoline runner.

stuhlmueller:
If it's easy to do, switch it now so that we don't forget about it?

PR #290, opened 2016-01-18
Watchify grunt task

null-a:
This adds a `grunt watchify` task as requested by @longouyang. When changes are detected this only recompiles `webppl.js`, it doesn't run uglifyjs. (Which would be much slower than the incremental compilation step.) Packages can be included in much the same way as with `grunt compile`. e.g. `grunt watchify:webppl-timeit`.

I've also split the `compile` task into `browserify` and `uglify` tasks. I thought that been able to run `uglify` after `watchify` might be useful.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -2,7 +2,7 @@
 
 var _ = require('underscore');
 var open = require('open');
-var execSync = require('child_process').execSync;
+var child_process = require('child_process');
 
 var jslintSettings = {
   options: {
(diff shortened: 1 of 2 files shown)


stuhlmueller:
Nice! This does speed up re-compilation a lot.

Add a brief note to the [Compiling for browser](http://webppl.readthedocs.org/en/dev/development/workflow.html#compiling-for-browser) docs that describes the `watchify` functionality, and `npm install -g watchify` to the [Installation from Github](http://webppl.readthedocs.org/en/dev/development/install.html) instructions?

null-a:
I've added something to "Compiling for browser". I didn't add anything to the "Installation from Github" section as that didn't already mention installing `browserify` or `uglifyjs`. What we have isn't perfect, but I figured it was good enough -- see what you think.

stuhlmueller:
Makes sense, thanks!

PR #291, opened 2016-01-19
More flexible trampoline runner selection

longouyang:


longouyang:
In PR #289, Andreas noted that our browser-based tests would be hamstrung by the pausy trampoline runner (asserts would not be seen in time if test takes > 100 ms).

Here, I've made it so you can force the trampoline transform to use a particular runner by setting the global variable `forceTrampoline` to be either `cli` (non-pausy) or `web` (pausy). I've changed the browser test file to use `cli`.

stuhlmueller:
Could we pass the `forceTrampoline` option in as part of the `options` argument to [`webppl.run`](https://github.com/probmods/webppl/blob/dev/src/main.js#L175) and `webppl.compile`? This is a little more effort to implement, but I think it's a cleaner way to deal with transform parameters.

longouyang:
Okay -- I made that change.

One consequence of this is that now the trampoline transform is parameterized by runner. This changes the calling interface a little. Before, you could just do

```js
var trampolineTransform = require('./src/transforms/trampoline').trampoline;
```

but now you have to do:
```js
var trampolineTransform = require('./src/transforms/trampoline').trampoline({runner: require('./src/main').trampolineRunners['cli']})
```

I had to make some design decisions:
- Above, `runner` is a *function*. The trampoline transform takes this function, runs its `toString()` method, and inserts the result into the generated code.
- I decided to store the list of all trampoline runners in `main.js` because that's now the point where we decide which runner to use.

These decisions could use another pair of eyes, though.

stuhlmueller:
Thanks!

Why not select the runner by providing a string (defaulting to `cli`, with `web` as alternative) to the `trampoline` function? I'm imagining that `trampoline.js` would export both trampoline runners under different names, and `webppl` would re-export both as well (without something like `getTrampolineRunner` that automatically selects the appropriate one).

Are there reasons not to go down this path? I think it's desirable to keep the definition of `trampolineRunners` in `trampoline.js` if possible.

longouyang:
Good idea -- implemented.

stuhlmueller:
Now, calling `require('./transforms/trampoline').trampoline` will change whether `require('./transforms/trampoline').runner` is bound to the `cli` or `web` runner. This is somewhat surprising behavior -- I would expect the call to the trampoline transform to be purely functional. Wouldn't it be simpler and less error-prone to just export both `webRunner` and `cliRunner`?

Also, in `tests/browser/index.html`, the `<script>var forceTrampoline = 'cli';</script>` line is no longer needed.

longouyang:
> Now, calling require('./transforms/trampoline').trampoline will change whether require('./transforms/trampoline').runner is bound to the cli or web runner. This is somewhat surprising behavior -- I would expect the call to the trampoline transform to be purely functional. Wouldn't it be simpler and less error-prone to just export both webRunner and cliRunner?

My thought was that we should expose the *selected* trampoline runner. That way, a library that uses `runTrampoline` works regardless of browser / CLI context (though I guess for this to work smoothly, I should export `runTrampoline` from `src/main.js`, not `getTrampolineRunner`)

> Also, in tests/browser/index.html, the <script>var forceTrampoline = 'cli';</script> line is no longer needed.

Ah, yes.

stuhlmueller:
> My thought was that we should expose the *selected* trampoline runner. That way, a library that uses `runTrampoline` works regardless of browser / CLI context

I agree that it's desirable for libraries to be able to use `runTrampoline` without being aware of the context. However, changing what `require('./transforms/trampoline').runner` refers to whenever we call the transform (1) feels very implicit/magical and (2) makes the assumption that there is a single global context, which isn't necessarily true (maybe we transform some code with one runner and some with another before we run either of them).

I'm not sure what the best solution looks like. I think it would be reasonable to have a `getTrampolineRunner` function that can take the selector string (`cli` or `web`) as argument, and if it doesn't get an argument, it chooses a default based on `util.runningInBrowser()`. (Maybe `trampolineMainWrapper` should also use this sort of default instead of `cli`.)

longouyang:
I refactored so that the trampoline selection is done at run time.

`compile` now returns a function that takes as input a function `runTrampoline` and returns the function that we used to just return immediately.

The commits are showing right above but the Travis tests didn't run, which is a little weird.

stuhlmueller:
Try resolving the conflicts with dev and see whether the tests run then?

longouyang:
Looks like that did the trick, although the process of fixing the conflict was kind of opaque to me (I should take some time to know git better).

Does this look merge-able now?

stuhlmueller:
Looks almost ready to merge! Some minor points:

- Is there a reason why you inlined the `require` calls in `test-transforms.js` and `webppl`? If not, I'd move them to the top of the file for consistency.
- The `newCode.indexOf('runTrampoline') > 0` test in `test-transforms.js` feels quite hacky, but this isn't an important part of the codebase, so I think it's okay to leave it if nothing better comes to mind.
- `trampolineRunners` doesn't feel like it belongs in `util.js`. Why not have it in `trampoline.js`?

longouyang:
> Is there a reason why you inlined the require calls in test-transforms.js and webppl? If not, I'd move them to the top of the file for consistency.

No good reason, I'll move them to the top.

> The newCode.indexOf('runTrampoline') > 0 test in test-transforms.js feels quite hacky, but this isn't an important part of the codebase, so I think it's okay to leave it if nothing better comes to mind.

It's much cleaner to have a separate `runTrampoline` helper that runs this logic, so I'll do this.

> trampolineRunners doesn't feel like it belongs in util.js. Why not have it in trampoline.js?

Two reasons:
1. I viewed trampoline.js as being about the source transformation. Now that we do trampoline selection at run time, it felt like I should put the runners somewhere else.
2. External code needs access to the runners. So putting the runners in util.js seemed natural because util.js methods are inserted into the global namespace. Full disclosure: external code here means both libraries like draw.js and the editor, which now can cancel long running jobs by doing a somewhat magical mutation on the web trampoline runner:

```js
// cancel job
util.trampolineRunners.web.__cancel__ = true;
```

PR #296, opened 2016-01-23
Faster array primitives

longouyang:
As discussed in #174, we can, for many array primitives, eliminate `slice` calls and greatly reduce the number of `concat` calls.

I've optimized all the array primitives except for `sort` and its helper methods, as these are kind of tricky and likely not a bottle neck for many users.


stuhlmueller:
Was the trampoline runner commit 5709ba8d0f4bb5e839965eb9916c20b8ff654d48 accidentally included in this PR?

longouyang:
Yeah, I accidentally based the branch on the wrong commit.

stuhlmueller:
Remove the commit so that we can merge this without causing conflicts for the other PR?

longouyang:
Okay -- I removed the first *two* commits, as they were clutter (5709ba8 and 1d16b16)

Note to self on how to do this:

```sh
git rebase -p --onto UNWANTEDSHA^ UNWANTEDSHA
git push --force --no-verify origin BRANCHNAME
```

(HT http://sethrobertson.github.io/GitFixUm/fixup.html and http://stackoverflow.com/a/8991659/351392 )

stuhlmueller:
Thanks!

Now that the functions are more complex, it would be (even more) helpful to have a test for each of them. This shouldn't take much time -- just place a `.wppl` file per function in `tests/test-data/deterministic/models/`, and the corresponding results file in `tests/test-data/deterministic/expected/`.

null-a:
> It was causing certain Enumerate and IncrementalMH tests to fail with a message about "concat not defined". Switching to append() fixed this.

I think I've fixed this in #297.

longouyang:
I'm starting to write tests but it is a somewhat tedious process.

I want to write a lot of tests for each primitive (perhaps stochastically generated) but the current harness does not facilitate this.

stuhlmueller:
Add very basic tests (maybe only one per function) and create an issue that describes what kind of test harness you would like?

longouyang:
Tests written so far:
- [x] map
- [ ] map2
- [ ] mapN
- [ ] mapIndexed
- [x] mapPairs2
- [x] mapPairsC
- [x] mapPairsNC
- [x] reduce
- [x] mapReduce
- [x] mapReduce1
- [x] all
- [x] any
- [ ] zip
- [x] filter
- [x] find
- [x] remove
- [ ] dropWhile
- [ ] takeWhile
- [ ] span
- [ ] groupBy
- [ ] repeat
- [ ] _merge
- [ ] _sort

longouyang:
Okay, I wrote at least one test for every primitive.

I tried to cover a base case (typically, empty array), small case (array size < 10), and large case (array size 50k)

stuhlmueller:
Looks great! Looking forward to try this in practice.

longouyang:
I want to do some more testing; I'd like to be more confident that the way I wrote these is compatible with addressing for MH.

The MHKernel code is intimidating though.. what's the easiest way for me to modify it so that I can see at every step (1) which address was perturbed and (2) what the resulting trace was?

PR #297, opened 2016-01-25
freevars bug fix

null-a:
This is a fix for the issue Long ran into [here](https://github.com/longouyang/webppl/commit/1d04dec8dc123868fb0b2405bf1498ed707c2e1a).

The bug is fixed in the second commit of this PR.

@dritchie: Does this look OK to you?


dritchie:
Yeah, that looks good to me. I didn't know that estraverse also provided access to the parent node. Was that documented somewhere, and I just missed it?

null-a:
Great, thanks! It's shown in use in the [readme](https://github.com/estools/estraverse), not sure if that's where I got it from though.

PR #299, opened 2016-01-26
Don't include values with probability 0 in aggregation

stuhlmueller:
This is the behavior of enumeration before #287, and some models with nested queries rely on this behavior.

### DIFF
--- a/src/aggregation.js
+++ b/src/aggregation.js
@@ -36,6 +36,9 @@ var Distribution = function() {
 };
 
 Distribution.prototype.add = function(value, score) {
+  if (score === -Infinity) {
+    return;
+  }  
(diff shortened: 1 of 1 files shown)


null-a:
> This is the behavior of enumeration before #287.

Oops!

We should add a test, though I'm not entirely sure whether the current set-up lends itself to this. We could merge this and I could take care of that later if you like?

stuhlmueller:
It would be good more generally to have a few tests for nested enumeration (if we don't have them already). I created an issue. Let's do this as part of a separate PR, and merge this one now if it looks good to you.

null-a:
Sounds/looks good to me.

PR #305, opened 2016-01-29
Defer loading header & packages until run/compile.

### LINKED ISSUE #301: Don't parse packages on page load
Instead, it should be delayed until the first code box is run. Parsing packages on page load slows down the page and (if the webppl script is loaded in the header) it also delays the loading and execution of subsequent scripts.

null-a:
Closes #301.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -2,6 +2,7 @@
 
 'use strict';
 
+var _ = require('underscore');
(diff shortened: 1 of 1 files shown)


stuhlmueller:
Nice, thanks for fixing this so quickly!

Issue #303, opened 2016-01-28
AD transforms lead to strange Math behavior


longouyang:
This code works fine in regular node (result is 1) but weird in webppl (result is 0.9999999999937503):

```js
var x = Math.min(5, 1);
console.log(x)
```

This is because the AD version of `min` is approximate, which is what you want for HMC but not what you want in general. Do we need to apply the AD transforms if we're not using HMC?

PR #307, opened 2016-01-29
update ad.js dependency

iffsid:
This change reflects commit 4facddf in ad.js that undoes the shadowing
of hard functions by their smoothed versions
   - max and min were replaced originally by their approrimate,
     differentiable-everywhere versions
   - this has now been undone and the smoothed versions moved to `smax`
     and `smin`.

Fixes: #303

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "test": "tests"
   },
   "dependencies": {
-    "ad.js": "^1.4.1",
+    "ad.js": "iffsid/ad.js#simple",
     "amdefine": "^1.0.0",
     "ast-types": "^0.8.13",
     "escodegen": "^1.7.0",


stuhlmueller:
Thanks! Why not update the version of `ad.js` on npm?

iffsid:
It's complicated -- the packages (multiple versions) were built for `master`, but this is the `simple` branch. The `master` branch is under a big overhaul with the adnn-related changes, and I didn't know how to make packages that were branch specific, or switch branches beneath the versioning on npm :-)

stuhlmueller:
Wouldn't it work to just check out the branch, run `npm version patch`, then `npm publish`? If not, we can go with the PR as it is.

iffsid:
I believe that would work, but the branch that the package was built under would switch. I didn't want to do that. If I was a proper software dev, I would have made a `release` branch that served as the base for npm package publishing but I think that's a lost cause... :-)

Anyways, I'll revert to using the npm versioned packages again once the changes that have been in the pipeline for a while now get pushed out.

stuhlmueller:
OK, thanks! I hope to see the new version of ad.js soon — it would be great to support HMC for multivariate distributions! :-)

Issue #257, opened 2015-10-20
Add verboseLag to MCMC


mhtess:
Currently, the MCMC /  MH kernel has a `verbose` option. This is useful for gauging how long the program will run. When iterations are relatively cheap, however, the verbose option is too verbose: Printing the iteration number for each iteration slows down the program substantially. I implemented a `verboseLag` option for incrementalMH, which allows the user to specify how often to print the iteration number. A similar option for plain MCMC (and other inference algorithms generally) would be very useful.

@stuhlmueller has relayed to me that this may be somewhat nontrivial, since MCMC is written in such a way that the same kernel is implemented on each iteration (and thus, an action that operates over a time-scale different than each iteration is not immediately possible).

stuhlmueller:
To clarify, all I was trying to say is that to implement this, you'd probably write a new kernel, like the `printCurrIter` kernel we already have, or possibly extending that one. I don't think there's any difficulty here.

null-a:
A couple of thoughts: 

1. Earlier today I realised that rather than writing the current iteration to a new line, it would be nicer to overwrite the current line. This would avoid filling the terminal buffer with junk as we do now. If that solves the same problem as you're having, it might be a bit of a cleaner way to go about it.
2. I've been thinking that long-term the `verbose` option could be implemented via callbacks/hooks. #197.

null-a:
> Printing the iteration number for each iteration slows down the program substantially.

Sorry - missed that bit. Maybe we could overwrite the current line, but do it at most once a second or something like that.

dritchie:
AFAIK, there's no browser-compatible way to do console line overwriting (i.e. node has `process.stdout.write`, but the browser has nothing analogous). If anyone knows of a way around this, I'd be super interested to hear about it.

stuhlmueller:
> AFAIK, there's no browser-compatible way to do console line overwriting (i.e. node has `process.stdout.write`, but the browser has nothing analogous).

Even if there's no browser-compatible way, this is still worth doing just for the command line.

> I've been thinking that long-term the `verbose` option could be implemented via callbacks/hooks. #197.

Agreed.

Issue #197, opened 2015-09-17
Instrument inference algorithms with hooks


null-a:
This was originally raised as part of #64.

> Also, instrument inference algorithms with hooks such that we can pass in functions to collect and compute diagnostics without the diagnostics code being part of the core algorithms.

These hooks might also be used to implement the `--verbose` option of `MCMC`. (Coming in #169.)

null-a:
I have a branch where I've been playing with adding callbacks/hooks to `MCMC`. At the moment you pass an object something like the following to `MCMC`:

````js
{
  init: function() {},
  iteration: function(trace) {},
  finish: function() {}
}
````

The functions get called when you would expect, and it should be pretty easy to implement #257 with this.

My current thinking is that these callbacks will be written in `js` rather than `wppl` as hooking these into inference if they are in CPS appears a bit fiddly.

Another thought is that it might be useful to see if we can make this general enough to allow us to have generic progress indicators (implemented as callbacks) which work across all inference algorithms.

PR #312, opened 2016-02-02
Better verbose mode output for MCMC

null-a:
This changes `MCMC`'s verbose mode to overwrite the current line when printing the current iteration to the terminal under Node.js. (#257.)

This doesn't work in the browser, and the behavior there is not changed by this PR. To make it possible to experiment with the idea of showing a progress bar in the browser (#70), this PR also adds a `callbacks` option to `MCMC`. (#197.)

I think it would be reasonable to commit this as is. The following would then be follow-up work:

* Unify this with the `verboseLag` option. (They differ most in the browser.)
* Add callbacks to `IncrementalMH`, and perhaps other algorithms,

### DIFF
--- a/src/inference/mcmc.js
+++ b/src/inference/mcmc.js
@@ -23,6 +23,7 @@ module.exports = function(env) {
     var callbacks = options.verbose ?
         [makeVMCallbackForPlatform()].concat(options.callbacks) :
         options.callbacks;
+    _.invoke(callbacks, 'setup', numIters(options));
 
     var aggregator = (options.justSample || options.onlyMAP) ?
         new aggregation.MAP(options.justSample) :
@@ -61,6 +62,10 @@ module.exports = function(env) {
     return kernels.tap(function(trace) {
       fn(trace.value, trace.score);
     });
+  }
+
+  function numIters(opts) {
+    return opts.burn + (opts.lag + 1) * opts.samples;
   }
 
   // Callbacks.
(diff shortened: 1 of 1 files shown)


stuhlmueller:
Slick!

Could we also show the acceptance ratio while inference is running? This would allow people to stop long-running chains if the acceptance ratio looks unexpected.

I'm really looking forward to seeing progress bars in the browser. This will make long-running computations a lot more palatable.

null-a:
> Could we also show the acceptance ratio while inference is running?

Sure. I've added it to both node and the browser. If it's too much in the browser, it's no trouble to put that back how it was.

> I'm really looking forward to seeing progress bars in the browser.

Me too. As a proof of concept, I made an example of using these callbacks to drive a [progress bar in node](https://github.com/null-a/webppl-node-progress-bar).

stuhlmueller:
> As a proof of concept, I made an example of using these callbacks to drive a [progress bar in node](https://github.com/null-a/webppl-node-progress-bar).

Cool! (It's flickering for me, so probably needs some throttling.)

It will be interesting to think about how to do this for enumeration as well. For unbounded enumeration we'll have to estimate the total number of executions based on what we've seen so far.

null-a:
Thanks!

> It's flickering for me, so probably needs some throttling.

The JS progress bar library I used does appear to use throttling. I tried adding more throttling and nothing seemed to change. I also tried the very latest version from the repo (as they have unreleased changes) but that didn't work at all. I'm going to leave this as it is for now, I'll revisit it if anyone starts using it seriously and runs into problems.

PR #315, opened 2016-02-09
better err for factor(NaN)

jsalvatier:
Previously you just got `true == false`, which is not very informative.

### DIFF
--- a/src/header.js
+++ b/src/header.js
@@ -77,7 +77,7 @@ module.exports = function(env) {
   };
 
   env.factor = function(s, k, a, score) {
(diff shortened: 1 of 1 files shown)


stuhlmueller:
Thanks for the PR! This error message should definitely be more informative.

The PR failed to pass the linter, which prefers single over double-quoted strings. Can you make that change? (In the future, if you `grunt` locally before you make a PR, it will run both the linter and the test suite. Then, if linting fails, it can often be fixed automatically using `grunt fixjsstyle`. )

Issue #293, opened 2016-01-20
Support Enumerate inside HMC


null-a:
Previous discussion:

* Original [HMC issue](https://github.com/probmods/webppl/issues/81#issuecomment-153118693)
* #287

stuhlmueller:
To do this, it could be helpful to generalize `transformERP` so that it transforms all files in the repository that end in `.ad.js`.

null-a:
I think I have this working for the case where tapes enter the enumerated model as ERP parameters. See [this test](https://github.com/null-a/webppl/blob/3b0eead6c054f0feaf22fcdfce704cdd527a02d4/tests/test-data/stochastic/models/nestedEnum2.wppl) (based on Andreas' example from #287) for an example.

(See my [hmc-over-enum](https://github.com/null-a/webppl/commits/hmc-over-enum) branch for the rest of the code.)

It's of course possible to write down models where tapes enter in some other way, but I couldn't come up with any examples that seemed useful in practice. Does what I have cover the cases we're interested in?

So far I've adified appropriate bits of code by hand to get a feel for what needs transforming. It's not clear to me *exactly* how to best generalize `transformERP`, but I think we could happily merge this without such a generalized transform to begin with.

stuhlmueller:
This may already be covered by what you have, but two important cases to get right are sequential and nested enumerations, as in these examples:

Sequential:

    var model = function() {      
      var x = uniform(0, 1);
      
      var firstMarginal = Enumerate(function() {
        var y = flip(x);
        factor(!y ? 0 : -2);
        return y;
      });
    
      var secondMarginal = Enumerate(function(){
        var z = sample(firstMarginal);
        factor(z ? 1 : -1);
        return z;
      });
      
      var w = sample(secondMarginal);      
      return x * w;
    };

Nested:

    var model = function() {    
      var x = uniform(0, 1);
      
      var outerMarginal = Enumerate(function() {
            
        var innerMarginal = Enumerate(function(){
          var y = flip(x);
          factor(y ? 1 : -1);
          return y;
        });
        
        var z = sample(innerMarginal);
        factor(!z ? 0 : -2);
        return z;
      });
    
      var w = sample(outerMarginal);
      return x * w;
    };

ngoodman:
We do have a lot of cases where params flow into nested Enumeration not directly as an ERP param. Eg. within the enum there might be a `factor(param1*x + param2*y)` kind of thing.

[...]

PR #334, opened 2016-02-12
Split aggregators out into separate files.

null-a:
This change is motivated by #293 where it will be useful to have `Distribution` (which doesn't need `ad`) separated out from `Histogram` (which does).

Even without that motivation, I think this PR is worth merging. Since adding `Distribution` to `aggregators.js` it's felt like there is too much going on in that file.

I've split this out into a separate PR for ease of review.


stuhlmueller:
Looks good. Thanks for splitting this out into a separate PR!

PR #335, opened 2016-02-12
Remove superfluous methods.

null-a:
These came along with the new trampoline runner, but I'm not sure they do anything useful.

`compile` doesn't appear to take a `trampolineRunner`option, and `run`'s second arg is a continuation not and options object.

### DIFF
--- a/tests/browser/index.html
+++ b/tests/browser/index.html
@@ -4,12 +4,12 @@
     <meta charset="utf-8">
     <title>WebPPL</title>
     <script src="../../compiled/webppl.min.js"></script>
-    <link rel="stylesheet" href="http://code.jquery.com/qunit/qunit-1.19.0.css">
+    <link rel="stylesheet" href="http://code.jquery.com/qunit/qunit-1.21.0.css">
   </head>
   <body>
     <div id="qunit"></div>
     <div id="qunit-fixture"></div>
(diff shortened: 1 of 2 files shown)


longouyang:
The change to `compile` looks right. I think we need to use the cli trampoline runner, though, so I think `run` should be:

```js
var topK = function(s,x) { 
  console.log(x);
}

var run = function(code) {
  return webppl.run(code, topK, {runner: 'cli'});
}
```

(added continuation argument and switched from `trampolineRunner` to `runner`)

null-a:
> I think we need to use the cli trampoline runner

Why? These tests are passing right now. Doesn't it make sense to use the `web` runner in browser tests?

stuhlmueller:
> Why? These tests are passing right now. Doesn't it make sense to use the web runner in browser tests?

The worry is that longer-running tests will pass by accident if the web runner is used. The web runner makes use of `setTimeout`, which detaches part of the test computation from the computation that starts the test. Failures towards the end of the test might go unnoticed, since (by the time they happen) the test results have already been evaluated (in the thread that started the test) and the test runner has moved on to the next test, or finished.

null-a:
Ah, gotcha, good point, thanks. QUnit can test [asynchornous](http://api.qunitjs.com/async/) things. So we should use that to exercise the `web` runner, right?

Issue #313, opened 2016-02-08
When webppl is loaded in the browser, console.log its version


stuhlmueller:
Log the git tag (which corresponds to the npm version) if we're at a tag, otherwise the git commit hash.

stuhlmueller:
It would be useful to have this for each package as well.

null-a:
The way this is implemented at the moment [doesn't work](https://github.com/probmods/webppl/pull/341#issuecomment-185926219) with everyone's git workflow. The fix is to pull the version from `package.json` instead. This could also work for #342.

Issue #324, opened 2016-02-10
Document header functions


null-a:
We should probably do #323 first.

null-a:
@hawkrobe added some [annotations to the header](https://github.com/probmods/webppl/commit/03ae8889546f2383f3636f8c769605e7c8afcf44).

null-a:
The work @hawkrobe [started on this](https://github.com/probmods/webppl/commit/03ae8889546f2383f3636f8c769605e7c8afcf44) describes header functions by comparison to similar functions in `underscore`. This approach is a little surprising to me, but that might just be because I have a different perspective on webppl over here. (I suppose if I *did* compare header functions to something else in the docs, I'd be more likely to choose their JS counterparts.) 

So, before I start working on this myself I just want to check: do you want me continue in this vein, describing functions where possible by their differences to `underscore` versions?

stuhlmueller:
I'd describe the functions without comparison to anything, and maybe name the JS counterpart where appropriate.

ngoodman:
I'm going to call this done, since additions or refinements can happen organically now.

PR #341, opened 2016-02-15
Show webppl version in the browser

null-a:
This includes the output of [`git describe`](https://git-scm.com/docs/git-describe) in the browser when webppl is loaded. (Addressing the first part of #313.)

I've updated `webppl --version` to also include this information.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -68,12 +68,10 @@ module.exports = function(grunt) {
 
   function browserifyArgs(args) {
     var pkgArg = '';
-    if (args.length > 0) {
-      var requires = _.chain(_.toArray(args))
-          .map(function(name) { return ['--require', name]; })
-          .flatten().value();
-      pkgArg = ' -t [' + ['./src/bundle.js'].concat(requires).join(' ') + ']';
-    }
+    var requires = _.chain(_.toArray(args))
+        .map(function(name) { return ['--require', name]; })
+        .flatten().value();
+    pkgArg = ' -t [' + ['./src/bundle.js'].concat(requires).join(' ') + ']';
     return pkgArg + ' -g brfs src/browser.js -o compiled/webppl.js';
   }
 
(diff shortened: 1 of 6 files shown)


null-a:
You can see an example of the output from this in this travis log:

https://travis-ci.org/probmods/webppl/jobs/109358639

stuhlmueller:
Nice! I didn't know about `git describe`, but it's exactly what I had in mind.

Why the switch from `"git-rev-2": "^0.1.0"` to `"git-rev-2": "yanatan16/git-rev#bfc97b3d"`?

null-a:
> Why the switch

Because the `git-rev-2` package hasn't been updated since the `git describe` functionality was added.

longouyang:
`git describe` reports different things depending on whether you've fetched the latest tags (which I often haven't). To make this more robust, can we hard code the version number in the source code and only use the hash part of the output from `describe`?

stuhlmueller:
Note that the version number is also in `package.json`.

null-a:
Interesting. So you can end up with the code for say v0.7, but without the git tag? Out of interest, how does that happen?

> Note that the version number is also in package.json.

Indeed, this is what I had in mind for #324.

longouyang:
If I understand correctly, `git pull` by default only downloads the commits, not the tags.

To get the tags you need to do `git fetch --tags` or [add this to the default pull behavior in .gitconfig](http://stackoverflow.com/a/16678319/351392)

null-a:
Ah, ok, thanks. My workflow involves `git fetch` and `git merge` rather than `git pull` which probably explains it.

I don't use `--tags` or the config you linked, but I still get the tags. From the man page:

> By default, any tag that points into the histories being fetched is also fetched;

Maybe that's why.

As for fixing this, I'll consider making the change discussed here to be part of #313 which is still open and has been updated accordingly.

PR #350, opened 2016-02-22
Bump up HMC over Enumerate test tolerances.

null-a:
`HMC - nestedEnum5` [failed](https://travis-ci.org/probmods/webppl/jobs/110436667) on dev after we merged #344. This gives us a little more breathing room.


stuhlmueller:
It looks like the test tolerances are still too low (or there is a bug in the algorithm):
https://travis-ci.org/probmods/webppl/jobs/111013425

Issue #292, opened 2016-01-20
Readable runtime error messages via source maps


chrisranderson:
I've talked with Andreas a bit about figuring out how to tackle this, and after looking through the code and the old stack trace branch, here's my best shot for how to get it done:

- have sweet.js generate source maps [here](https://github.com/probmods/webppl/blob/4b8c735ee982a04815230d30e93c1f645fdc71a3/src/main.js#L54).
- when [the code is evaluated](https://github.com/probmods/webppl/blob/4b8c735ee982a04815230d30e93c1f645fdc71a3/src/main.js#L179), catch errors and translate the JS stack traces to WebPPL stack traces using the source map.

**Questions**
- What do I do with the code that you compile for every program? It looks like you combine the user's source with something else (header.wppl?).
- I haven't thought much about converting the error message itself. Is it okay if I work on that after stack traces are converted?
- Should I use something like [stacktrace.js](https://github.com/stacktracejs/stacktrace.js) or some other library for parsing the JS stack trace?

Hopefully I'm on the right track.

stuhlmueller:
To clarify, full stack traces are probably too difficult for now, since trampolining eliminates the stack in the compiled code. Our modest goal should be to locate errors in the original webppl code when runtime errors happen. In other words, we're only looking at the top-most element of the stack trace.

I'll let Paul (@null-a) reply to your approach and questions (tomorrow), since he is the person most familiar with `main.js`.

null-a:
>  here's my best shot for how to get it done

I think you're pretty close! Here's a sketch of what I think we need to do:

**1. Get an annotated AST from sweet.js**

What we get back from `sweet.compile` is an AST, and we need it to be annotated with source code locations. We'll need line number, column number and filename. It looks to me as though the AST is already annotated with line and column information, but not the filename.

The parser which sweet.js uses does have an option to include the filename, but there's no way to pass options from `sweet.compile` through to the parser. I can think of two ways to fix this:

* Add the filename information to the AST in a separate post-processing step. (This is [how the sweet.js parser does it](https://github.com/mozilla/sweet.js/blob/da661ba1eb9b22049f7e2f0eed12d79b4de04780/src/parser.js#L5759) anyway.)
* Patch sweet.js to pass the options through. (This isn't as bad as it sounds as we already maintain a fork.)

Of those, I think the first is definitely the way to go.

[...]

PR #357, opened 2016-02-24
Make sources available within compile.

null-a:
The key change here is to `parseAll`, where instead of returning just the ast, we augment the 'bundle' we took as input with the ast. The makes all original source code available in `compile` which we appear to need for #292.

The rest is mostly just renaming `extras` to `bundles` everywhere, as the thing returned from `parsePackageCode` is now an array of 'bundles', and not this awkward object which I often called extras. (Which looked like `{ ast: '', macros: [] }`.)

I've tested compiling for the browser with packages, and it still works.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -24,16 +24,16 @@ var load = _.once(function() {
     if (pkg.js) { global[pkg.js.identifier] = pkg.js.path; }
     pkg.headers.forEach(webppl.requireHeaderWrapper);
   });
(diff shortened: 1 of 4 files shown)


stuhlmueller:
Thanks for the helpful description. Looks good.

At some point we might want to think about if `main.js` can be refactored in some way to make it easier to read, maybe by splitting it up into multiple files. There isn't that much going on, but I still find it difficult to wrap my head around.

PR #359, opened 2016-02-26
Document browser usage

longouyang:


stuhlmueller:
People might be confused about what "compiled file" refers to. The previous sentence mentions the `geometric.js` compiled file, but here it seems that "compiled file" refers to the browserified webppl code. Maybe the additions in this PR should be combined with the existing description under [Compiling for browser](http://docs.webppl.org/en/master/development/workflow.html#compiling-for-browser)?

longouyang:
What if I change "compiled file" to "browserified package" and have that be a link that points to the Compiling for browser section?

stuhlmueller:
Sounds good.

PR #363, opened 2016-03-01
Rejection - default for numSamples arg

null-a:
Before this, `Rejection` would enter an infinite loop if `numSamples` was omitted.


stuhlmueller:
Wouldn't it be more appropriate to throw an error if `numSamples <= 0`?

null-a:
Sure, done.

PR #364, opened 2016-03-01
Use current convention for erp and cont. args

null-a:


null-a:
Test failure was:

> HMC - mixed1Factor
> Message: Expected hist: {"true":0.130435,"false":0.869565}, actual: {"true":0.421,"false":0.579}
> Error: false == true

I've just run that test locally 10K times and it only failed once! I've also run it using the random seed that caused the failure but with more samples, and the posterior does look like it converges. Shall we leave it as it is and see how we go?

stuhlmueller:
A potential confusion is that we have a module (imported in `header.js`) that's called `erp`, and so naming other variables `erp` shadows that module variable. This doesn't matter to the functionality of the code at the moment, but it makes it a little harder to read, since one has to find the next function scope and look at the arguments to determine whether `erp` refers to an argument or the module. One possible solution would be to rename `erp.ad.js` to `erps.ad.js`.

> I've just run that test locally 10K times and it only failed once! I've also run it using the random seed that caused the failure but with more samples, and the posterior does look like it converges. Shall we leave it as it is and see how we go?

Ok.

null-a:
It's a reasonable point. It's kind of awkward that `erp` is assigned at the top of the file but then used only once at almost the very bottom. Perhaps we could consider whether we could split out the definition of the coroutine bits into a separate file, and then they could be imported/re-exported in a similar way to everything else, and we'd side-step this shadowing problem in the process.

Having said that, the original renaming only seems worth doing if it's relatively quick and painless. (This is just something I split off form a bigger change to try and make that easier to manage.) I'd be happy to kill this PR and submit another one with just the changes to `enumerate.ad.js` if you are.

stuhlmueller:
> Perhaps we could consider whether we could split out the definition of the coroutine bits into a separate file, and then they could be imported/re-exported in a similar way to everything else, and we'd side-step this shadowing problem in the process.

Is the idea that a new `coroutine.js` file would be structured like the inference modules (exporting a function that takes `env` as an argument), and that we'd get rid of the `addExports` specific to the inference interface (`factor`, `sample`, ...) and instead deal with it like we deal with other header modules? If that's possible without any complications, it sounds like a desirable simplification.

I'm going to merge this PR as is, since it's an improvement overall and the variable shadowing issue is minor, but consider trying that simplification at some point.

null-a:
I didn't think it through in detail, but's that the general idea. I'll create an issue to remind me to try it.

Thanks for the review & merge!

Issue #316, opened 2016-02-09
IncrementalMH behavior when sampled from ERP depends on random choices


null-a:
I noticed that in the following model, the marginal given by `IncrementalMH` differs from those given by other algorithms.

````js
var mostlyTails = bernoulliERP.withParameters([0.2]);
var mostlyHeads = bernoulliERP.withParameters([0.8]);

var model = function() {
  var x = flip();
  sample(x ? mostlyHeads : mostlyTails, []);
};

//Enumerate(model);
// ERP:
//     false : 0.5
//     true : 0.5

//MCMC(model, { samples: 10000 });
// ERP:
//     false : 0.5045000000000001
//     true : 0.4955


IncrementalMH(model, 10000);
// ERP:
//     false : 0.7888788878887889
//     true : 0.2111211121112111
// ERP:
//     true : 0.788978897889789
//     false : 0.211021102110211
````

As a guess, the ERP which is passed to `sample` on the first execution is used for all executions. Is this a bug, or am I missing something?

I came up with this while trying to understand why I couldn't get my nested Enumeration tests (#293) passing for `IncrementalMH`. I think this is why.

dritchie:
I'll dig into this in a little bit when I get some time. I seem to 
recall having some problems with IncrementalMH not playing well with 
direct calls to 'sample' before, but I thought I'd resolved that issue. 
We'll see.

Issue #246, opened 2015-10-13
Interface to sample/ERPs


null-a:
Thinking about some of the open issues relating to `ERP`s (#119, #231) has made me think about the choice to have the interface to `sample` be:

````js
sample(erp, params);
````

Rather than:

````js
sample(erp(params));
````

Having thought about it a little, it seems to me that switching to the second of these might have some benefits.

More specifically, the suggestion would be to switch to having a function represent a family of distributions (I'll call these `Dist`s here to disambiguate them from `ERP`s) which when called returns a particular distribution from the family. It would look like this:

````js
var coin = bernoulliDist(0.5);
coin.sample();
coin.score(true);
sample(coin);
````

Switching to this approach might help in the following ways:

* It might help to simplify the inference code which deals with importance distributions a little. For example at present `MHKernel` is concerned with [constructing arrays of parameters](https://github.com/probmods/webppl/blob/7e4ff8f3544c323c6c389454b139323e28fbb5e0/src/inference/mhkernel.js#L62) to hand off to drift kernels. This is already a little messy and when I spent a little time thinking about #231 it seemed its implementation would make things worse. The hope is that by bundling the `erp` and `params` together, working with `Dist`s would allow us to more easily factor this massaging of parameters out of the inference code.

* It would allow us to remove the current notion of [parameterized](https://github.com/probmods/webppl/blob/7e4ff8f3544c323c6c389454b139323e28fbb5e0/src/erp.js#L69) and unparameterized `ERP`s.

This, for example:

````js
var FairCoin = bernoulliERP.withParameters([0.5]);

var model = function() {
  var x = sample(FairCoin);
  var y = sample(withImportanceDist(FairCoin, bernoulliERP.withParameters([x ? .5 : 1])));
  return [x, y];
};
````

Would become something like:

````js
var FairCoin = bernoulliDist(0.5);

var model = function() {
  var x = sample(FairCoin);
  var y = sample(withImportanceDist(FairCoin, bernoulliDist(x ? .5 : 1)));
  return [x, y];
};

````

* We already have a few functions in `header.wppl` which behave rather like `Dist`s. For example, you might expect `deltaERP` to behave like this:

````js
sample(deltaERP, [0]) => 0
deltaERP.score([0], 0) => 0
deltaERP.score([0], 1) => -Infinity
````

However, `deltaERP` is actually a function which returns an "unparameterized" `ERP`, so it actually behaves like this:

````js
sample(deltaERP(0), []) => 0
[...]

PR #365, opened 2016-03-01
Fixes #311 and #316

### LINKED ISSUE #311: IncrementalMH doesn't work with globalStore
Running this model (with latest dev version b5eaaf9):

```js
var addToGlobalStore = function(x) {
  globalStore.foo += x;
}

var model = function() {
  globalStore.foo = 0;
  var a = gaussian(0,1);
  addToGlobalStore(a);

  var b = beta(1,1);
  addToGlobalStore(b);

  return [a,b]
}

IncrementalMH(model, 100)
```

Gives this error:

```
/Users/long/webppl/src/inference/incrementalmh.js:258
      if (s1[prop] !== s2[prop])
            ^

TypeError: Cannot read property 'foo' of undefined
    at storesEqual (/Users/long/webppl/src/inference/incrementalmh.js:258:13)
    at FunctionNode.registerInputChanges (/Users/long/webppl/src/inference/incrementalmh.js:457:10)
    at IncrementalMH.cachelookup (/Users/long/webppl/src/inference/incrementalmh.js:977:19)
    at IncrementalMH.incrementalize (/Users/long/webppl/src/inference/incrementalmh.js:950:19)
    at env.incrementalize (/Users/long/webppl/src/header.js:107:26)
    at /Users/long/webppl/tmp.js:1305:44
    at runTrampoline (/Users/long/webppl/tmp.js:22:17)
    at /Users/long/webppl/tmp.js:27:9
    at Object.<anonymous> (/Users/long/webppl/tmp.js:1335:1)
    at Module._compile (module.js:435:26)
```

The problematic function is `storesEqual` in *`incrementalmh.js`*:

```js
  function storesEqual(s1, s2) {
    var prop;
    for (prop in s1) {
      if (s1[prop] !== s2[prop])
        return false;
    }
    for (prop in s2) {
      if (s1[prop] !== s2[prop])
        return false;
    }
    return true;
  }
```

The issue is that `s1` is undefined. I think adding this check at the very beginning of the function would help:

```js
if (s1 === undefined) {
  return (s2 === undefined)
}
```

But I don't really know how incrementalization works, so I'm not sure if this fix is appropriate.

dritchie:


dritchie:
This fixes issue #311

null-a:
> // TODO: Have ERPs provide 'equal' method to check whether they represent
> //    the same distribution, even though they may refer to different objects.

It's my intention to add this as part of #246. I'll take care of using it here too.

PR #371, opened 2016-03-07
Add 'grunt watch'.

null-a:
This adds a new grunt task that automatically runs `scipts/adify` when `.ad.fs` files are changed.

It's unfortunate that we now have `grunt watch` and `grunt watchify`, but I think it's best if I try and improve that as a separate PR.


stuhlmueller:
Looks good. I agree that we should clean up the confusing naming (as a separate PR).

Issue #342, opened 2016-02-15
Make --version work when webppl is installed globally


null-a:
In such cases the directory in which `webppl` sits won't be a git repository, so `webppl --version` won't do anything useful.

PR #372, opened 2016-03-07
Rework version/git info.

### LINKED ISSUE #313: When webppl is loaded in the browser, console.log its version

null-a:
Following [the discussion](https://github.com/probmods/webppl/pull/341#issuecomment-185926219) we had about git workflows, the webppl version number shown by `webppl --version` and in the browser is now fetched from `package.json`.

The fancy features we got from `git describe` (number of commits behind most recent tag, not showing the commit hash when at a tag) have been dropped.

(We could bring this back for people who *do* fetch tags by checking if the most recent tag matches the version number in `package.json`. It wasn't clear that this was useful enough to incur the extra complexity though, so I've not done that here.)

Version info is also shown for packages in the browser. Example:

````
loaded webppl-fs [v0.0.1-f1eac96]
loaded webppl-linked-list [v0.0.1-6e68233]
loaded webppl [v0.6.1-03a4d4e]
````

Also, we now handle the case where webppl/packages are not git repos, as well as the case where the `git` binary is not available. (By showing only the version number and not the git commit hash.)

None of the existing `git-rev-*` npm packages provided the functionality I needed to implement this so I've added my own simple wrapper around `git branch` etc. If it offends anyone not using a package here I can either try and get an existing package updated or create our own, though my own opinion is that this is fine for now as it is.

Closes #313, #342.


stuhlmueller:
> (We could bring this back for people who *do* fetch tags by checking if the most recent tag matches the version number in `package.json`. It wasn't clear that this was useful enough to incur the extra complexity though, so I've not done that here.)

I agree, I don't think this would warrant the increase in complexity. 

Getting rid of the `git-rev-*` dependence also looks like the right call.

Issue #286, opened 2016-01-12
Make HMC work for multivariate distributions


null-a:
Switching to the latest version of ad.js (once available) will give us support for multivariate distributions in HMC.

The following changes also need to be made at the same time:

* Move the version of untapify in aggregation.js into the ad package.
* Don't hardcode the path to the ad.js macros file.

null-a:
The current plan is to implement this by switching to @dritchie's ad implementation.

null-a:
Closed by #503.

PR #381, opened 2016-03-18
Use adnn for AD

null-a:
This changes the ad library used for HMC from ad.js to adnn.

This is a first step towards #286, #259.

For the kinds of models I've used while implementing & testing HMC I think adnn is a little slower than ad.js.

For example, I ran a few iterations of the model MHT has been using to experiment with HMC under both versions. The avg. run-time (over 5 runs) for each was:

````
ad.js |  8.8s
adnn  |  9.5s
````

The gap would have been bigger than this, but Daniel and I have spent some time staring at the output of the profiler and found some improvements.

(These timings were taken using the most recent commit on master. We won't pull the last tweak we made into WebPPL until there's another npm release of adnn.)

I think we're at a point where this is close enough to merge.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "test": "tests"
   },
   "dependencies": {
-    "ad.js": "iffsid/ad.js#simple",
+    "adnn": "^1.0.0",
     "amdefine": "^1.0.0",
     "ast-types": "^0.8.13",
     "escodegen": "^1.7.0",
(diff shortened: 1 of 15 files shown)


null-a:
I should have mentioned that I've tested this in the browser (by running HMC over a super simple model) and it worked OK.

stuhlmueller:
Looks good. There is an issue (#382) related to the code touched by this commit that we encountered while working on AgentModels, but it's not introduced by the changes here, so should probably be addressed as part of a separate PR.

PR #384, opened 2016-03-24
Make Histogram not break prototype chain when it adds values.

dritchie:
I noticed that when a webppl inference thunk has an object return value, `Histogram` will lose that object's prototype when stripping the object of any AD tapes. This changes ensures that prototypes are preserved.

### DIFF
--- a/src/ad.js
+++ b/src/ad.js
@@ -9,11 +9,11 @@ var valueRec = function(x) {
   } else if (_.isArray(x)) {
     return _.map(x, valueRec);
   } else if (_.isObject(x) && !_.isFunction(x)) {
-  	// Ensure prototype chain is preserved
-  	var proto = Object.getPrototypeOf(x);
-  	var y = _.mapObject(x, valueRec);
-  	Object.setPrototypeOf(y, proto);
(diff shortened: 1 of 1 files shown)


null-a:
Good catch! A couple of thoughts:

* It would be nice to have a test for this somewhere. Coincidentally, I ran into a similar problem yesterday on the daipp branch. There, the problem was with `valueRec` messing with Tensors. To fix that I think I need this fix and also proper handling of `Float64Array`s in `valueRec`. The reason I mention this is that once we have inference tests in daipp, then we will have tests exercising this code, so we probably don't need to worry about tests here. (So, just to be clear, I think it's fine to merge this as is.)
* The CI failure was caused by the linter. I realize you likely know this already, but I think you'll be able to fix the errors with `grunt fixjsstyle`.

dritchie:
Derp--fixed that style issue.

dritchie:
Hmm, it looks like CI is still failing b/c at least one of its tests is running in an environment that doesn't support `setPrototypeOf` (Safari doesn't yet support it, it seems). I'll see about working around this.

PR #388, opened 2016-04-01
Make arguments work in all contexts.

### LINKED ISSUE #31: Make 'arguments' work in all contexts
Currently, the varargs transform happens before trampolining (so closures introduced there are handled correctly), but after cps (so if cps introduces a closure boundary that separates the occurrence of `arguments` from its corresponding function top level, `arguments` won't work correctly). See example in e22b7e25d5f654f785eed124b4923b53bae4e00f.

null-a:
This is intended to fix #31.

As suggested in the ticket, I've moved the `varargs` transform before `cps`. I side stepped the stack overflow which results from doing this by switching the strategy used to find a node's enclosing function. Specifically, we now maintain a stack of function nodes as we traverse, rather than annotate the AST. With that, the failing test passes.

One thing I'm not sure about is whether I've done the right thing with the tests. I've changed `transformAstVarargs` so that `varargs` runs before `cps` in the var args tests, and by doing it this way I don't think I've changed any other tests. (e.g. by including the `varargs` transform in tests that previously didn't include it.) Does this seems right, or does changing the order of the transforms require us to make more substantial changes to the tests?

I'm not entirely sure whether the existing `seenByVarargs` check is necessary any more. I can't quite fathom why it was needed originally, so for now I've done the conservative thing and left it as is.

(This probably isn't top priority, but I've had it laying around for a while and figured I should clean it up and make a PR.)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -147,10 +147,10 @@ function compile(code, options) {
   var transforms = options.transforms || [
     thunkify,
     naming,
+    varargs,
     cps,
     store,
     optimize,
-    varargs,
     trampoline
   ];
 
--- a/src/transforms/varargs.js
+++ b/src/transforms/varargs.js
@@ -37,11 +37,6 @@ function getEnclosingFunctionNode() {
 
 function varargs(node) {
 
-  if (node.seenByVarargs) {
-    return node;
-  }
-  node.seenByVarargs = true;
-
   switch (node.type) {
 
     // assign 'arguments' as first statement in body, rename to make it
@@ -76,11 +71,10 @@ function varargsMain(node) {
       node, {
         enter: function(node) {
           pushFn(node);
+          return varargs(node);
         },
         leave: function(node) {
-          var ret = varargs(node);
           popFn(node);
-          return ret;
         }
       });
   return node;
--- a/src/transforms/varargs.js
+++ b/src/transforms/varargs.js
@@ -13,23 +13,26 @@ function makeArgumentsIdentifier() {
   return '_arguments' + argumentsIdCounter;
 }
 
-function findEnclosingFunctionNode(node) {
-  var ancestor = node;
-  while (ancestor !== undefined) {
-    if (ancestor.type === 'FunctionExpression') {
-      break;
-    }
-    ancestor = ancestor.parentNode;
+var fnStack = [];
+
+function pushFn(node) {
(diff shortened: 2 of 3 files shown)


stuhlmueller:
> As suggested in the ticket, I've moved the `varargs` transform before `cps`. I side stepped the stack overflow which results from doing this by switching the strategy used to find a node's enclosing function. 

Out of curiosity, why did moving `varargs` before `cps` result in a stack overflow again?

> Does this seems right, or does changing the order of the transforms require us to make more substantial changes to the tests?

One principle we could follow with the transform tests is to test each transform in the context that it is actually used, i.e. applied to code that is pre-transformed using all transforms that precede the tested transform in the webppl compilation order. In this case, we'd want to include the `varargs` transform among the transforms that are applied in the `cps`, `store`, and `optimize` tests, and we'd also want to remove these three transforms from the `varargs` test. (We didn't quite follow this principle for the `varargs` test before, and your solution does look like the correct minimal edit, so if you'd rather just merge the PR as it is, I'm happy with that as well.)

> I'm not entirely sure whether the existing `seenByVarargs` check is necessary any more. I can't quite fathom why it was needed originally, so for now I've done the conservative thing and left it as is.

It doesn't look like it's doing anything (anymore); feel free to remove it.

null-a:
> Out of curiosity, why did moving varargs before cps result in a stack overflow again?

I don't know. The stack overflow comes from a recursive call to esmangle's `deepCopyInternal` within the `optimize` transform. How it is that the setting `parentNode` and the extra transforms are interacting to cause this is a mystery to me.

> One principle we could follow with the transform tests is to test each transform in the context that it is actually used in

I think this is tricky to do in general because for the `varargs` tests we need to run `cps` and `store` after the `varargs` transform. (Because the `varargs` transform slices the original JS `arguments` in a way that assumes they includes the store, continuation and address.) You can imagine doing something about this of course, but I'm not sure it buys us much?

> we'd want to include the varargs transform among the transforms that are applied in the cps, store, and optimize tests

I've done this bit though.

> feel free to remove it.

Done.

stuhlmueller:
> I think this is tricky to do in general because for the `varargs` tests we need to run `cps` and `store` after the `varargs` transform. (Because the `varargs` transform slices the original JS `arguments` in a way that assumes they includes the store, continuation and address.)

Good point.

Issue #268, opened 2015-11-11
Make all arguments to ERPs named parameters


stuhlmueller:
How would people feel about replacing calls such as `gaussian(0, 1)` and `cauchy(2, 1.5)` with `gaussian({mean: 0, std: 1})` and `cauchy({location: 2, scale: 1.5})`?

stuhlmueller:
This is really two proposals in one. First, one could change how the wrapper functions in `header.wppl` work (as illustrated above). Second, one could change the ERP interface itself, e.g. turning `sample(gaussianERP, [0, 1])` into `sample(gaussianERP, {mean: 0, std: 1})`. We don't necessarily have to do both of these.

ngoodman:
i think this is a great idea -- it'd hard to keep track of the params as arrays. it seems most reasonable to change the ERP interface, and then to have the wrapper functions work both ways if possible (because sometimes you just want to write `flip(0.5)` or `gaussian(0,1)`....).

ngoodman:
Note this is related to #246.

null-a:
Closed by #393.

PR #393, opened 2016-04-11
New ERP Interface

null-a:
This implements the following changes to ERPs:

1. ERP are now represented by instances of object types rather than by an object (containing `sample`, `score`, etc.) and an array of parameters. (#246)
2. ERPs take named parameters. (#268)

In practice, this means:

````js
sample(gaussianERP, [0, 1]) // is now =>
sample(gaussianERP({mu: 0, sigma: 1}))

factor(gaussianERP.score([0, 1], x)) // is now =>
factor(gaussianERP({mu: 0, sigma: 1}).score(x))

gaussian(0, 1) // still works
gaussian({mu: 0, sigma: 1}) // also works. similarly for all multi-argument erp helpers
````

The notion of (un/)parameterized ERP has been replaced by object types and their instances, so:

````js
var FairCoin = bernoulliERP.withParameters([0.5]);
var BentCoin = bernoulliERP.withParameters([0.2]);
sample(withImportanceDist(FairCoin, BentCoin))

// is now =>
var FairCoin = bernoulliERP({p: 0.5});
var BentCoin = bernoulliERP({p: 0.2});
sample(withImportanceDist(FairCoin, BentCoin))
````

Bonus features:

````js
entropy(bernoulliERP({p: 0.1})) // => 0.3250829733914482
display(bernoulliERP({p: 0.1})) // prints "bernoulli({ p: 0.1 })" (only in node)
util.serialize(bernoulliERP({p: 0.1})) // => '{"probs":[0.1,0.9],"support":[true,false]}'
````

A few things have changed on the backend as a result of all this:

1. The logic for testing for erp equality has been moved out of `IncrementalMH` and into `erp.isEqual`. It would be a good idea to ask @dritchie to review the changes to `IncrementalMH` if we decide we want to merge this PR.
2. Responsibility for un-lifting ad nodes has moved from the inference algorithms to the ERP.
3. `multiplexERP` has been removed.
4. `categoricalERP` (in `header.wppl`) was previously a function which took `ps` and `vs` args and returned an "unparameterized" ERP. It's now takes a `params` object an returns an instance of a categorical ERP. Similarly for `deltaERP`.

### CODE REVIEW

**Comment by null-a** [comment]
I've tested these changes against a couple of simple models and several inference algorithms. I ran inference ten times for each algo/model combo and took the average of the runtimes.

````
|----------------+-------+-------------+------------|
| algo           | model | before (ms) | after (ms) |
|----------------+-------+-------------+------------|
| enumerate      |     1 |        6365 |       6075 |
| hmc            |     2 |        6904 |       6647 |
| incremental mh |     1 |        6521 |       6571 |
| incremental mh |     2 |        6637 |       7086 |
| mh             |     1 |        8331 |       7467 |
| mh             |     2 |        8046 |       7847 |
| rejection      |     1 |        8513 |       6974 |
| rejection      |     2 |        6970 |       6778 |
| smc            |     1 |        7685 |       6223 |
| smc            |     2 |        8242 |       7861 |
|----------------+-------+-------------+------------|
````

The models were:

````js
var model1 = function() {
  return sum(repeat(20, function() { return flip(0.5); }));
};

var model2 = function() {
  return sum(repeat(20, function() { return gaussian(0, 1); }));
};
````

Only `IncrementalMH` looks to have taken a hit. One possible reason is that we now perform a deep equality check on parameters when testing for erp equality. Doing so ensures that the equality check returns the expected result when an erp has an array valued parameter. (e.g. `discreteERP`.) However, it's not clear to me whether performing this deep check in `IncrementalMH` makes sense. I'll try and understand that, and I'll also a) take a closer look and check that this *is* the problem and b) see if I can make it quicker.

**Comment by dritchie** [comment]
I looked it over, and I think the changes to `IncrementalMH` all make sense. The speed hit isn't that bad. I would be curious to know where it's coming from, though. Neither of those test models have array-valued parameters, so it would seem that the culprit is some combination of (a) mapping over an key-value object of parameters, instead of an array of them, and (b) overhead from invoking `_.isEqual`.

**Comment by hawkrobe** [comment]
In your example of showing what score now looks like, I assume you mean 

```
factor(gaussianERP({mu: 0, sigma: 1}).score(x))
```

instead of invoking the score function twice? Otherwise I might have a minor complaint ;)

null-a:
@hawkrobe Well spotted, thanks! Now fixed.

stuhlmueller:
Nice, this will substantially improve code readability!

Notes:
- I agree with the discussion on slack that we should get rid of the ERP terminology: 
    - It makes sense to call the distribution objects just "distributions" (shortened to `dist` for use in places where we currently use `erp`; the base class `ERP` would become `Distribution`). 
    - I don't have a strong preference between "distribution type", "-generator", and "-class". I'd personally probably go with "type", but either would work. 
    - We should use capital letters to indicate distribution types (`gaussianERP` -> `Gaussian`).
    - All of this could happen in a follow-up PR.
- The distribution type definitions feel a little verbose due to the four required steps (create function, assign prototype, assign prototype constructor, extend with mixins and sample/score). Would it be difficult to refactor this so that we can write code that looks something like the following?

        var Uniform = makeDistributionType({
          mixins: [continuousSupport],
          sample: function() { ... },
          score: function() { ... },
          support: function() { ... }
        });

    or

        var Uniform = new DistributionType({
          mixins: [continuousSupport],
          sample: function() { ... },
          score: function() { ... },
          support: function() { ... }
        });

    One drawback of this approach is that the distribution type name will have to be specified manually, but I think this will naturally fit in with specifying other metadata (next point).

    If we follow this approach, maybe we wouldn't need the `erpCtor` macro? (If we keep it, we might want to rename it - the name is difficult to parse.)

- We might want additional metadata to be part of the distribution type specs, including parameter names, descriptions, and default values, and a description of the distribution type. This could help with error messages and auto-generated documentation. This could also happen as a follow-up PR.

Minor:
- Why does the MAP function return a `{ val, prob }` object and not a `{ val, score }` object? (This wasn't introduced by your commit, but it's particularly salient in your refactored version that the exponentiation is an extra step.)
- In the entropy function, why check for `score === -Infinity`? Isn't `Math.exp(-Infinity)` already equal to `0`?
- Use explicit parameter names in `semiMarkovRandomWalkConstrained.wppl`?
- In `multivariateRegression.wppl`, the now explicit parameter names highlight that we're passing `sigmaSquared` as the `sigma` parameter.
- Typo `erp.ad.js:115`: should be `hierarchy`
- In `smc.js:67`, you added `ad.value`. Is this related to the changes in this PR, or did we just overlook adding this previously?

null-a:
> Only IncrementalMH looks to have taken a hit.

I've added a [commit](https://github.com/probmods/webppl/pull/393/commits/88b79412086e41e27b0470deddee95b2e56674dd) which seems to fix this. New numbers:

````
|-------+----------+------------------+------------------|
| model | dev (ms) | original pr (ms) | most recent (ms) |
|-------+----------+------------------+------------------|
|     1 |     6521 |             6571 |             6038 |
|     2 |     6637 |             7086 |             6670 |
|-------+----------+------------------+------------------|
````

>  I would be curious to know where it's coming from, though.

In these tests, it was the use of `_.all`. (Only a tiny overhead, but when you're calling it ~100K times you notice.) The fix also adds special handling for parameters of type `number`, but that doesn't make any noticeable difference to the tests here. I'm not yet sure this is a good idea - I may remove it before we merge this.

> I looked it over, and I think the changes to IncrementalMH all make sense.

Great, thanks! This PR introduces another change that I'd not previously thought about. Consider these two ERPs:

````
var e1 = Enumerate(flip)
var e2 = Enumerate(flip)
````

At present I think `IncrementalMH` will consider these to be not equal, but with this PR we'll have `e1.isEqual(e2) === true`. They're considered equal because they have the same constructor (`marginal`) and their parameters are deep equal. (Marginal ERPs have one parameter, the map built by `Enumerate` during inference.) It's not clear to me whether this is desirable - I need to think about it.

null-a:
@stuhlmueller: Thanks for the review! I'll address the minor points first:

> Why does the MAP function return a { val, prob } object and not a { val, score } object?

I don't know. I'm happy to change it if you like?

> Isn't Math.exp(-Infinity) already equal to 0?

Yes, but `Math.exp(-Infinity) * -Infinity` is equal to `NaN`. The check makes `entropy` do the conventional thing when there are values in the support with zero probability. This wasn't so much of a problem before as we could only call `entropy` on marginal ERPs, and they don't include anything with zero probability. (Actually, it might have worked on any 'parameterized' ERP...)

> Use explicit parameter names in semiMarkovRandomWalkConstrained.wppl?

Good catch! Fixed.

> In multivariateRegression.wppl, the now explicit parameter names highlight that we're passing sigmaSquared as the sigma parameter.

For now I've simply renamed the variable (so I don't have to fix the inference tests!) but this might not be what the original author had in mind?

> Typo erp.ad.js:115: should be hierarchy

Fixed, thanks.

> you added ad.value. Is this related to the changes in this PR

Yes, it's related. Previously, `choiceScore` and `importanceScore` were computed in such a way that they would always be un-lifted, but that's no longer the case. (This relied on having access to un-lifted parameters in e.g. `SMC.prototype.sample` which we no longer have.) If you take this change out the test `ParticleFilterRejuvHMC - gaussianMeanVar` fails.

ngoodman:
For the meta-data that andreas mentioned, i was thinking that something like the following would be helpful for debug messages, auto-docs, etc:
~~~
{...
name: "Gaussian",
description: "The univariate Gaussian family of distributions, which has support on all reals.",
params: {mu: {description: "Mean."}, sigma: {default: 1, description: "The standard deviation (default is 1)."}}
}
~~~

It could also be useful to know what helpers are defined for each distribution. I guess this depend on the mixins and could be ready off the object directly?

null-a:
> The distribution type definitions feel a little verbose

I agree. I've added a `makeErpType` function.

> maybe we wouldn't need the erpCtor macro?

The reason it's there is that I think things are cleaner this way, not because I couldn't make it go away, so of course I think we should keep it! The main benefit of the current set-up is that we maintain a nice separation between js and wppl code. Everything in `erp.js` is plain js, and the module is added to the wppl environment as the global `erp`. This means that all the functions in there are callable from wppl as `erp.blah()`. The wppl functions which wrap the js constructors are defined in the header as regular wppl functions (no manual cps) using the macro.

Previously, everything exported by `erp.js` was added to the top level, including js functions which it's not possible to call directly. This leads to wrapper functions which make the call via `_top`, which seems a bit awkward to me.

> If we keep it, we might want to rename it - the name is difficult to parse.

Agreed. `defineErpConstructor`/ `defineDistConstructor`? It's long, but descriptive. Any better suggestions? 

> I agree with the discussion on slack that we should get rid of the ERP terminology
> All of this could happen in a follow-up PR.

My only reservation is that someone might write something like `sample(bernoulli({p: .5}))` and stare at for a long time before they realize they should have written `sample(Bernoulli({p: .5}))`. I'm not very certain that this will turn out to be a problem, so I'm more than happy to give it a go! (Maybe in the future we can catch this if we check argument types in a debug mode.)

I guess doing this in a separate PR will keep things easy to manage. Do you have any concerns about `dev` sitting in-between these two changes for a time, possibly blocking us from making a release? (If we want all these changes to land in the same release.)

null-a:
> I guess this depend on the mixins and could be ready off the object directly?

Yes, I think so. For example `entropy` is only defined on dist/erp that actually support it, so a simple run-time check will work.

````
> e=new erp.gaussian({mu: 0, sigma: 1})
gaussian({ mu: 0, sigma: 1 })
> e.entropy
undefined
````

If `entropy` was implemented for all erp/dist but threw an error when not supported, this would be less straight-forward.

stuhlmueller:
> > Why does the MAP function return a { val, prob } object and not a { val, score } object?
> 
> I don't know. I'm happy to change it if you like?

I'd change it.

> Previously, everything exported by `erp.js` was added to the top level, including js functions which it's not possible to call directly. This leads to wrapper functions which make the call via `_top`, which seems a bit awkward to me.

Agreed, the current solution is definitely nicer.

> > If we keep it, we might want to rename it - the name is difficult to parse.
>
> Agreed. `defineErpConstructor`/ `defineDistConstructor`?

Fine with me.

> My only reservation is that someone might write something like `sample(bernoulli({p: .5}))` and stare at for a long time before they realize they should have written `sample(Bernoulli({p: .5}))`. 

I share that concern, but we should be able to address it by making `sample` print a helpful error message when its called on a non-distribution.

> Do you have any concerns about `dev` sitting in-between these two changes for a time, possibly blocking us from making a release? (If we want all these changes to land in the same release.)

I don't think it's an issue; given our planned timeline for milestones, this time period can't be too long in any case. :)

What is the reasoning behind f427e427a30818d791e5c841e1cccb59f5087c5b? The `ps` no longer sum to 1, but the comment in `categorical` still says that ps is expected to be normalized.

null-a:
>>>Why does the MAP function return a { val, prob } object and not a { val, score } object?

>>I don't know. I'm happy to change it if you like?

>I'd change it.

Done.

> What is the reasoning behind f427e42?

Motivated by this [CI failure](https://travis-ci.org/probmods/webppl/jobs/123131050), I changed the `ps` to get exact answers on all versions of Node.

Good catch on the unnormalized probs - now fixed.

stuhlmueller:
One more thought: In `src/header.wppl`, the sampling functions (`multinomial`, `gaussian`, `flip`) currently allow parameters to be given as objects (`{mu: 0, sigma: 1}`) when there are multiple parameters, but not when there is a single parameter. Wouldn't it be more consistent to always allow the explicit style with named parameters?

Otherwise this is ready to be merged, right?

null-a:
> Wouldn't it be more consistent to always allow the explicit style with named parameters?

It's certainly more consistent. At the time I was thinking that naming the params seemed much less useful in the single param case, but on reflection I think you're right. Particularly in cases where the helper is named after the ERP, because then `dist(params)` is always equivalent to `sample(Dist(params))` which is nice.

I'll fix this soon.

> Otherwise this is ready to be merged, right?

The only other thing outstanding is the following:

> Consider these two ERPs:
> 
> ````
> var e1 = Enumerate(flip)`
> var e2 = Enumerate(flip)
> ````
> 
> At present I think IncrementalMH will consider these to be not equal, but with this PR we'll have e1.isEqual(e2) === true. They're considered equal because they have the same constructor (marginal) and their parameters are deep equal. (Marginal ERPs have one parameter, the map built by Enumerate during inference.) It's not clear to me whether this is desirable - I need to think about it.

I'm still not entirely sure about this. I don't _think_ it affects correctness but I'm not sure what effect it will have on the performance (run-time) of models where `Enumerate` is nested within `IncrementalMH`. We're definitely doing more work when comparing two marginal ERP for equality, the question is whether the cases where we notice the two ERP are equal allows us to save more work than we already did.

Perhaps @dritchie has a better idea about this than me?

ngoodman:
> I'm still not entirely sure about this. I don't think it affects correctness but I'm not sure what effect it will have on the performance (run-time) of models where Enumerate is nested within IncrementalMH. 

i'd go ahead and merge this in anyhow. i think the more aggressive equality check is "right", and we can back off to a coarser check if we find that speed of the fine-grained check becomes an issue.... maybe just make a note somewhere.

stuhlmueller:
> Particularly in cases where the helper is named after the ERP, because then dist(params) is always equivalent to sample(Dist(params)) which is nice.

We could even automatically generate such a helper for each ERP (if not now, then in the future; the argument order could be taken from metadata stored with the ERP).

dritchie:
> I'm still not entirely sure about this. I don't think it affects correctness but I'm not sure what effect it will have on the performance (run-time) of models where Enumerate is nested within IncrementalMH.

Right, it shouldn't affect correctness. As for performance: the only thing `IncrementalMH` uses the ERP equality check for is to determine whether a random choice needs to be rescored. So the question then becomes what's more expensive: re-scoring a marginalized ERP b/c `IncrementalMH` thinks it has changed, or doing a deep equality test to see that it in fact hasn't changed? From what I can tell, the rescore could very well be cheaper (this is just a table lookup followed by a `Math.log`, as opposed to an equality check of an entire histogram table).

So yeah, in the case of marginalized ERPs, it probably makes more sense for `IncrementalMH` to consider them always unequal. But there might be other situations where we really do need to know if marginalized ERPs are truly equal. Maybe there should be two notions of equality for ERPs---say, a `shallowEqual` and a `deepEqual` test? Idk, just throwing out ideas, here.

null-a:
**ERP `isEqual`**

Thanks for the comments.

If we stick with having a method like `isEqual` then it seems that it should be general purpose and have a clear meaning.

If we go down the route of tweaking it for `IncrementalMH` then I think we should move it out of `erp.js` and back into `incrementalmh.js`, since these tweaks would be an optimization and the trade-offs involved are specific to this algorithm.

I've just noticed that the current `isEqual` is going to do a deep equality check on the whole graph when parameters are ad nodes. Rather than fixing that (to keep it general purpose) I think my preference is to move this back into `incrementalmh.js` now. (Doing so also removes the burden from our future selves of maintaining this general purpose check, even though we don't need it yet.) If we do move it, it seems to me that we should also revert to not doing the deep check on marginals, since I don't think this PR should modify `IncrementalMH` without a good reason for doing so. (It's also an easy change to make I think.)

null-a:
>> Wouldn't it be more consistent to always allow the explicit style with named parameters?

> I'll fix this soon.

Done.

(Although I'm not sure there's any value at all in been able to do `flip({p: 0.5})`. Will this ever be used or are we just adding (a tiny) overhead? If it's consistency we're interested in, then of course this function should be called `bernoulli`.)

null-a:
I re-ran the same tests I ran when I first submitted this PR, everything still looks OK.

````
|----------------+-------+----------------+-------------|
| algo           | model | before pr (ms) | latest (ms) |
|----------------+-------+----------------+-------------|
| enumerate      |     1 |           6365 |        6187 |
| hmc            |     2 |           6904 |        6657 |
| incremental mh |     1 |           6521 |        6133 |
| incremental mh |     2 |           6637 |        6508 |
| mh             |     1 |           8331 |        7605 |
| mh             |     2 |           8046 |        7744 |
| rejection      |     1 |           8513 |        7133 |
| rejection      |     2 |           6970 |        6652 |
| smc            |     1 |           7685 |        6363 |
| smc            |     2 |           8242 |        7738 |
|----------------+-------+----------------+-------------|

````

PR #396, opened 2016-04-18
Make sampleGuide work when nested within other inference routines

stuhlmueller:
I rewrote the sampleGuide cps loop in a style that is closer to what we use for other inference algorithms. It's probable that there is some smaller change to the previous version that would have been sufficient as well (but simply defining `exit` didn't do it).

### DIFF
--- a/src/inference/sampleGuide.js
+++ b/src/inference/sampleGuide.js
@@ -20,6 +20,8 @@ module.exports = function(env) {
     this.k = k;
     this.a = a;
 
+    this.hist = new Histogram();
+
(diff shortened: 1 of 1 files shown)


null-a:
I think I just neglected to reinstate the previous coroutine at exit, which I've now [fixed](https://github.com/probmods/webppl/commit/2a2f75c422d69e4b850e6baa3f548c4bfbf2f257).

Seem ok?

stuhlmueller:
Ah, so that's what was missing – thanks!

PR #399, opened 2016-04-27
Get package names from package.json.

### LINKED ISSUE #398: Resolved package identifier uses package directory name instead of npm package name
Clone the webppl-json package into a folder with a different name:
```
git clone https://github.com/stuhlmueller/webppl-json.git wj
```
Run the test in the package and require the working directory:
```
cd wj
webppl test.wppl --require .
```

You will get an error saying `ReferenceError: webpplJson is not defined`.

This is because instead of putting the package name into scope (`webppl-json`) in camel case form, it puts the *folder name* into scope. This is because of the `read` logic in `src/pkg.js`, which uses path resolution, but does not parse `package.json` to get the correct name.

null-a:
This fixes #398.

I've also checked it by running the tests in [webppl-package-template](https://github.com/probmods/webppl-package-template). (Doing so requires [this fix](https://github.com/probmods/webppl-package-template/pull/1) merging.)

I've checked the list of packages on the wiki, and I don't see any that rely on the old wonky behavior.

I've also check the docs, and they already suggest that the global var name comes from the package name, so I don't think they need updating.

### DIFF
--- a/src/pkg.js
+++ b/src/pkg.js
@@ -46,21 +46,22 @@ var read = function(name_or_path, paths, verbose) {
       var candidate = path.resolve(candidates[0]);
       var candidatePackagePath = path.join(candidate, 'package.json');
       if (fs.existsSync(candidatePackagePath)) {
-        var name = path.basename(candidate);
+        var manifest = require(candidatePackagePath);
+        manifest.webppl = manifest.webppl || {};
+        var name = manifest.name;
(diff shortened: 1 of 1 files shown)


null-a:
The CI failure is caused by an unrelated inference test failure ('HMC - constrainedSum') which still passes locally for me.

PR #401, opened 2016-04-27
Enforce dirichlet bounds

longouyang:
Currently, `dirichletSample` can return values that are outside the support, e.g., `[0,1]`. This happens because `dirichletSample` works by sampling n Gammas and normalizing. Gamma sampling might underflow; we actually catch this inside `gammaSample` and replace 0 with `Number.MIN_VALUE`, but this isn't enough because those values get further divided when we normalize, so they become 0 again.

This PR corrects for underflow.

A couple remarks:

1. We currently don't work hard to ensure that Dirichlet samples sum to 1 and this PR continues that "tradition" (e.g., here, we correct `[0, 0, 1]` to `[ϵ, ϵ, 1-2ϵ]` but we don't correct `[0, 0.5, 0.5]` to `[ϵ, 0.5-ϵ/2, 0.5-ϵ/2]`).
2. We could further mitigate gamma underflow by switching to `expGammaSample` (a [la my proposal from a few months ago](https://github.com/probmods/webppl/pull/277#issuecomment-165332720)) but some informal testing suggested that this doesn't result in the right marginals for `dirichletDriftERP`, so I'll need to investigate more before switching.

### DIFF
--- a/src/erp.ad.js
+++ b/src/erp.ad.js
@@ -324,7 +324,7 @@ var cauchy = makeErpType({
 function sum(xs) {
   'use ad';
   return xs.reduce(function(a, b) { return a + b; }, 0);
-};
+}
 
 
 var discrete = makeErpType({
@@ -716,19 +716,33 @@ var poisson = makeErpType({
   }
 });
 
-
 function dirichletSample(alpha) {
+  var n = alpha.length;
+
   var ssum = 0;
   var theta = [];
   var t;
-  for (var i = 0; i < alpha.length; i++) {
+  for (var i = 0; i < n; i++) {
     t = gammaSample(alpha[i], 1);
     theta[i] = t;
     ssum = ssum + t;
   }
-  for (var j = 0; j < theta.length; j++) {
+
+  var numUnderflowCorrections = 0;
+  for (var j = 0; j < n; j++) {
     theta[j] /= ssum;
+    if (theta[j] === 0) {
+      theta[j] = Number.EPSILON;
+      numUnderflowCorrections += 1;
+    }
   }
+
+  for (var k = 0; k < n; k++) {
+    if (theta[k] === 1) {
+      theta[k] -= Number.EPSILON * numUnderflowCorrections;
+    }
+  }
+
   return theta;
 }
 
--- a/src/erp.ad.js
+++ b/src/erp.ad.js
@@ -150,8 +150,8 @@ function makeErpType(options) {
   // output of `console.log` when it's called on an ERP that uses the
   // default constructor.
   var erp = _.has(options, 'constructor') ?
-        options.constructor :
-        function(params) { this.params = params; };
+      options.constructor :
+      function(params) { this.params = params; };
 
   erp.prototype = Object.create(options.parent.prototype);
   erp.prototype.constructor = erp;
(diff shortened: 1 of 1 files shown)


stuhlmueller:
> We currently don't work hard to ensure that Dirichlet samples sum to 1 and this PR continues that "tradition" (e.g., here, we correct `[0, 0, 1]` to `[ϵ, ϵ, 1-2ϵ]` but we don't correct `[0, 0.5, 0.5]` to `[ϵ, 0.5-ϵ/2, 0.5-ϵ/2]`).

- In the case where we do have underflow corrections, can't we count the number of Dirichlet values `n` that are greater than (say) `ϵ * numUnderflowCorrections` and correct each such value by
`- ϵ * numUnderflowCorrections/n`? We can merge the PR as it is, but unless there are obstacles I don't see, this could be a worthwhile change. 

- Is there a test case you could add that reliably fails with the current system and succeeds with your fix?

- It would be good if the PR didn't include unrelated whitespace changes. They make it more difficult to see the actual changes. Don't worry if it's difficult to revert—I'm mainly pointing it out for future PRs.

longouyang:
Doing corrections in the way you suggest runs into subtle problems, as subtracting off fractional multiples of ϵ can misbehave due to floating point issues, e.g., `1 - (5/6*ϵ) == 1 - ϵ` and `1 - (1/6*ϵ) == 1`.

stuhlmueller:
Ok, then let's stick with the mechanism in the PR.

longouyang:
It just occurred to me that we could use a little math to mitigate this issue, e.g., use `(6 - 5ϵ)/6` instead of `1 - (5/6)ϵ`. I'll do some checking -- if this is sufficiently robust, I'll use it. If not, I'll ask you to merge the PR.

longouyang:
Latest thoughts:

- Can we keep this PR only responsible for underflow / overflow catching and leave normalization to another PR?
- Our test framework (*`test-samplers.js`*) currently assumes that support elements are real numbers and so doesn't play well with Dirichlet. Can we leave testing to another PR as well?

stuhlmueller:
The new version that only replaces 0/1 with 0/1-epsilon looks pretty uncontroversial to me, so I'll just go ahead and merge it.

null-a:
I'm not sure this is a big deal, but is there a reason we bump zeros to `Number.EPSILON` (`2.2e-16`) rather than `Number.MIN_VALUE` (`5e-324`)?

(Likewise for ones, though `1-Number.EPSILON` is closer to the next smallest representable float from `1` I think.)

longouyang:
This was to keep things symmetric. `1-Number.MIN_VALUE` is just `1`.

I just did some tests and it looks like the smallest number we can subtract from 1 is `0.5*Number.EPSILON`

So we could change the bounds to `[ϵ/2,1-ϵ/2]`

null-a:
> This was to keep things symmetric.

How come? I think I would have tried to bump to the nearest number representable in fp. (Which is `1-Number.EPSILON/2` for the upper bound as you mention, and `Number.MIN_VALUE` for the lower bound of course.)

longouyang:
I think this was influenced by my earlier fix for Beta, e.g., suppose the Dirichlet sampler returns `[0,1]` --  using symmetric bounds means that the sample still sums to 1, which isn't true if we let the lower bound go to `Number.MIN_VALUE`.

Of course, this doesn't work if the size of the Dirichlet isn't 2 and we've also punted on the summing-to-1 issue.

null-a:
> using symmetric bounds means that the sample still sums to 1

Gotcha, thanks!

Making sure the sample _always_ sums to one sounds tricky, iirc fp addition isn't even associative...

PR #412, opened 2016-05-04
Rename ERP -> Dist.

### LINKED ISSUE #400: Replace "ERP" terminology with "distribution"
Based on the discussion in #393, let's call the distribution objects just "distributions" (shortened to `dist` for use in places where we currently use `erp`; the base class `ERP` would become `Distribution`). We should also use capital letters to indicate distribution types (`gaussianERP` -> `Gaussian`).

null-a:
This PR renames `ERP` -> `Dist` etc. (see #400) throughout nearly the entire code base. The only remaining references to 'ERP' are in `src/analysis`.

I tried to not sneak any other changes in on this PR, however there are a few places where something other than the straight-forward rename happens:

* The file `erp.ad.js` is now `dists.ad.js`. I pluralized it so that when it's imported as `dists` we can use `dist` as a function arg without shadowing the import.
* In `caching.js` I choose a new name for a variable and tweaked the wording of a comment.
* `webppl` programs which return a marginal distribution now output something like the following when running under node:

```
* Program return value:

Marginal:
    1 : 0.375
    2 : 0.375
    0 : 0.125
    3 : 0.125
```

(Previously 'Marginal' was 'ERP'.)

Closes #400.


stuhlmueller:
Looking over the rewritten examples, I think [your comment](https://github.com/probmods/webppl/pull/393#issuecomment-210605915) that people might write (for example) `sample(bernoulli({p: .5}))` instead of `sample(Bernoulli({p: .5}))` does highlight a real concern. Do we have error messages that make such errors reasonably easy to debug? This could be important to have before we merge the PR, so that the process of converting existing models to the new style will be a little less painful for users.

null-a:
> Do we have error messages that make such errors reasonably easy to debug?

No. I've started #413 for this.

PR #417, opened 2016-05-06
Bump adnn version.

null-a:


null-a:
Once this is merged, maybe we could git cherry-pick it into master and release v0.6.3?

Issue #86, opened 2015-04-13
Unified Infer function for marginal inference  (with single `options` argument)


stuhlmueller:
Options can then be an object such as:

    {
      numParticles: 100,
      rejuvSteps: 1000
    }

ngoodman:
In the interest of unifying the interface, and rather than breaking the current fns, I'd prefer to see a new function `Marginal(foo, options)` that takes a thunk `foo` and takes an options object, where the options object includes both a method and method-specific options: `{method: "Enumerate", depth: 100}`.

ngoodman:
Btw. it would be neat if the options structure for SMC was somewhat compositional. E.g.:
~~~
{
  method: "Enumerate",
  numParticles: 100,
  rejuvenation: {method: "mcmcmix", 
                 kernels: [{method: "MH", iterations: 10}, 
                           {method: "HMC", leapfrogsteps: 5}]}
}
~~~

null-a:
Slight variation on the theme:
```
Marginal("MH", function() {
...
}, {steps: 100})
```
This might make it easier to pick out the method/algorithm, especially when passing the model as an anonymous function.

At the risk of revealing my ignorance, I admit that when I first looked at webppl I was confused by the way in which `Enumerate` and friends are described as marginalization operations - it was conditioning or inference that I was looking to perform. I suppose I would have been less confused by something like ```infer("MH", function() {...})``` at the time.

I get that this doesn't really work when you don't have `factor` statements, but that might be a price worth paying in order to make things easier for new users. (This assumes my experience is typical, which isn't clear. Thought this was worth sharing though!)

ngoodman:
Good point: probably we should call it `InferMarginal(...)`, and we could alias to `Marginal` and `Infer` for convenience. 

By the same token, `factor` is actually not an obvious word to those who aren't used to factor graphs... We could have an alias like `WeightExecution` or something. And we should definitely add `condition(expr)` as an alias for `factor(expr?0:-Infinity)`.

I prefer putting the method in the options object, because even though it isn't quite as easy to pick out the algroithm, it better supports the compositional descriptions of *how* to infer.

dritchie:
Just wanted to weigh in while this is still open:

[...]

PR #419, opened 2016-05-09
Add `Infer` function

null-a:
This PR completes the initial implementation of an `Infer` function.

I went with the name `Infer` since that seems to be the name people are using most often when talking about this. (Though personally I've become found of calling this `Marginal` as originally suggested!) I'm not convinced that we should have _multiple_ names for this as originally discussed, that seems likely to cause confusion to me?

Of course, I'm happy to make any changes necessary if people have strong feelings about any of this or if the thinking has changed since #86 was opened.

This PR doesn't include the recent idea of having some heuristic for choosing a good algorithm when `method` isn't specified. To avoid switching the behavior of `Infer(thunk)` on people in the future, I've chosen to not specify a default value for the `method` option of `Infer` in this initial version.

As I mentioned in #86, when deciding whether to apply the caching transform for incremental MH we now look out for `{method: 'IncrementalMH'}` in programs as well as `IncrementalMH`. This is rather crude, but hopefully good enough.

### DIFF
--- a/docs/inference.rst
+++ b/docs/inference.rst
@@ -1,57 +1,74 @@
 Inference
 =========
 
+.. js:function:: Infer(thunk, options)
+
+   :param function thunk: Program to perform inference in.
+   :param object options: Inference options.
+
+``Infer`` computes the marginal distribution on return values of a
+program. The program is specified as a function of zero arguments,
+also known as a `thunk`.
+
+The inference algorithm to use must be specified using the ``method``
+option. For example::
+
+  Infer(thunk, {method: 'Enumerate'})
+
+The following algorithms are available:
+
 Enumeration
 -----------
 
-.. js:function:: Enumerate(thunk[, maxExecutions])
+.. js:function:: Infer(thunk, {method: 'Enumerate'[, ...]})
 
-   :param function thunk: Program to perform inference in.
-   :param number maxExecutions: Maximum number of (complete) executions to enumerate.
-   :returns: Marginal ERP
+   This method performs inference by enumeration.
 
-   This method performs inference by enumeration. If ``maxExecutions``
-   is not specified, exhaustive enumeration is performed. Otherwise,
-   paths through the program are explored using a "most probable first"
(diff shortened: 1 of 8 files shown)


stuhlmueller:
Looks good.

> I went with the name `Infer` since that seems to be the name people are using most often when talking about this.

I like it.

> I'm not convinced that we should have _multiple_ names for this as originally discussed, that seems likely to cause confusion to me?

Yes, that sounds confusing.

One other comment: We could have a single `Enumerate` method and specify the search strategy as an option. Similarly, we could have a single `MH` method and specify `incremental` as an option. Either of these could happen as a separate PR.

null-a:
> We could have a single Enumerate method and specify the search strategy as an option.

Good call, I think we should do that. (As a separate PR would be my preference.)

> Similarly, we could have a single MH method and specify incremental as an option.

Do you mean having:

```
Infer(thunk, {method: 'MCMC', kernel: 'MH'}) == Infer(thunk, {method: 'MH'})
Infer(thunk, {method: 'IncrementalMH'}) == Infer(thunk: {method: 'MH', incremental: true})
```

I'm not so sure this helps too much. In my mind, the way it is at present better reflects the current implementation, and having a clear separation between kernels and methods seems useful for the purpose of confusion avoidance. More specifically, I'm worried that the above might suggest to people that these will work:

```
Infer(thunk, {method: 'HMC'})
Infer(thunk, {method: 'MCMC', kernel: 'IncrementalMH'})
```

stuhlmueller:
Sorry for the unclear MH proposal! :) 

I agree with you that we should keep kernels and methods separate. I meant:

~~~~
Infer(thunk, {method: 'IncrementalMH'}) => Infer(thunk, {method: 'MCMC', kernel: 'MH', incremental: true})
~~~~

There's a decision to be made about whether the options object should be aligned with our implementation (in which case `Infer(thunk, {method: 'IncrementalMH'})` is better), or whether it should be aligned with the functionality of the algorithms from a user's perspective.

Issue #147, opened 2015-07-09
Add test for MH with custom proposer


null-a:
While reviewing #144 I think I spotted another bug the MH acceptance probability calculation.

I have tests that demonstrate the problem and a crude fix [here](https://github.com/null-a/webppl/commit/dbc8f2718fb7534fe9890905f782f24c85a0397c), I'll work on something better.

I've tried to explain the problem in this [comment](https://github.com/null-a/webppl/commit/dbc8f2718fb7534fe9890905f782f24c85a0397c#diff-e6a81eea28f03b04580fee336cee7895R95).

stuhlmueller:
Yeah, that's a bug! Your explanation and fix makes sense to me. Can you create a pull request with a test that catches this and with your current fix (or a slight refactor)? We'll refactor this more generally as part of #145.

null-a:
This was fixed by #169.

I still need to add a test which should be easier after #119.

PR #423, opened 2016-05-10
Support drift kernels in IncrementalMH.

null-a:
This addresses #421.

IncrementalMH computes large parts of the acceptance probability incrementally as the algorithm proceeds. Since I'm not super familiar with how this goes, I can't be confident this change is correct. Ideally, someone who knows the algorithm well will confirm that this is all we need to do.

However, the existing `gaussianDrift` and `uniformDrift` tests do pass. Also, as an extra check, I've confirmed that this doesn't happen to suffer from the problem described in #147.


dritchie:
This looks fine to me.

Issue #332, opened 2016-02-11
Add run-time argument checking with nice errors for ERP and inference args


null-a:
Similar to what we have for probmods.

PR #424, opened 2016-05-11
Run-time checks for sample, score, distribution params

### LINKED ISSUE #413: Extra checks to smooth transition to new ERP interface
Some significant fraction of programs/models will need to be updated as a result of the introduction of the new ERP interface. Since the reporting of errors in webppl isn't great (yet), we want to consider what simple things we can do to smooth this process.

For reference, here are what I expect will be the most common changes that need to be made to programs.

(I'm using a specific ERP here to make the examples clearer, but this applies to all ERP of course.)

##### `sample`

 `sample(gaussianERP, [0,1]) => sample(Gaussian({mu: 0, sigma: 1}))`

Note that the sampling helper functions `flip`, `gaussian`, etc. continue to work as before.

##### `score`

`gaussianERP.score([0, 1], val) => Gaussian({mu: 0, sigma: 1}).score(val)`

A special case of this is:

```
var marginal = Enumerate(thunk)
marginal.score([], val) => marginal.score(val)
```

null-a:
Following the discussion in #413, this add checks that catch the following:

```
sample(bernoulli({p: .5}))
// => sample() expected a distribution but received "false".

Bernoulli({p: .5}).score([], false);
// => The score method of Bernoulli expected 1 argument but received 2.

Bernoulli()
// => Parameters not supplied to Bernoulli distribution.

Bernoulli({})
// => Parameter "p" missing from Bernoulli distribution.
```

The last of those wasn't suggested in #413, but it seems useful and was asked for anyway in #332.

I ran a few of the tests I ran when we merged the new ERP interface (#393) to see how much time is spent on these checks. It looks like we give back some of the gains we got from the new interface, but things aren't noticeably slower than what we currently have in `dev`.

```
|-----------+-------+------------------+-----------------+----------|
| algo      | model | before #393 (ms) | after #393 (ms) | now (ms) |
|-----------+-------+------------------+-----------------+----------|
| enumerate |     1 |             6365 |            6187 |     6384 |
| mh        |     1 |             8331 |            7605 |     8025 |
| rejection |     1 |             8513 |            7133 |     7798 |
|-----------+-------+------------------+-----------------+----------|
```

Closes #413.

### DIFF
--- a/src/dists.ad.js
+++ b/src/dists.ad.js
@@ -29,9 +29,6 @@
 //
 // - All distributions of a particular type should share the same set
 //   of parameters.
-// - All distribution constructors should take a single params object
-//   argument and store a reference to it as `this.params`. See
-//   `clone`.
 
 'use strict';
 
@@ -157,12 +154,17 @@ function makeDistributionType(options) {
     };
   }
 
+  var extraConstructorFn = options.constructor;
+
   // Note that Chrome uses the name of this local variable in the
   // output of `console.log` when it's called on a distribution that
   // uses the default constructor.
-  var dist = _.has(options, 'constructor') ?
-        options.constructor :
-        function(params) { this.params = params; };
+  var dist = function(params) {
+    this.params = params;
+    if (extraConstructorFn !== undefined) {
+      extraConstructorFn.call(this);
+    }
+  };
 
   dist.prototype = Object.create(options.parent.prototype);
   dist.prototype.constructor = dist;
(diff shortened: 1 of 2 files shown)


stuhlmueller:
Nice! Thanks for profiling!

For the parameter specs, I wonder if we should use arrays instead of space-delimited strings. Either works for now, but I imagine that, before long, we'll want to add more metadata, and then `params` will likely be an array, with objects for parameters that have metadata (e.g., `[{name: 'mu', description: 'vector of means'}, {name: 'cov', description: 'covariance matrix'}]`). Using arrays right away seems a little more forward-looking. But it won't be a difficult to change to make, so I'd be happy to merge the PR as is, too.

null-a:
> I wonder if we should use arrays instead of space-delimited strings

Yeah, I'll definitely be changing this in the PR that adds distributions to the docs, but I wasn't sure exactly how it would look so I just added what I needed and figured I'd evolve it as I go.

I've already started the distribution docs and will probably submit a PR tomorrow, so merging this as is would suit me.

Issue #314, opened 2016-02-09
Stack traces


stuhlmueller:
We'd like to have full stack traces (tracebacks) when errors happen, going beyond immediate error locations (#292). This is challenging, since the CPS transform eliminates the stack that is usually used for tracebacks.

In CPS, the stack is represented explicitly using the continuation function. One could try to approach this issue by associating extra information with this function, such as the file and location where it is defined and a pointer to the continuation function that corresponds to the next stack frame. I don't know how difficult it would be to do the latter.

ngoodman:
Alternatively, perhaps the trampoline function could be responsible for keeping around a stack trace: each time it is called it pushes the current location, on return it pops. If this incurs a lot of overhead we could have a debug mode where the tracing trampoline is used.

stuhlmueller:
Yes, it could make sense for the [trampoline runner](https://github.com/probmods/webppl/blob/dev/src/util.js#L9) to keep around a stack. It's only called once itself (apart from a few exceptional circumstances), so the modification of the stack would have to originate elsewhere. We could extend the code built by the trampolining transform such that it adds to the stack in places that previously were [compound function calls](https://github.com/probmods/webppl/blob/dev/src/transforms/trampoline.js#L41), and pops in places that were return statements. (This may be what you meant.)

stuhlmueller:
Do we have a plan for this? If it's not clear how to do it, we can push it to 1.0 or later.

null-a:
> Do we have a plan for this?

I don't. Let's discuss it once 0.8 is out of the way.

longouyang:
Would it be enough for the trampoline runner to keep track of the current address? (i.e., can we always map addresses back to locations in the user code or the header?)

If so, I think we'd have to keep track of `s`, `k`, and `a` in the runner -- I'm not sure if this would break anything else.

null-a:
Assorted thoughts on this:

> perhaps the trampoline function could be responsible for keeping around a stack trace: each time it is called it pushes the current location, on return it pops.

> We could extend the code built by the trampolining transform such that it adds to the stack in places that previously were compound function calls, and pops in places that were return statements. 

I'm still a little hazy on exactly what this would look like, but it seems to me that just maintaining a single stack somewhere won't be sufficient. When we invoke a continuation we may need to effectively replace the current stack with an entirely different one, and if we're only maintaining a single stack I'm not sure where the information to do that comes from? Am I missing something?

[...]

PR #425, opened 2016-05-12
Source mapped errors when running under Node

null-a:
This is a PR for the work that's been going on in #292. Conceptually, there are two main changes included here:

1. Finish the work that allows us to map the source of a run-time error back to its original location in webppl code. (The other relevant changes were #327, #357.)
2. Use this information to show more readable errors when running webppl under Node. (This correctly handles `webpplEval` and code loaded from packages.)

Notably it doesn't:

1. Fix the stack trace. (#314)
2. Add readable errors to the browser editor. However, the information made available here is a prerequisite for that change. (probmods/webppl-editor#10)

Here's some example output:

```js
var a = function () {
    return {a:1}[notDefinedVariable]
}
return a()
```

```
notDefinedVariable is not defined
    at test.wppl:2

1| var a = function () {
2|     return {a:1}[notDefinedVariable]
   -----------------^
3| }

[stack trace...]
```

An example where the error is raised by code in `header.wppl`:

```js
map('not a function', [0])
```

```
fn is not a function
    at header.wppl:301

300| var map = function(fn, l) {
301|   return map_helper(0, l.length - 1, function(i) { return fn(l[i]) })
     ----------------------------------------------------------^
302| }

[stack trace...]
```


stuhlmueller:
Looks very promising!

I tried it on the following invalid program:

~~~~{js}
var foo = null;
foo[1];
~~~~

It seems to run into an error during error processing:

~~~~
$ webppl ~/Documents/testme.wppl
Processing /Users/stuhlmueller/Documents/testme.wppl
fs.js:584
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                                                  ^

Error: ENOENT: no such file or directory, open '<anonymous>'
    at Error (native)
    at Object.fs.openSync (fs.js:584:18)
    at Object.fs.readFileSync (fs.js:431:33)
    at Object.showFriendlyError (/Users/stuhlmueller/Projects/webppl/src/errors.js:62:10)
    at run (/Users/stuhlmueller/Projects/webppl/webppl:45:12)
    at main (/Users/stuhlmueller/Projects/webppl/webppl:133:3)
    at Object.<anonymous> (/Users/stuhlmueller/Projects/webppl/webppl:136:1)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
~~~~

On the other hand, if I place the following content in the same file...

~~~~{js}
var foo = null;
foo();
~~~~

...I get a nice readable error message as expected. Any idea what might be happening?

null-a:
Oops, good catch!

Seems like in this program, the top most stack frame isn't very informative. I think it might come from an anonymous function within the trampoline or something like that. (The error comes from not correctly handling the failure to match this to the original code.) None of the frames below that will help either, so I'll have to think about whether there's anything sensible we can do here, or whether we should instead postpone merging this until we've done the additional work to recover the stack. (#314)

null-a:
> Any idea what might be happening?

It turns out the CPS transform doesn't always pass on location information. I've pushed a fix for this specific case, so your example now works.

The example you found isn't the only problem, there are lots of other places in the CPS transform where the same thing can happen. It may be that the remaining cases are much harder to trigger (at least I found it tricky), but rather than try to figure out which of them might be problematic I think I will just apply the same fix everywhere. This will be fairly straight-forward with the code I've written already I think.

Since it's quite possible that there will be examples of this that I miss, I've made the error you see when it happens more informative. For example, without the fix, the example you gave would now show something like this:

```
Failed to generate friendly error message. Original error was:

Cannot read property '1' of null

Stack trace:
  at ...
```

null-a:
>  rather than try to figure out which of them might be problematic I think I will just apply the same fix everywhere.

Having played with what we have at moment a little more, I find myself pretty reluctant to do this.

There are more candidates for where this change _could_ be made than I originally thought. At the same time, I've not been able to find any more cases where this is actually causing a problem.

Rather than make a fairly big change that could be unnecssary or even incorrect, I'd now prefer to leave this as it is and fix any problems as we come across them.

It may be that working on fixing-up the stack trace will help turn up any problems I'm missing. In the mean time, the better error message I added should help anyone who runs into the problem - we could even update the text to specifically ask people to open an issue when they see it.

src/errors.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+}
+
+function getSrcFromMap(sourceMap, fileName) {
+  var mapConsumer = new SourceMap.SourceMapConsumer(sourceMap);
+  return mapConsumer.sourceContentFor(fileName);
+}
+
+function writeFriendlyError(error, pos, src) {
```
stuhlmueller: This sounds very similar to `showFriendlyError`. Give it a more distinctive name, or inline it?

src/main.js
```diff
@@ -196,8 +230,19 @@ global.webpplEval = function(s, k, a, code, runner) {
...
-  var compiledCode = compile(code);
-  return eval.call(global, compiledCode)(util.trampolineRunners[runner])(s, k, a);
+  var codeWithMap = compile(code, {filename: 'webppl:eval'});
+  try {
+
+    // Generally CPS and exceptions don't mix very well. However, I
+    // think catching errors will work here because the eval'd code is
+    // run in its own trampoline, and therefore within this try/catch.
+    // In contrast, if the eval'd code ran in the top-level trampoline
+    // this approach wouldn't work.
```
stuhlmueller: This depends on the trampoline used. If we're using the "pausy" web trampoline runner (with `setTimeout`), this won't always work correctly.

longouyang: We could make the pausy runner take an error-handling function as an argument (this would also simplify cross-browser error handling a little, see probmods/webppl-editor#9)

null-a:
> Give it a more distinctive name, or inline it?

Done. (Inlined.)

> This depends on the trampoline used.

Good catch! Fixed.

> We could make the pausy runner take an error-handling function as an argument (this would also simplify cross-browser error handling a little,

I did something similar to this, but I doubt it's exactly what you had in mind so it might not help much by itself. Unless my change is obviously incompatible with some other changes we know we need to make, shall we go with this for now?

longouyang:
The trampoline fix looks good to me (note to self -- we'll have to change webppl-editor to use the new parameterized pausy runner)

stuhlmueller:
I tried the most recent version of this branch on the following program, which should cause an error since we are converting a circular structure to JSON:

~~~~{js}
var f = function(){
  var a = { };
  var b = { a: a };
  _.assign(a, {b: b});
  return JSON.stringify(a);
};

f();
~~~~

I expected to see an error that looks something like this:

~~~~
/Users/stuhlmueller/Documents/runtime-error.wppl:30
  return JSON.stringify(a);
              ^

TypeError: Converting circular structure to JSON
    at Object.stringify (native)
    at f (/Users/stuhlmueller/Documents/runtime-error.wppl:30:15)
    at Object.<anonymous> (/Users/stuhlmueller/Documents/runtime-error.wppl:33:1)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Function.Module.runMain (module.js:457:10)
    at startup (node.js:136:18)
    at node.js:972:3
~~~~

I got this message:

~~~~
Processing /Users/stuhlmueller/Documents/runtime-error.wppl
/Users/stuhlmueller/Projects/webppl/node_modules/source-map/lib/source-map-consumer.js:539
      throw new TypeError('Line must be greater than or equal to 1, got '
      ^

TypeError: Line must be greater than or equal to 1, got null
    at SourceMapConsumer_findMapping [as _findMapping] (/Users/stuhlmueller/Projects/webppl/node_modules/source-map/lib/source-map-consumer.js:539:13)
    at SourceMapConsumer_originalPositionFor [as originalPositionFor] (/Users/stuhlmueller/Projects/webppl/node_modules/source-map/lib/source-map-consumer.js:604:22)
    at getErrorPosition (/Users/stuhlmueller/Projects/webppl/src/errors.js:121:38)
    at Object.showFriendlyError (/Users/stuhlmueller/Projects/webppl/src/errors.js:55:13)
    at run (/Users/stuhlmueller/Projects/webppl/webppl:45:12)
    at main (/Users/stuhlmueller/Projects/webppl/webppl:133:3)
    at Object.<anonymous> (/Users/stuhlmueller/Projects/webppl/webppl:136:1)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
~~~~

It looks like source map handling ran into a problem. Is this something we could easily fix?

null-a:
Oops. It will be the `(native)` that's the problem. I'd not thought of that, nor did I run into it while testing. Thanks!

> Is this something we could easily fix?

Yes.

null-a:
Errors in native code now look something like this:

```
Processing native.wppl

Failed to generate friendly error message. Original error was:

Converting circular structure to JSON

Stack trace:
    at Object.stringify (native)
    at eval (eval at <anonymous> (/path/to/webppl/src/main.js:222:10), <anonymous>:14:42)
    at /path/to/webppl/src/util.js:49:13
    ...
```

PR #427, opened 2016-05-13
Distribution documentation

null-a:
This adds basic documentation for distributions. I've shared a preview of the [HTML version](https://dl.dropboxusercontent.com/u/501760/webppl/html/distributions.html).

As written, we need to check the generated docs into git. This is because readthedocs doesn't support using node as part of the build process, as best as I can tell. Other than re-writing this in Python, I don't see a way around this at the moment.

(I think [these](http://read-the-docs.readthedocs.io/en/latest/development/buildenvironments.html) are the relevant docs.)

I've not added a great deal of extra information to the docs besides the list of distributions and their parameters. (I didn't want to waste time adding stuff that would be of little value.) So, if anyone spots anything that will definitely be useful, let me know and I'll add it. Otherwise, I think we can just add to this as we go?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -140,4 +140,9 @@ module.exports = function(grunt) {
       throw e;
     }
   });
+
+  grunt.registerTask('generate-docs', 'Generate documentation.', function() {
+    var output = child_process.execSync('scripts/distributionDocs > docs/distributions.rst');
+    grunt.log.writeln(output);
+  });
 };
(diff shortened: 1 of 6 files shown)


stuhlmueller:
Looks great! I think it's fine to check the generated docs into git in this case; it's consistent with our other docs being present in the repo as well. 

We could consider a server-side `pre-receive` or `update` git hook that checks that the docs are up to date with the code and that rejects updates where this isn't the case. (Alternatively, we could consider some way of automatically updating the docs in the repo.)

PR #428, opened 2016-05-13
Replace multiple Enumerate methods with strategy option.

null-a:
As suggested in #419.

### DIFF
--- a/docs/inference.rst
+++ b/docs/inference.rst
@@ -30,19 +30,20 @@ Enumeration
 
       Maximum number of (complete) executions to enumerate.
 
-   If ``maxExecutions`` is not specified, exhaustive enumeration is
-   performed. Otherwise, paths through the program are explored using
-   a "most probable first" heuristic until the maximum number of
-   executions is reached.
+      Default: ``Infinity``
 
-   Alternatively, the search strategy can be specified explicitly
-   using the ``EnumerateBreadthFirst``, ``EnumerateDepthFirst`` or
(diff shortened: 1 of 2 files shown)


stuhlmueller:
Would you be opposed to using full strategy names (`likely-first`, `breadth-first`, `depth-first`) instead of `lfs`/`bfs`/`dfs`? This would make inference operators more interpretable for people who don't read the docs.

null-a:
> Would you be opposed to using full strategy names

Not at all! I'll take care of it.

stuhlmueller:
Thanks! The PR looks good otherwise, so feel free to merge once you've made that change.

By the way, I'm not sure if we should use camelCase or lisp-case for parameter value strings. The former is more consistent with the naming of keys, and with JavaScript conventions in general, but the latter feels more readable. I'd probably go with camelCase, but not without a tinge of sadness in my lispy heart...

null-a:
>  so feel free to merge once you've made that change.

I've already made the change! I noticed the deviation from camelCase but took your lisp-case suggestion literally anyway. 

On a related note, I realised that because of the way I implemented `Infer`, the values for the `method` key have inherited the convention of using initial capitals from our coroutine naming scheme. (e.g. `IncrementalMH`, `Enumerate`.) Perhaps _all_ of these need to follow a single convention? (Separate PR?)

ngoodman:
it's probably a good idea to use a common convention for params structures in Infer. i'm fine with either camel of lispy. separate PR for sure...

Issue #414, opened 2016-05-05
Add `score` helper


null-a:
It’s probably confusing that sometimes the correct way to interact with a distribution is to call a method on it (e.g. `dist.score(x)`), and sometimes it’s to call a top-level function (e.g. `sample(dist)`). Perhaps we should add `score(dist, x)` for consistency, and promote its use.

Note that we already have `entropy` and `MAP` helpers that serve a similar purpose to the proposed `score` function.

@stuhlmueller pointed out that a minor consideration here is that people commonly use `score` as a variable name.

longouyang:
We could alias it to `scorer`.

ngoodman:
i have mixed feelings about this. on the one hand, we do need to clean up the confusion about `dist.score(val)` vs `sample(dist)`. introducing `score(dist, val)` is the most straightforward solution. 

on the other hand, it is much more idiomatic in js to use method calls, and i like `dist.score(val)`. the other solution would be to make `dist.sample()` alias to the co-routine sampler within webppl. this might be tricky to do, and kind of ugly while #358 is open...

stuhlmueller:
If we're planning to address #358 before 1.0 (which would be desirable, as it would help a lot with making bigger models more modular), I'd just keep `sample`/`score` as they are for now, and then introduce `dist.sample()` as the canonical way to sample from distributions. If we don't expect that we'll get to #358 for 1.0, I'm in favor of introducing the `score` helper.

Issue #358, opened 2016-02-25
Allow webppl functions to be called using member expressions


stuhlmueller:
For complex programs, it is useful to bundle up functions into objects. For example, we might want to represent agents as follows:

~~~~{js}
var agent = { 
  act: function(state){...}, 
  expectedUtility: function(state, action){...},
  params: ...
};
~~~

Currently, we first have to reassign methods to use them (`var act = agent.act; act(...);`) so that they don't get interpreted as primitive function calls. This makes the resulting models less readable. There should be a way to directly call webppl functions as methods.

Here are two possibilities that come to mind (I'm open to others):

1. We could do lexical analysis to determine whether an (object) variable is bound within a webppl program; if it is, any method calls will be interpreted as compound calls, not primitives.
2. We could have a macro that marks objects for which method calls will be interpreted as compound. This could be something home-baked, or we could use the ES6 class syntax and decide that all classes defined in this way will support method calls.

null-a:
Another possibility (which I'm sharing, but don't know is a good idea) would be to make this work by using something other than member expressions for our foreign function interface. This is probably a bigger change than the first 2 suggestions, but I think it has at least some merit.

1. I wonder if it might be easier to use the language if we make it a little more obvious where the boundary between webppl and JS lies. In particular I'm thinking about people who are new to the language. I wonder whether once someone knows you can seamlessly call JS functions from webppl, it's likely they'll make the inference that the two interoperate seamlessly in general, and assume you can pass/call anything you like across the boundary. I might have seen this happen once, though those of you who have taught webppl will have a much better idea about this than I do of course!
2. As we write more packages the current approach of adding all webppl functions from all loaded packages to the top-level seems likely to break down. There are other ways to handle this, but perhaps in the future calling webppl functions via member expressions will be common enough to dedicate the member expression syntax to it.

stuhlmueller:
Rethinking the FFI is a good idea and should be done before 1.0. I also agree that adding all webppl functions from packages to the top-level is not a long-term solution. (This would likely come up when we rethink require/import anyway.)

For now, a good next step would be to make a comprehensive list of all of the member expression calls that we have been using (in examples, dippl, forest, etc). Then we can think about what each instance would look like under different proposals.

stuhlmueller:
A workaround that may improve code readability in some situations:

~~~~{javascript}
var call = function(method, arg1, arg2, arg3) {
  return method(arg1, arg2, arg3);
}

var world = {
[...]

PR #440, opened 2016-05-20
Add score helper to the header.

null-a:
This adds a `score(dist, val)` helper that is equivalent to `dist.score(val)`. See #414 for the rationale behind this.

I don't know whether there's a consensus that this is a good idea.

If we do decide to add this, it makes sense to do it soon so that its use can be promoted in documentation etc.


stuhlmueller:
Should we close this for now, and re-open if it looks like we won't get to #358 (member exprs) before 1.0?

null-a:
Sure.

PR #441, opened 2016-05-20
Add Delta distribution.

null-a:
At present `Delta` is implemented in terms of `Categorical`, this change adds a separate `Delta` distribution that is equivalent to that.

The main benefit of adding this now is to have `Delta` show up in the docs.

This might be overkill if we only wanted the documentation, but having a `Delta` distribution has proved useful in VI so I think we'll want to add this at some point anyway.


ngoodman:
it might be worth adding some tests for Delta to this PR? does mcmc do the right thing with a Delta? what about when the support is a random value? eg:
~~~
function() {
  var v = gaussian(0,1)
  var d = delta(v)
  factor(d)
  return d
}
~~~

null-a:
> it might be worth adding some tests for Delta to this PR?

Sure thing. FWIW, the only reason I didn't is that I think of this as been part of #246, during which I choose (for better or worse) not to try and improve test coverage.

> does mcmc do the right thing with a Delta?

Nope.

PR #464, opened 2016-05-25
Fix binomial score edge cases

longouyang:
This PR fixes up some issues for binomial scorer:
- fixes a bug (score is `NaN` when `p` is 0 or 1)
- makes sure that the scorer returns `-Infinity` if the value is outside the support
- switches to an exact implementation of the log-factorial function
- adds some tests


src/dists.ad.js
```diff
@@ -616,12 +616,42 @@ var Binomial = makeDistributionType({
     'use ad';
     var p = this.params.p;
     var n = this.params.n;
-    // exact formula
-    return (lnfact(n) - lnfact(n - val) - lnfact(val) +
-            val * Math.log(p) + (n - val) * Math.log(1 - p));
+
+    if (!_.contains(this.support(), val)) {
```
stuhlmueller: I'm not sure if it matters much, but we could make this a bit more performant for large `n` by avoiding building the entire list of support elements and instead just checking whether `val >= 0 && val <= n  && val % 1 == 0`.

longouyang: Added in ba94037

src/dists.ad.js
```diff
@@ -731,7 +761,18 @@ function lnfact(x) {
   return sum;
 }
 
-
+function lnfactExact(x) {
+  'use ad';
```
stuhlmueller: Throw an error if `x < 0`?

longouyang: Added in	02ac3a7

stuhlmueller:
Looks ready to merge once the conflicts are resolved.

longouyang:
Gah, fixing the conflicts was somehow a disaster. I'll redo this in a different branch.

PR #474, opened 2016-06-03
Fix #473 for multivariate gaussian sampler

### LINKED ISSUE #473: Bug in multivariate Gaussian sampler
The function multivariateGaussianSample in src/dists.ad.js does not produce a sample from the multivariate Gaussian distribution due to an error. The transpose operation on line 336 is erroneous as numeric.svd returns V which contains eigenvectors arranged in a column-wise fashion.

This error was discovered by Siddharth Ancha, Roger Grosse and Daniel Roy, using an implementation of bidirectional Monte Carlo (BDMC) in WebPPL, when theoretically impossible results were observed in models using the multivariate Gaussian.

siddancha:
Remove transpose on svd.V while computing scaledV.


null-a:
Good catch! Thanks for putting this PR together, it looks good to me.

Issue #323, opened 2016-02-10
Clean-up header.wppl


null-a:
Andreas said:

>  I think there are too many functional programming utils in the header. we should keep the important ones and move the rest to a package

null-a:
We should also consider splitting `header.wppl` into a number of smaller files, as suggested in #353.

null-a:
To help me understand which header functions we can move out to a package I've [counted](https://gist.github.com/null-a/7890b9640c5e3736f0d438525e172141) how often each function is used across the webppl examples, inference tests, forest and agent models.

I intend to share a list of which functions I suggest we move out in the next week or so.

In the meantime, please let me know if there is any more webppl code floating around that I can use in this analysis. Thanks.

null-a:
As promised, here's the list of functions I suggest we move out of the header:

```
plus
minus
mult
div
and
or
eq
neq
lt
gt
leq
geq
isEven
isOdd
cons
snoc
rest
mapN
ringForward
ringBackward
mapPairs2
mapPairsC
mapPairsNC
mapReduce
mapReduce1
insertAt
removeAt
replaceAt
drop
take
dropWhile
takeWhile
span
push
everyOther
sortOn
```

I'm planning to add all of these to a package except for the functional versions of the binary operators (`plus`, `minus` etc.) as it sounds like these only exist because they are left over from a time when webppl didn't support binary ops. I'll probably do this in a few days time.

None of these were used in the code I looked at, details [here](https://gist.githubusercontent.com/null-a/9c62722e320ba8fe1f74fad088648dd5/raw/2458f9a54a3cc94b58552944bf1ea1b3aeb68dff/header-fn-analysis.txt).

ngoodman:
the only of these that i can imagine using (e.g. in probmods) are `rest`, `take`, and `drop`. we should keep `rest` for historical reasons. maybe there are more idiomatic ways to do `take`/`drop` anyhow?

otherwise sounds good!

null-a:
Great, thanks! Will make sure to hang on to `rest`.

> maybe there are more idiomatic ways to do take/drop anyhow?

`rest(arr)`, `drop(..., arr)` and `take(..., arr)` are just thin wrappers around `arr.slice(...)`, which I suppose is the idiomatic JS thing.

null-a:
Before I start on this, it's probably worth having a quick think about what we're going to call the package that these functions will end up in.

[...]

PR #480, opened 2016-06-08
Have indexOf follow the behavior of the JS version

null-a:
Returning a falsey value when the item is not found isn't super useful
since zero is falsey too. Rather than giving the impression that this
might be equivalent to something like _.contains when used as a
predicate, I think it's clearer to have this follow the behavior of the
JS version.

I noticed this while working on #323. None of the 3 uses of `indexOf` I found there would be affected by this change.

### DIFF
--- a/src/header.wppl
+++ b/src/header.wppl
@@ -542,9 +542,7 @@ var takeWhile = function(p, ar) {
 };
 
 var indexOf = function(x, xs) {
-  // prototype method doesn't return falsy value if not found
-  var i = xs.indexOf(x);
-  return i < 0 ? undefined : i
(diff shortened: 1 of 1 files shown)


stuhlmueller:
If it just redirects to the array prototype method, do we need it at all? Why not use that directly?

null-a:
The reason I didn't remove it is that I didn't want to break any code that relies on it, but of course we could.

stuhlmueller:
Given your count of 3 uses, I think it's reasonable to remove it.

null-a:
Ok, in that case, I'll just remove it as part of #323.

PR #481, opened 2016-06-08
Drop support for node < 0.12.4.

null-a:
Older versions of node aren't currently compatible with adnn. Short of a compelling reason to do otherwise, it seems like the sensible thing to do is drop support for these versions from webppl too.

We might also consider dropping support for the 0.12 series? My experience has been that if I write something that doesn't work on 0.10 and 0.11, it tends not to work on 0.12 either, so dropping support for that might save us some effort that we can better spend else where.


stuhlmueller:
I think it's fine to drop 0.12 support. We might want to print a warning on startup when webppl is run on an unsupported node version.

PR #483, opened 2016-06-08
Clean up verbose output of main webppl script

stuhlmueller:
Previously:

~~~~
$ webppl ~/Documents/test.wppl
Processing /Users/stuhlmueller/Documents/test.wppl

* Program return value:

true
~~~~

Now:

~~~~
$ webppl ~/Documents/test.wppl
false
~~~~

This feels cleaner and is more in line with the command-line output of other interpreters.

### DIFF
--- a/webppl
+++ b/webppl
@@ -22,8 +22,7 @@ function printWebPPLValue(x) {
   }
 };
 
-function topK(s,x) {
-  console.log('\n* Program return value:\n');
+function topK(s, x) {
   printWebPPLValue(x);
 };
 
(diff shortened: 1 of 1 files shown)


ngoodman:
i think this is ok, but i'd then like to have a --verbose mode that gives more information about what's happening (so that i know whether time is spent in compiling, or running, etc).

stuhlmueller:
Sounds useful, but that request seems independent of this PR—the log statements we're removing take place at the very beginning and at the very end, so we didn't get any info on what time is spent on before either.

stuhlmueller:
Actually, we do have verbose mode that prints the information you want, even after this PR:

~~~~
$ webppl ~/Documents/test.wppl --verbose
parsePackageCode: 1799.440ms
compile: 648.036ms
false
run: 2.429ms
~~~~

ngoodman:
oh. ok then. ;)

PR #491, opened 2016-06-11
Use webppl-editor for homepage

longouyang:
Demo is up at https://web.stanford.edu/~louyang/wp-site-core/

The homepage is now powered by a React app that uses CommonJS `require()` to include webppl-viz and webppl-editor and assumes that browserified webppl.js is available (I'm currently building *`webppl.js`* by temporarily checking out the version declared in *`webppl-version.txt`*, running `grunt browserify`, and copying the resulting file back into the gh-pages branch).

Build with `make build` (changes to *`src/index.js`* won't be reflected on webppl.org unless you do this before you commit and push)

(Tested with node 6 and npm v3)


src/index.js
```diff
@@ -0,0 +1,534 @@
...
+  }
+};
+
+// Extend _.indexOf to work with functions instead of values
+// Based on http://stackoverflow.com/questions/12356642/
+var indexOfValue = _.indexOf; // save a reference to the core implementation
+
+_.mixin({
+    // return the index of the first array element passing a test
+    indexOf: function(array, test) {
```
null-a: This looks similar to [`findIndex`](http://underscorejs.org/#findIndex).

longouyang: Switched to `findIndex` in c47cdd5 -- thanks!

stuhlmueller:
What's the story for integration with dippl.org (where we need the same literate editor)?

longouyang:
Can you check the Google analytics for dippl to see how many non-Stanford people actually use the dippl editor?

If there's not many people, I think we could just eliminate the dippl literate editor.

stuhlmueller:
Good idea. In the last 30 days, there were only five distinct visitors to `editor.html` outside of Stanford / Palo Alto / Sunnyvale, and none seem to be using it with any regularity. The number of people in our area is similarly low. We can probably get rid of it.

stuhlmueller:
It would be good to improve the CSS a bit before we merge this PR. At the moment, the movement buttons aren't aligned with the boxes, there's not space between movement buttons and code boxes, the run buttons are equidistant between boxes above and below (there should be more space below), and the output boxes are closer to the code box below than to the (run button of the) code box above that they belong to. If you want me to make these changes, you can give me push access to your repository.

longouyang:
Okay, I gave you push access.

Issue #451, opened 2016-05-23
Extend Distribution docs


null-a:
Integrate changes made to distribution docs on the `docs` branch.

null-a:
Closing this, as I think all of the general ideas in the docs branch for improving our distribution docs have now been integrated.

PR #493, opened 2016-06-11
Fill out descriptions in distributions docs

hawkrobe:
Incorporates some of the direct changes from the docs branch into the dists.ad.js source so that docs can be automatically built.

Partially addresses #451

### DIFF
--- a/src/dists.ad.js
+++ b/src/dists.ad.js
@@ -229,7 +229,7 @@ var Uniform = makeDistributionType({
 
 var UniformDrift = makeDistributionType({
   name: 'UniformDrift',
-  decs: 'Drift version of Uniform. ' +
+  desc: 'Drift version of Uniform. ' +
       'Drift kernels are used to narrow search during inference. ' +
       'UniformDrift proposes from a symmetric window around the current value x, [x-r, x+r]',
   params: [{name: 'a', desc: 'Lower bound (real)'},
--- a/src/dists.ad.js
+++ b/src/dists.ad.js
@@ -208,7 +208,8 @@ function makeDistributionType(options) {
 var Uniform = makeDistributionType({
   name: 'Uniform',
   desc: 'Continuous uniform distribution on [a, b]',
-  params: [{name: 'a'}, {name: 'b'}],
+  params: [{name: 'a', desc: 'Lower bound (real)'},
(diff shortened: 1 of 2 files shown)


src/dists.ad.js
```diff
@@ -353,7 +365,10 @@ function multivariateGaussianScore(mu, cov, x) {
 
 var MultivariateGaussian = makeDistributionType({
   name: 'MultivariateGaussian',
-  params: [{name: 'mu', desc: 'mean vector'}, {name: 'cov', desc: 'covariance matrix'}],
+  desc: 'n-dimensional Gaussian.',
+  params: [{name: 'mu', desc: 'mean vector (array of reals)'},
+	   {name: 'cov', desc: 'covariance matrix  (array of array of reals' +
```
stuhlmueller: There's a space missing at the end of this string (currently compiles to `realsthat`).

src/dists.ad.js
```diff
@@ -228,7 +229,12 @@ var Uniform = makeDistributionType({
 
 var UniformDrift = makeDistributionType({
   name: 'UniformDrift',
-  params: [{name: 'a'}, {name: 'b'}, {name: 'r', desc: 'drift kernel radius'}],
+  desc: 'Drift version of Uniform. ' +
+      'Drift kernels are used to narrow search during inference. ' +
+      'UniformDrift proposes from a symmetric window around the current value x, [x-r, x+r]',
+  params: [{name: 'a', desc: 'Lower bound (real)'},
+	   {name: 'b', desc: 'Upper bound (real > a)'},
```
stuhlmueller: It would be good to standardize the capitalization of parameter descriptions, and either always capitalize or never. The current docs do the latter.

hawkrobe:
It might be useful at some point in the future to separate out the description of a parameter from the values it can take -- currently provided in parentheses -- so that we can format them differently in the generated docs. When Long & MH directly modified the .rst, for instance, they put the description in italics but left the set of values in plain text.

null-a:
> It might be useful at some point in the future to separate out the description of a parameter from the values it can take

This is already underway. See the email discussion from yesterday for details.

hawkrobe:
ah, right -- I understand what that's about now :) sorry for the confusion.

null-a:
No problem at all! (I only mentioned it so you know that I plan to take care of it.)

PR #496, opened 2016-06-13
Don't mutate arguments in binomialSample

### LINKED ISSUE #495: Irrelevant sampling affects MCMC marginal
This model

```
var n = 15
var model = function() {
  var p = uniform(0,1)    // sample a coin weight from a uniform distribution
  var totallyIrrelevantGarbage = binomial(p, n)
  return p
}

Infer({samples: 10000, burn: 5000, verbose:true}, model)
```

does not return uniform. Changing `binomial` to `gaussian` returns uniform, appropriately.

Some notes [as an issue at RWebPPL](https://github.com/mhtess/rwebppl/issues/20)

Thanks to @hawkrobe for discovering this

longouyang:
Would fix #495

### DIFF
--- a/src/dists.ad.js
+++ b/src/dists.ad.js
@@ -600,7 +600,8 @@ function binomialG(x) {
 
 // see lemma 6.1 from Ahrens & Dieter's
 // Computer Methods for Sampling from Gamma, Beta, Poisson and Binomial Distributions
-function binomialSample(p, n) {
+function binomialSample(_p, _n) {
+  var p = _p, n = _n; // avoid mutating arguments
   var k = 0;
   var N = 10;
   var a, b;


hawkrobe:
Ah, nice. Could we add our example as a test to catch something like this in the future?

stuhlmueller:
Can you explain how that helps? The reason it's non-obvious is that Javascript passes primitive arguments by value, not reference:

```
> var foo = function(x){x+=1}
> var y = 10
> foo(y)
> y
10
```

Is the reason for this bug that this is no longer the case once we apply AD (turning the arguments into compound objects)? If so, how does the change address this?

longouyang:
Oh, you're right -- this PR doesn't do anything.

I went back to the original issue, #495, and found that I can't actually reproduce the bug.. I get: 

dist

longouyang:
You're right -- `binomialSample` doesn't actually mutate anything. I did some testing that suggests that the problem is something else, so I'll close this PR.

Issue #343, opened 2016-02-16
HMC - Transform constrained variables


null-a:
The current HMC implementation handles constrained variables using the scheme described in "MCMC using Hamiltonian dynamics". (Section 5.1, p36.) I'm not convinced that this will always be correct. We might want to transform each constrained variable into an unbounded space instead. (Previous [discussion](https://github.com/probmods/webppl/issues/81#issuecomment-154168113).)

null-a:
This would also allow us to support the Dirichlet distribution in HMC. I believe it would also address [this concern](https://github.com/probmods/webppl/issues/275#issuecomment-161664265).

PR #503, opened 2016-06-20
HMC for multivariate distributions

null-a:
This updates HMC to work with (unconstrained/unbounded) multivariate distributions. (From our current set of distributions, these are just the various flavors of multivariate Gaussians.)

Our current scheme for handing variables with bounded support doesn't straight-forwardly handle constraints, so we'll probably want to tackle #343 to make this work with the Dirichlet.

Attempting to use HMC with the Dirichlet throws an informative error.


stuhlmueller:
Neat!

Transforming Dirichlet to unbounded space sounds good, too.

PR #512, opened 2016-06-22
Some work on guides & Optimize docs

### LINKED ISSUE #507: Add background to `Guides` page in docs

null-a:
Here's an attempt to improve some of the documentation on guides and `Optimize`. There's still room for improvement of course, but I think this is a step in the right direction.

I've also added a note about using guides to specify importance distributions in SMC. As part of that I've tried to add a sentence to suggest *why* you might want to do such a thing, but I'm worried it's a rather vague/imprecise -- see what you think.

Closes #507.

### DIFF
--- a/docs/guides.rst
+++ b/docs/guides.rst
@@ -18,7 +18,7 @@ For example::
   sample(Cauchy(params));
 
 A number of inference strategies make use of an auxiliary distribution
-which we call a *guide distribution*. There are specified like so::
+which we call a *guide distribution*. They are specified like so::
 
   sample(dist, {guide: guideDist});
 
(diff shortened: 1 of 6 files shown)


docs/guides.rst
```diff
@@ -3,38 +3,27 @@
...
-   If ``sd`` is omitted the initial value is ``mean``. If ``mean`` is
-   omitted it defaults to zero.
+For example::
 
-   Example::
+  sample(Cauchy(params));
 
-     scalarParam(0, 1)
+A number of inference strategies make use of an auxiliary distribution
+which we call a *guide distribution*. There are specified like so::
```
stuhlmueller: "There" - typo?

stuhlmueller:
Looks great!

docs/guides.rst
null-a: Thanks!

PR #516, opened 2016-06-23
Some header function documentation

null-a:
Partially addresses #324.


ngoodman:
merging this as-is so that others can extend when needed...

PR #517, opened 2016-06-23
Remove bold from inline code within js:function.

null-a:
We document functions using the js:function markup from Sphinx's
Javascript domain. Inline code within these blocks currently uses font
weight bold. This looks bad, and it differs from the formatting used
outside of these blocks. This fixes that.

Before:
before

After:
after

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ node_modules
 npm-debug.log
 bundle/
 docs/_build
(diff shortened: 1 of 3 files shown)


null-a:
This may have broken the docs:

http://webppl.readthedocs.io/en/dev/

Investigating now...

PR #526, opened 2016-06-24
Remove bold from inline code within js:function.

null-a:
This implements the same thing as #517. I've tested this on RTD, looks like it works this time.


null-a:
CI failure was `uniformDrift`.

PR #527, opened 2016-06-24
Remove unused math functions

null-a:
I planned to move `trigamma` into `special.js`, but when I looked it's not currently in use. @longouyang: any reason to keep this function around?

We don't appear to use `binomialG` anymore either.

We retain a copy of `digamma` in `special.js`.

### DIFF
--- a/src/dists.ad.js
+++ b/src/dists.ad.js
@@ -869,18 +869,6 @@ function betaSample(a, b) {
 }
 
 
-function binomialG(x) {
-  'use ad';
-  if (x === 0) {
-    return 1;
-  }
-  if (x === 1) {
-    return 0;
-  }
-  var d = 1 - x;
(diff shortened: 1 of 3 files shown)


longouyang:
In #269, we planned to rewrite the Beta and Dirichlet samplers in terms of expGamma (I really should get around to this -- next week, hopefully).

We won't be exposing expGamma to end users but we should write tests for it anyway -- `digamma` and `trigamma` are useful for computing the expected moments of that distribution.

null-a:
> We won't be exposing expGamma to end users but we should write tests for it anyway -- digamma and trigamma are useful for computing the expected moments of that distribution.

Ok, cool. I close this and submit a new pr the retains `trigamma` but moves it into `special.js`.

PR #532, opened 2016-06-27
Make webppl installable on Windows

longouyang:
I tested on Windows 7 using a [free VM that Microsoft provides](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/). Installing this:

```sh
npm install -g longouyang/webppl#d824cce
```
worked.

### DIFF
--- a/scripts/adify
+++ b/scripts/adify
@@ -28,7 +28,7 @@ function adRequirePath(srcFilename) {
   var dir = path.dirname(srcFilename);
   var adFileName = path.join(__dirname, '..', 'src', 'ad.js');
   var adPath = path.normalize(path.dirname(path.relative(dir, adFileName)));
-  return adPath + path.sep + 'ad.js';
+  return adPath + '/' + 'ad.js';
 }
 
 // Don't match text editor temporary files. e.g. src/.#dists.ad.js
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,3 +1,1 @@
 NOTE: `helpers/helpers.js` cannot be moved to the top-level tests directory. For Windows compatibility, we alias `npm test` to `nodeunit tests`, which runs all the .js files in the top-level directory; helpers.js is not a valid nodeunit file, so this would cause tests to fail.
-
-TODO: document testing infrastructure more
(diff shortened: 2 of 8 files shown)


null-a:
Nice! Would you be able to try loading a package from `~/.webppl`? It would be neat to know whether [this](https://github.com/probmods/webppl/blob/234ea4594b1513686a145ab111debe67b2c6827f/src/pkg.js#L17) works.

I wonder if we need a note to explain what's going on with the `helpers/helper.js` directory structure? (I can imagine this getting undone as part of a 'clean-up' in the future.)

scripts/adify
```diff
@@ -28,7 +28,7 @@ function adRequirePath(srcFilename) {
   var dir = path.dirname(srcFilename);
   var adFileName = path.join(__dirname, '..', 'src', 'ad.js');
   var adPath = path.normalize(path.dirname(path.relative(dir, adFileName)));
-  return adPath + path.sep + 'ad.js';
+  return adPath + '/' + 'ad.js';
 }
```
null-a: It seems odd to me that this worked. I used `path.sep` precisely because I hoped it would work cross-platform. Any idea what's happening?

longouyang: Node require paths must use `/` but `path.sep` on Windows is `\`

null-a: Aha, thanks!

longouyang:
> Nice! Would you be able to try loading a package from ~/.webppl? It would be neat to know whether this works.

Yes - this worked!

> I wonder if we need a note to explain what's going on with the helpers/helper.js directory structure? (I can imagine this getting undone as part of a 'clean-up' in the future.)

I added a comment in `helpers.js` explaining this and also stubbed a tests README with this info as well.

tests/README.md
```diff
@@ -0,0 +1,3 @@
+NOTE: `helpers/helpers.js` cannot be moved to the top-level tests directory. For Windows compatibility, we alias `npm test` to `nodeunit tests`, which runs all the .js files in the top-level directory; helpers.js is not a valid nodeunit file, so this would cause tests to fail.
+
+TODO: document testing infrastructure more
```
stuhlmueller: Move this remark to an issue (unless you're planning to address this as part of this PR)?

longouyang: You mean the TODO part?

stuhlmueller: Yes

longouyang: Okay, done

Issue #515, opened 2016-06-22
Add sampler tests for Beta


longouyang:

longouyang:
Closed by #536

PR #536, opened 2016-06-28
Test Beta sampler

### LINKED ISSUE #269: beta sometimes returns NaN

longouyang:
This would close #269 and #515

### DIFF
--- a/src/dists.ad.js
+++ b/src/dists.ad.js
@@ -858,14 +858,17 @@ var Beta = makeDistributionType({
 });
 
 function betaSample(a, b) {
-  var x = gammaSample(a, 1);
-  var y = x / (x + gammaSample(b, 1));
-  if (y === 0) {
-    y = Number.MIN_VALUE;
-  } else if (y === 1) {
-    y = 1 - Number.EPSILON / 2;
+  var log_x = expGammaSample(a, 1);
+  var log_y = expGammaSample(b, 1);
+  var v = 1 / (1 + Math.exp(log_y - log_x));
+  if (v === 0) {
+    util.warn('beta sample underflow, rounded to nearest representable support value');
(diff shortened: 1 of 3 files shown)


longouyang:
(test failure was unrelated - HMC mixed3)

PR #540, opened 2016-07-03
Test Gaussian sampler

longouyang:


tests/test-data/sampler/gaussian.js
```diff
@@ -0,0 +1,57 @@
+var dists = require('../../../src/dists');
+var util = require('../../../src/util');
+var statistics = require('../../../src/math/statistics');
+
+var doubleFactorial = function(x) {
```
stuhlmueller: Move to `src/math/special.js`?

longouyang:
I can't see this being used anywhere else, though. If that's the case, should we still move it to `special.js`?

stuhlmueller:
Either way works. I would probably move it as a matter of principle, but I'm happy to merge with it in place as well.

PR #547, opened 2016-07-11
In browser, don't console.log version or package names

longouyang:
As of 0.9.2, editor now intercepts `console` methods, so let's not `console.log` the webppl version once it loads.

(We still have access to this information because the `webppl` object has a `version` property)


stuhlmueller:
Is there still a way to access package versions in the browser? If not, we should make one available before we remove the corresponding `console.log` statements.

stuhlmueller:
Thanks!

PR #553, opened 2016-07-13
Reset env before running code in the browser.

### LINKED ISSUE #550: Reset env.coroutine before each run/after exception in browser
Not doing this causes problems: probmods/ppaml2016#1, probmods/ppaml2016#2 I think.

null-a:
Without this, an exception occurring in one call to `run` leaves `env` in an unpredictable state for subsequent calls.

Ideally this would be handled entirely internally, but at present the editor calls `compile` and `eval` when running code, so for now it will also have to call `webppl.resetEnv`. The work I've been doing on #314 should allow us to remedy this. It will also make it much easier for me to write a test for this.

Closes #550.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -33,7 +33,7 @@ function run(code, k, options) {
     options = {};
   }
   var optionsExtended = _.extend({bundles: load()}, options);
-
+  webppl.resetEnv();
   return webppl.run(code, k, optionsExtended);
 }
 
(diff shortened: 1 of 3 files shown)


stuhlmueller:
This looks fine as a temporary measure. 

Would calling `reset` at the beginning of `header.wppl` be a way to accomplish this internally, or are there problems with this alternative?

null-a:
> Would calling reset at the beginning of header.wppl be a way to accomplish this internally, or are there problems with this alternative?

I think was trying to to something more local than this, but on reflection this seems like a good idea to me. I'll open a fresh PR.

PR #558, opened 2016-07-13
Add `mapData` primitive + mini batch support for VI

null-a:
This is a cleaned up and simplified version of the `mapData` implementation that we choose not to merge along with the rest of the VI machinery.

The interface may need to change at some point, but by merging it now we can include it in the summer school material.

I'll create a follow-up issue to remind me to write some docs at some point.


src/headerUtils.js
```diff
@@ -112,6 +112,69 @@ module.exports = function(env) {
...
+      throw new Error('mapData: Invalid batchSize.');
+    }
+
+    // Query the coroutine to determine the subset of the data to map
+    // over.
+    var ix = env.coroutine.mapDataFetch ?
+        env.coroutine.mapDataFetch(data, batchSize, a) :
+        // The empty array stands for all indices, in order. i.e.
+        // `_.range(data.length)`
+        [];
```
stuhlmueller: I wonder if we should use another value as a placeholder for all indices. As it stands, there is no way for the coroutine to actually indicate the empty array (which wouldn't be an entirely implausible thing to do, e.g. to initialize inference without conditioning on the data). Maybe use `null` instead?

src/inference/elbo.ad.js
```diff
@@ -208,6 +209,54 @@ module.exports = function(env) {
...
 
+    mapDataFetch: function(data, batchSize, address) {
+
+      var ix;
+      if (batchSize === data.length) {
+        // Use all the data, in order.
+        ix = [];
+      } else {
+        ix = _.times(batchSize, function() {
+          return Math.floor(util.random() * data.length);
```
stuhlmueller: Would it be better to sample these indices without replacement? (See e.g. [here](http://arxiv.org/pdf/1202.4184v1.pdf).)

stuhlmueller:
If we're going to use it in the summer school, let's also add it to the docs? We can create an issue if you don't want it to be part of this PR.

On the development side, it would also be nice to have a brief comment somewhere that explains the interface for `mapDataFetch` and `mapDataFinal`. 

Add a test case that includes a factor in the `mapData` `obsFun`, and also run the `mapData` test for other inference algorithms such as MCMC, just to make sure that the behavior is identical to regular `map` in those cases?

src/inference/elbo.ad.js
null-a: Thanks for the reference, I'd not come across that result before. I went with "with replacement" for simplicity, and for the data set I've been using it with (mnist) both strategies amount to about the same thing. Having "without replacement" may well be worth while (and is on my list of things to look at), but in the interests of getting to some of the other things we want for 0.9 can we merge this pr without it?

null-a:
I've done everything other than the add support for without replacement. (Extra comment above.)

Once this is merged I'll create an issue to remind me to add this to the docs.

null-a:
CI failure was the Beta sampler.

stuhlmueller:
Looks good, thanks!

Issue #444, opened 2016-05-20
Docs: making documentation even more beginner/user-friendly


justinek:
Comments on [webPPL documentation](http://docs.webppl.org/en/master/), to be taken with a grain of salt!

### Quick start

- Maybe the very first bullet should be, "If you haven't yet, install nodejs"?

### Usage

- In general, it might be helpful to accompany each line of code with a short note about when it would be useful to perform it. For example, when might one want to set a seed for the random number generator? When would it be useful to compile to Javascript?

### Inference

- How is EnumerateLikelyFirst different from the default heuristic of "most probable first"? 
- It would be helpful to include more details on the different search strategies, a well as some rough guidelines on when each strategy should be used

### WebPPL packages
- It might be useful to include a brief description/reminder of the differences between WebPPL  and regular JS, to make clear when packages are useful to work around certain limitations of WebPPL (e.g. immutability)
- Maybe include a note on when macros are useful?
- Under Javascript functions and libraries, there's a gentle warning that "Note that packages must export functions as properties of an object. Exporting functions directly will not work as expected." Maybe it would be useful to give an example of what works and what doesn't to make the correct usage even more clear?
- In the example for under "Additional header files," it was not immediately clear to me what k is.
- It might be helpful to include a bit more documentation for the [package template](https://github.com/probmods/webppl-package-template/blob/master/README.md), with information about what the package does and which parts should be changed to implement user's own package

chrisranderson:
I have a draft for a new quick start page here: https://github.com/chrisranderson/webppl/blob/quick-start-rewrite/docs/quickstart.rst

@null-a, it's ready-ish but I'd like some input first. Is it better for me to just submit a PR or comment like this? Also, what do you think about the rewrite?

longouyang:
That looks great to me. Could you also mention that installation requires git? (that is certainly true on Windows and I think it might be true of OS X / Linux too)

A typo: "help **l**in learning the syntax"

chrisranderson:
How is git required?

longouyang:
Some of our npm dependencies are github repos (e.g., `probmods/sweet.js#browserify`) -- when I tried installing on Windows without git, it choked when trying to set these up./

Also, I believe the webppl package system depends a little on running the git binary.

PR #559, opened 2016-07-14
Quick start rewrite

chrisranderson:
For #444.

~~My only question: does the `-g` flag for npm installs require sudo?~~ Checked. According to https://docs.npmjs.com/getting-started/installing-npm-packages-globally it's discouraged.


docs/quickstart.rst
```diff
@@ -1,14 +1,48 @@
 Quick start
 ===========
 
-Install using `nodejs <http://nodejs.org>`_::
+Installation 
+------------
+
+First, install `git <https://git-scm.com/downloads>`_. 
+
+Second, install `Node.js <http://nodejs.org>`_. Webppl is written in JavaScript, and requires Node to run. After it is installed, you can use npm (**n**\ ode **p**\ ackage **m**\ anager) to install webppl::
```
null-a: I *think* "WebPPL" is used more often in the docs at the moment, so I'd use that for now. It also means we can punt on whether or not "webppl" should be "Webppl" at the beginning of a sentence...

docs/quickstart.rst
```diff
@@ -1,14 +1,48 @@
...
+    return phrase
+
+Run it with this command::
+
+    webppl test.wppl
+
+Learning
+--------
+
+If you're new to probabilistic programming, `Probabilistic Models of Cognition <https://probmods.org/>`_ is a great place to start learning the paradigm. It uses Church for demonstrations, but the principles will apply when you are using webppl.
```
null-a: We maybe shouldn't assume people will know what Church is. Maybe we just say "uses the language Church..." or something.

docs/quickstart.rst
```diff
@@ -1,14 +1,48 @@
...
+        return flip(.5) ? "Hello" : "Howdy"
+    }
+    
+    var audience = function () {
+        return flip(.5) ? "World" : "Universe"
+    }
+    
+    var phrase = greeting() + ", " + audience() + "!"
+    
+    return phrase
```
null-a: As you probably know, we automatically `return` the last expression in the program. I wonder if we should introduce that idea here by dropping the `phrase` variable?

null-a:
Looks good! I had a quick pass through and added a few comments inline.

docs/quickstart.rst
chrisranderson: Read my mind! I was wondering...

I'll change that.

docs/quickstart.rst
chrisranderson: I would count that as distracting, but I'm okay with changing it. It is kinda weird coming from JS land.

chrisranderson: On the other hand, it would be nice to let them know "this isn't just JS".

docs/quickstart.rst
chrisranderson: What if I add a link to http://projects.csail.mit.edu/church/wiki/Church?

stuhlmueller:
Yes, looks good. Ready to merge once `[W]ebppl` has been replaced with `WebPPL` throughout.

chrisranderson:
Eh, sorry. I'm getting confused with all of my branches it looks like.

stuhlmueller:
I assume you've seen it already, so just in case: there is a [leftover bit](https://github.com/probmods/webppl/pull/559/files#diff-9d8a6f5b32d4a70a56734fd0eb50dd5bR41) from the merge.

chrisranderson:
@stuhlmueller  my git incompetence shows through. Should be good now.

stuhlmueller:
Thanks! There's still a small issue: `<<<<<<< HEAD` shouldn't be in `quickstart.rst`.

chrisranderson:
Sorry :/ fixed.

stuhlmueller:
Thanks!

PR #565, opened 2016-07-15
Reset env at start of every program.

### LINKED ISSUE #550: Reset env.coroutine before each run/after exception in browser

null-a:
Supersedes #553.
Closes #550.

This is important when `run` is called multiple times in the same process. Without this, an exception occurring in one call to `run` leaves `env` in an unpredictable state for subsequent calls.

### DIFF
--- a/src/header.js
+++ b/src/header.js
@@ -53,7 +53,7 @@ module.exports = function(env) {
 
   // Inference interface
 
-  env.coroutine = {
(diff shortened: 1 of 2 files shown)


null-a:
CI failure was Beta sampler.

Issue #308, opened 2016-01-29
Support drift proposals for more ERPs and document how to use them


stuhlmueller:
Also, support fine-tuning of drift parameters (#231).

stuhlmueller:
Also, add tests. We should have noticed the wonky concentration parameter for the Dirichlet drift ERP (#309).

null-a:
>  We should have noticed the wonky concentration parameter for the Dirichlet drift ERP 

Please could someone explain what the idea is behind the dirichlet drift kernel? Proposing from a dirichlet with params `alpha * 10` seems it *might* be a useful thing to do if the prior is something like a symmetric dirichlet with `alpha=0.1`, but if the prior has parameters `alpha=1` then aren't we proposing from a dirichlet with its mode in the middle of the simplex? The thing I'm most surprised by is the fact that the value of the previous sample is not used at all, wouldn't it be useful to propose from a dirichlet with it's mode at the value of the previous sample as we do with the gaussian?

stuhlmueller:
> The thing I'm most surprised by is the fact that the value of the previous sample is not used at all, wouldn't it be useful to propose from a dirichlet with it's mode at the value of the previous sample as we do with the gaussian?

Wait, that's not what we are doing? You're right, of course, `params` in `dirichletProposalParams` should be `prevVal`.

ngoodman:
Btw in general if a Dist is parameterized by mean (and other params) we want the drift proposal to be sampled from the same family but mean equal to the current value of the choice (and other params set based on a proposal width parameter). For Dists that don't naturally have mean as a param, we may want to use a different Dist for proposals (eg for Uniform we'd probably want a scaled Beta).

ngoodman:
Also, we may want to unify the way we declare drift proposals and importance distributions, which are really very very similar.

null-a:
> we may want to unify the way we declare drift proposals and importance distributions

In my mind, I've been imagining we'll tackle this after the guide stuff/VI is merged. BTW, there's issue #119 for this.

longouyang:
FYI, the uniform drift I implemented in #418 uses a uniform proposal; we should switch this to scaled Beta as Noah suggests. (this might also help with the test failures reported in #430)

PR #568, opened 2016-07-15
Allow drift kernels to be specified in user code.

### LINKED ISSUE #119: Unify treatment of importance distributions and custom MH proposers

### LINKED ISSUE #231: Remove hard-coded drift kernel parameters
The [variance](https://github.com/probmods/webppl/blob/5c1212762cfa0184d39b652f4f8452451c20628c/src/erp.js#L618) and [concentration](https://github.com/probmods/webppl/blob/5c1212762cfa0184d39b652f4f8452451c20628c/src/erp.js#L623) parameters of drift kernels are currently hard-coded. Presumably it should be possible to specify these from within a program or perhaps have them tuned automatically.

null-a:
After this change we no longer have to add new distributions for each distribution/drift kernel pair we'd like to use. Instead, we can write things like:

```js
sample(Cauchy(params), {
  driftKernel: function(prevVal) {
    return Gaussian({mu: prevVal, sigma: width});
  }
});
```

A lot of the complexity in the implementation comes from having to call the cps drift kernel functions from hand written cps inference code, but it's not too bad.

I've added helpers to the header that mimic the behavior of the old drift distributions.

Closes #119. Closes #231. Partially addresses #308 (added docs).

@dritchie has already [looked over](https://github.com/probmods/webppl/issues/119#issuecomment-232826013) the incremental MH bits.

### DIFF
--- a/docs/distributions.rst
+++ b/docs/distributions.rst
@@ -68,12 +68,6 @@ Distributions
 
   `Wikipedia entry <https://en.wikipedia.org/wiki/Dirichlet_distribution>`__
 
-.. js:function:: DirichletDrift({alpha: ...})
-
-  * alpha: vector of concentration parameters *(>0)*
-
-  Drift version of Dirichlet. Drift kernels are used to narrow search during inference. Currently, the parameters guiding this narrowing are hard-coded.
-
 .. js:function:: Discrete({ps: ...})
 
   * ps: array or vector of probabilities *(in [0,1])*
@@ -107,13 +101,6 @@ Distributions
   Distribution over reals.
 
   `Wikipedia entry <https://en.wikipedia.org/wiki/Normal_distribution>`__
-
-.. js:function:: GaussianDrift({mu: ..., sigma: ...})
-
-  * mu: mean (real)
-  * sigma: standard deviation (real) *(>0)*
-
-  Drift version of Gaussian. Drift kernels are used to narrow search during inference. Currently, the parameters guiding this narrowing are hard-coded.
 
 .. js:function:: LogisticNormal({mu: ..., sigma: ...})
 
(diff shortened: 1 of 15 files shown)


stuhlmueller:
Looks really good! Thanks for the detailed documentation as well!

In `getProposalDist`, would it make sense to print a warning message if the sampled value is outside of the support of the prior distribution? The intention would be to help people notice if they write flawed or inefficient drift samplers.

stuhlmueller:
Thanks for adding the zero-probability drift proposal warning. Are we sure that the only way to end up at the warning (where it's currently located) is due to out-of-support drift proposals? What if an upstream proposal changed the parameters to a downstream distribution that happens to have a drift proposer associated with it, and changed it in such a way that the score for the current value of that distribution becomes zero? Would we not show the same warning then?

null-a:
> Would we not show the same warning then?

I think you're right, good catch! I've pushed a fix.

Issue #489, opened 2016-06-10
Make returnify error more readable for models where the last statement is not an expression


stuhlmueller:
This is a subset of #331

chrisranderson:
I'm guessing this is more than changing the string on this line: https://github.com/probmods/webppl/blob/dev/src/syntax.js#L117. If not, I can take a whack at it. :)

null-a:
> I'm guessing this is more than changing the string on this line:

Perhaps, but only a *little* more...

The problematic error arises when trying to run a program like this:

```js
var a = 0;
```

Which produces:

```
{ type: 'VariableDeclaration',
  declarations: 
   [ { type: 'VariableDeclarator',
       id: [Object],
       init: [Object],
       range: [Object],
       loc: [Object] } ],
  kind: 'var',
  range: [ 0, 9 ],
  loc: 
   { start: { line: 1, column: 0 },
     end: { line: 1, column: 9 },
     source: 'test.wppl' } }
returnify
path/to/webppl/webppl:56
      throw error;
      ^

Error: returnify
    at path/to/webppl/src/syntax.js:20:11
    ... more stack trace ...
```

That error is generated by the function returned from `fail` on the line you mentioned.

I imagine that making following changes would probably produce acceptable results:

1. Don't show the ast node.
2. Have a more descriptive message, explain how to fix the problem.
3. Perhaps loose the JS stack trace which doesn't seem useful for users. We could `throw` a string rather than an `Error` to achieve this.

So, perhaps not *too* much more than just changing the string?

chrisranderson:
How about this: if the user doesn't return something, we return the empty string for them. Maybe they just want to check syntax and don't care to see what the program returns, or they print the result themselves. Maybe they had a return value, but didn't want it to print out so they commented it out. Instead of changing the error message, we just drop the requirement that the program have a return value. Thoughts? I think that beats `Error: programs must end with a return value.`

null-a:
Yeah, this crossed my mind, and I think it's worth considering. If there are reasons why it wouldn't work out I don't know what they are.

This *might* be a bigger change than just changing the error, but I'm not sure what's involved without looking closely. If we're end up short of time, just changing the error would be still be a big improvement and worth having for v0.9.

> we return the empty string for them

I think I would use `undefined` instead, and then consider separately the question of whether to suppress output when the program evaluates to `undefined`. One reason is that functions that end with a statement rather than an expression already return `undefined`.

null-a:
Closed by #587.

PR #574, opened 2016-07-19
Change error for missing return value.

chrisranderson:
For #489.

- Does this change errors for other problems?
- Is the way I handled `fail()` hacky?
- The message line is too long to pass `grunt` and there was an unrelated error, so I'm not sure what to do about that:

```
Testing test-inference.js......................................MultivariateBernoulli({ ps: 
   Tensor {
     dims: [ 2, 1 ],
     length: 2,
     data: Float64Array { '0': 0.5, '1': 0.1 } } })
Exception: Error: Enumerate can only be used with distributions that have finite support.
F..........................................................................................................................................................................................................
>> Enumerate - multivariateBernoulli
>> Error: Enumerate can only be used with distributions that have finite support.
>> at getSupport (src/inference/enumerate.js:44:19)
>> at Enumerate.sample (src/inference/enumerate.js:54:23)
>> at env.sample (src/header.js:82:26)
>> at eval (eval at <anonymous> (src/main.js:220:10), <anonymous>:75:44)
>> at /usr/local/lib/node_modules/webppl/src/util.js:51:13
>> at /usr/local/lib/node_modules/webppl/src/util.js:14:14
>> at eval (eval at <anonymous> (src/main.js:220:10), <anonymous>:5:13)
>> at /usr/local/lib/node_modules/webppl/src/main.js:220:48
>> at Object.timeif (src/util.js:283:37)
>> at Object.run (src/main.js:219:8)
```

### DIFF
--- a/src/syntax.js
+++ b/src/syntax.js
@@ -13,12 +13,18 @@ function makeGenvar() {
   };
 }
 
-function fail(message, node) {
-  return function() {
-    console.log(node);
-    console.log(message);
-    throw new Error(message);
-  };
+function fail(message, node, simpleMessage) {
+  if (simpleMessage === true) {
+    return function() {
+      throw message;
+    }
+  } else {
+    return function() {
+      console.log(node);
+      console.log(message);
+      throw new Error(message);
+    };
+  }
 }
 
 // a clause matches a node type and calls a destructor with constituents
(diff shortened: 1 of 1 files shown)


null-a:
> Does this change errors for other problems?
Not that I know of. What do you have in mind?

> Is the way I handled fail() hacky?

How about just replacing the call to `fail` with an anonymous function?

```js
match(nodes, [
// ...
], function() {
  throw 'message';
});
```

> The message line is too long to pass grunt

So:

```js
var a = 'first bit' +
  'rest';
```
... or similar?

> and there was an unrelated error, so I'm not sure what to do about that:

The error you mentioned happened locally, right? I think doing `grunt build` will fix it.

One other thing. I think we *might* be better saying that a program needs to end with an "expression", rather than a "return value". Neither is perfect, but the first seems a little more precise to me. I'm also not sure about mentioning the `return` keyword. I don't think it's used in any of our examples, and even though it works it feels a bit strange to be using it outside of a function to me.

null-a:
@chrisranderson Thanks for your work on this so far, I'll take over from here.

Issue #567, opened 2016-07-15
Potentially confusing error message: "Assignment is allowed only..."


chrisranderson:
With the program:

```javascript
a = [1, 2, 3]
b = [4, 5, 6]
append(a, b)
```

I get the error message "Assignment is allowed only to fields of globalStore."

If that error message is true, then this would be illegal:

```javascript
var a = [1, 2, 3]
```

It looks like it's because it's making a global assignment since `var` is missing.  Maybe instead the message could be something like "Assignment in the global scope can only be to the globalStore variable."

null-a:
The error is correct, `var a = [1, 2, 3]` is a variable declaration rather than an assignment. That's not to say the error can't be improved! Perhaps we cover the case you highlight by adding something like "Perhaps you meant `var a = foo`" when we catch `a = foo`

longouyang:
Closed by #576

Issue #317, opened 2016-02-10
Replace `false==true` error with readable messages


stuhlmueller:
Check that all assertions have error messages, as in PR #315.

null-a:
I tend to use `assert` to describe things I believe should always hold and `throw` to signal that we've run into a situation that was anticipated but from which there's no sensible way to proceed. In principle we could have a build step which removes the assertions when not in debug mode. Another benefit is that the errors produced by `throw` are a little easier to read than those produced by `assert`. Does this sound reasonable? If so is it worth doing everywhere?

This doesn't change the fact that it would be nice to have readable errors, but it might mean that some `assert`s should be turned into `throw`s.

stuhlmueller:
Ok, let's go with that approach.

PR #576, opened 2016-07-20
More informative syntax error for assignment #567

longouyang:
Better errors, in particular:

```js
foo = 9         // Did you mean var foo = ?
foo.bar = 9     // You tried to assign to a field of foo but you can only assign to fields of globalStore
globalStore = 9 // You must assign to a field of globalStore, not globalStore itself
```

### DIFF
--- a/src/transforms/cps.js
+++ b/src/transforms/cps.js
@@ -55,9 +55,26 @@ function atomize(node, metaK) {
       });
     }),
     clause(Syntax.AssignmentExpression, function(left, right) {
+      var errorMessage = function() {
+        var line = left.loc.start.line;
+        var msg;
+        if (left.type === Syntax.MemberExpression) {
+          if (left.object.name !== 'globalStore') {
+            msg = 'You tried to assign to a field of ' + left.object.name +
+                ', but you can only assign to fields of globalStore';
+          }
+        } else {
+          if (left.name == 'globalStore') {
+            msg = 'You must assign to a field of globalStore, not globalStore itself'
+          } else {
+            msg = 'Did you mean var ' + left.name + ' = ?'
+          }
+        }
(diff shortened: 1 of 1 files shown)


null-a:
Nice!

This is a definite improvement, and I think we'd be fine to merge it.

Some nit-picky comments that crossed my while looking through it:

1. I think we should be using strict equality checks everywhere. (Unless we *want* the funny coercions of the sloppy version.)
2. At some point I think we should switch this from an assertion to a throw. See my comment in #317 for my thinking.
3. It might be nice if we only did the work of building the error message when we need it. We see very few assignment expressions so this isn't a big deal, and it will likely happen when we do 2 anyway.

longouyang:
Okay, now this does strict equality, throw instead of assert, and conditional message building. The test failure was unrelated - HMC constrainedSum.

null-a:
>  now this does strict equality

Thanks. btw, `!==` was fine, only `==` and `!=` are best avoided. (Happy to explain.)

longouyang:
Yeah, I'm curious -- why require `a === b` but not `(!(a === b))`?

null-a:
I don't think I understand the question, so perhaps I've confused things here! Anyway...

The standard argument for preferring `===` and `!==` to `==` and `!=` is that the former pretty much matches everyone's intuitions about how testing for equality should work, whereas the latter does type coercion which makes it harder to reason about and more likely to cause bugs.

I got this from Crockford's "JavaScript: The Good Parts" IIRC, though it already happened in webppl before I made any changes.

longouyang:
Ah - I knew that `===` cared about types but I didn't know that `!==` did too (that's why I thought I had to do `(!(a === b))`)

PR #581, opened 2016-07-21
Don't inc. accepted count when no proposal made.

### LINKED ISSUE #575: HMC gives unintuitive acceptance ratios when there are no discrete choices
When `HMC` is used as kernel for a program that has no discrete random choices, it still alternates between an HMC step and a vanilla MH step. The MH step immediately accepts, since there are no discrete choices. It records this as an acceptance in `trace.info`, which artificially inflates the acceptance ratio.

One possible fix might be to change [this](https://github.com/probmods/webppl/blob/dev/src/inference/mhkernel.js#L40) code so that `MHKernel` runs which finish immediately due to no proposable choices do not write anything to `trace.info`.

I've marked this as a 0.9 milestone because I use acceptance ratios in one of the summer school examples illustrating HMC, and it would be nice if those values were accurate so students can build the right intuitions / mental models.

null-a:
After this change both the MH and HMC kernels immediately return the old trace without incrementing the acceptance ratio counts when there are no random choices to propose to.

Because it's now possible for the number of proposals made to be zero, we check for this when computing the acceptance ratio to avoid division by zero.

Closes #575.

### DIFF
--- a/src/inference/hmckernel.js
+++ b/src/inference/hmckernel.js
@@ -100,6 +100,15 @@ module.exports = function(env) {
 
   HMCKernel.prototype.run = function() {
 
+    // Immediately return from coroutine if there are no continuous
+    // random choices to propose to.
+    var numChoices = this.oldTrace.choices.reduce(function(acc, c) {
+      return acc + (c.dist.isContinuous ? 1 : 0);
+    }, 0);
+    if (numChoices === 0) {
+      return this.continue(this.oldTrace);
+    }
+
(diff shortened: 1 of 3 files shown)


stuhlmueller:
Looks good. I wonder if we should print a warning somewhere if MH is run without choices to propose to, or HMC without continuous choices.

null-a:
Not sure - do you have a scenario in mind where doing so would help someone spot a problem that would otherwise escape them?

PR #587, opened 2016-07-25
Append "undefined" to programs without a final expression

null-a:
In #489 @chrisranderson suggested we return some default value rather than throw an error. This seems worth exploring to me, so this change adds the line `undefined` to the end of any program that would previously have throw the returnify error. I'm happy to close this submit a PR that just changes the current error message if this is risky/controversial, etc.

Things I like about it:

1. This feels pretty natural to me. I don't think seeing `undefined` in the output will block users' progress.
2. I occasionally find myself with a program that doesn't end with an expression that I do want to run. In such cases I tend to write `'done'` at the end of the program, which is a tiny inconvenience that this would remove.

One concern is that users may not realise that we print the value of last expression and use `display` instead:

```js
var a = 0;
display(a);
```

The hope is that because the above program will still have `undefined` as the last line of the output, that this will be enough to help people figure out what's happening.

If not, perhaps we could print a warning but still run the program. (Something like: "Note: This program will evaluate to undefined because the last statement is not an expression.")

### DIFF
--- a/src/syntax.js
+++ b/src/syntax.js
@@ -95,7 +95,7 @@ function returnify(nodes) {
     return nodes;
   }
   else {
-    nodes[nodes.length - 1] = match(nodes[nodes.length - 1], [
+    var returnNode = match(nodes[nodes.length - 1], [
       clause(Syntax.BlockStatement, function(body) {
         return build.blockStatement(returnify(body));
       }),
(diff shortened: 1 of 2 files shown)


ngoodman:
i like this change

stuhlmueller:
Nice! That's an elegant way of addressing the `returnify` issue. As Paul points out [here](https://github.com/probmods/webppl/issues/489#issuecomment-233741487), our functions already behave this way, so this makes a lot of sense.

PR #588, opened 2016-07-25
Allow params to be created outside of coroutines.

null-a:
Addresses probmods/ppaml2016#4.


stuhlmueller:
Is there a reason not to set the entry address for the default coroutine to the empty address?

null-a:
That would be fine, good idea. I'll take care of it.

Issue #326, opened 2016-02-10
Review installation/basic usage docs, check they are clear


null-a:

erindb:
Is this talking about [docs.webppl.org](http://docs.webppl.org/)?

null-a:
Yes, I think so. I guess the 'Quick start' and 'Usage' pages?

erindb:
OK. In the 'Usage' page, this is kind of a rabbit hole for me:

> To use webppl in web pages, include the browser bundle:
> 
> `<script src="webppl.js"></script>`
> `<script>webppl.run(...)</script>`

It's not obvious to me how to interpret the information in the [browser bundle](http://docs.webppl.org/en/master/development/workflow.html#browser-version) link if I'm a user of webppl and I've installed it globally with npm.

erindb:
I wonder if it might be nice to provide a compiled `webppl.js` download somewhere and link to that on the usage page?

chrisranderson:
Thoughts on this:

#### quick start
- Would a sentence or two about node/npm would be helpful? I expect most folks reading this aren't very familiar. I've found installing node painful in the past - a link to instructions could be better than a link to the homepage.
- A small sample program could be nice to make sure things are working. Is there a page that runs you through your first webppl program?

#### usage
- maybe running things in the browser deserves its own page, linked to from usage or quick start. Then maybe merge usage and quick start. Wouldn't browser usage be rare?
- random seeds seems out of place for a getting-started type page. Could that be placed in the debugging page?

chrisranderson:
Quick start was changed in #559.

null-a:
> random seeds seems out of place for a getting-started type page. Could that be placed in the debugging page?

Your new quick start page seems to cover the "getting-started" case.

The "usage" page seems to me to be mostly a reference for the options available when running webppl on the command line. 

Including `--random-seed` in this seems ok. It's already mentioned in the `Development -> Testing` section.

If this section *is* about command line usage, then perhaps we should also mention the `--require` option and link to the "Packages" page?

> Wouldn't browser usage be rare?

I would expect so. Moving those browser bits into `Development -> Workflow -> Browser version` might not be ideal, but it seems like an improvement. @longouyang may have thoughts about this?

[...]

PR #590, opened 2016-07-26
Factor the quick start docs.

null-a:
* Mention the code box on webppl.org.
* Split local install out to its own page.

(#326)


null-a:
The CI failure was from a failed `npm install` which doesn't appear to be related to this change.

PR #594, opened 2016-07-27
Improve VI error messages.

null-a:
This makes a couple minor tweaks to VI related error messages that are aimed at improving their usefulness a little.


stuhlmueller:
Thanks! This would definitely have saved Jason and me some time. 👍

PR #600, opened 2016-07-28
Add basic structure to docs.

null-a:
Here's my take on [the idea](https://groups.google.com/forum/#!msg/webppl-dev/MMtFEhWYge0/tdT0eZkfBAAJ) of adding some structure to the docs. It looks like this:

untitled

It's not perfect. For example, I've put "Debugging" under "Command Line", even though there's one pesky sentence on that page about the browser. Nevertheless, I think this is pretty reasonable, and improves things quite a bit. We can make further improvements as we think of them.


ngoodman:
i'm going to merge this to facilitate my planning for the ppaml-ss intro. we can go on tweaking as we like...

PR #627, opened 2016-08-16
More uniform treatment of entry addresses.

null-a:
As suggested in #588. This is exercised by the inference tests and by the deterministic tests in `guides.wppl`.


ngoodman:
looks good. we may want to visit the naming of params at some later point (to make it easier to use learned params in different locations).

null-a:
> looks good. we may want to visit the naming of params at some later point (to make it easier to use learned params in different locations).

Yeah, agreed. I have some ideas around this too.

PR #640, opened 2016-09-05
Better default guides for Uniform, Beta & Gamma.

### LINKED ISSUE #534: Mean-field with uniforms gives -Infinity guide log probability
E.g. this program:
```javascript
var model = function(_xs, _ys) {
  var xs = _xs || [1,2,3];
  var ys = _ys || [4,5,6];
  var m = uniform(-10, 10);
  var b = uniform(-3, 3);
  
  // TODO: sample sigma^2 from inverse-gamma
  var noiseDist = Gaussian({mu: 0, sigma: 1});
  
  var score = sum(map2(function(x,y) {
    var yhat = m*x + b, dev = y - yhat;
    return noiseDist.score(dev*dev)
  }, xs, ys))
  
  factor(score);
  
  return {m: m, b: b}
}

Optimize(model, {steps: 10})
```
produces this error:
```
ELBO: scoreDiff is not finite.
```
When `guide.independent` initializes the parameters for a mean-field approximate guide distribution, those parameters get value 0 (or `squishFn(0)`, if they are bounded), due to how Tensors initialize. This means that a `uniform` distribution has both its lower and upper bound parameters initialized to zero, which is a problem.

One way to get around this: change guide.independent so that it initializes parameters to the values used by the prior distribution. Are there any compelling reasons why this wouldn't be a good idea?

Also: I don't think it makes sense to use `uniform` as a guide distribution. The distribution can't be shaped, and the bounds are often used to specify inviolable range properties of the data, not tweakable parameters. I think the right thing to do here might be to make another special case (like `dirichletSpec` in `guide.js`) so that `uniform` actually uses a scaled `beta` as its guide distribution. This would require injecting additional transformation code, not just swapping out the dist, so that makes it a bit trickier to handle...

null-a:
This change adds two new distributions -- one of these is a distribution over a bounded real interval, the other is over the positive reals. Both are based on passing a sample from a Gaussian through a simple squishing function, so both support the reparameterization trick. These distributions are used as the default guides for Uniform, Beta and Gamma distributions, yielding something similar to [ADVI](http://arxiv.org/abs/1603.00788) I think.

The distribution over the positive reals uses the soft-plus squishing function since it generally seems to optimize better than `Math.exp`. It turns out ADVI also uses this transform, I think. (See page 15 of paper.) I'm not aware of a standard name for this distribution so I've left it out of the docs to avoid confusing people.

I've just realized that I've now added both `LogisticNormal` and `LogitNormal` which isn't consistent with the choice to use `Gaussian` rather than `Normal`. I'm not sure we should do anything about this though since you nearly always hear e.g. `LogNormal` rather than `LogGaussian`?

Closes #534.


stuhlmueller:
Looks good!

If we could make `a` and `b` of `LogitNormal` default to `0` and `1` respectively, we'd simplify the more standard use of this distribution. I think we don't have defaults for distribution parameters at the moment (right?). We can still merge the PR now, but let's add defaults later if we introduce the capability?

> I've just realized that I've now added both `LogisticNormal` and `LogitNormal` which isn't consistent with the choice to use `Gaussian` rather than `Normal`. I'm not sure we should do anything about this though since you nearly always hear e.g. `LogNormal` rather than `LogGaussian`?

I'd go with the more common names as well, using Wikipedia as arbitrator. However, that would imply using `Normal` instead of `Gaussian`. Maybe we should rename this?

null-a:
Thanks!

> I think we don't have defaults for distribution parameters at the moment (right?).

That's right. I imagine we'll perhaps end up with a scheme that allows us to add default parameter info to the distribution meta data, so that it can be included in the docs automatically. However, I think we could just directly write some the code to specify the defaults if we want something right away.

> However, that would imply using Normal instead of Gaussian. Maybe we should rename this?

It's a tough call! I imagine that `LogitNormal` and `LogisticNormal` won't be used explicitly by many people, so the inconsistency won't often cause problems. Perhaps the rename will be more trouble than it's worth, or maybe not...

stuhlmueller:
> That's right. I imagine we'll perhaps end up with a scheme that allows us to add default parameter info to the distribution meta data, so that it can be included in the docs automatically. However, I think we could just directly write some the code to specify the defaults if we want something right away.

I don't think we need the functionality right away.

By the way, should we add a `logitNormal` sampling helper? We seem to have those for all (?)  other documented distributions.

> It's a tough call! I imagine that `LogitNormal` and `LogisticNormal` won't be used explicitly by many people, so the inconsistency won't often cause problems. Perhaps the rename will be more trouble than it's worth, or maybe not...

Yeah, I don't think we should worry about the inconsistency. It does seem desirable to use the more standard name, and `Normal` seems a bit more standard. Going by Google search results, "normal distribution" has 6.5M hits, whereas "gaussian distribution" has 850k. I don't know if that's enough of a difference to justify the trouble.

null-a:
> By the way, should we add a logitNormal sampling helper? We seem to have those for all (?) other documented distributions.

Good catch, I've added it. It could use a test, though adding an inference test seems disproportionate. I'm out of time now, so I'll come back to this tomorrow.

>  I don't know if that's enough of a difference to justify the trouble.

No, me neither.

Issue #430, opened 2016-05-14
Investigate uniformDrift test failure


stuhlmueller:
https://travis-ci.org/probmods/webppl/jobs/130057558

longouyang:
I think we can just bump the tolerance

stuhlmueller:
Failed again: https://travis-ci.org/probmods/webppl/jobs/137013619

PR #641, opened 2016-09-09
Reparameterization trick for Exponential and Cauchy distributions.

null-a:


dritchie:
Is it a good idea to use `Uniform({a: 0, b: 1})` as the base distribution? When the base values are being optimized, they won't obey the implicit [0,1] bounds. Wouldn't it be better to use a sigmoid somewhere in `transform`, and have `base` be `Gaussian`? Let me know if I'm missing something here.

null-a:
> When the base values are being optimized

I'm not sure what you mean by "base values"? Nothing to do with the base distribution is optimized, and auto-guides still generate the same guide parameters + squishing functions as before.

dritchie:
Ah, right--the base distribution is only ever sampled. I was mixing up my internal warning signals from HMC with the ones from variational ;)

null-a:
CI failure is unrelated to this change. (It was the uniformDrift test, #430.)

PR #646, opened 2016-09-15
Integrate new adnn changes

dritchie:
I got tired of all the confusion/questions around "what can you / can't you do with tensors in WebPPL?," so I made a fairly small change to adnn that lets us point directly to adnn and say "everything in ad.tensor is usable, and for convenience, ad.tensor = T." Hopefully, this will:
 - Make it clearer what is/isn't supported by moving it all into one place.
 - Direct people to adnn (and not WebPPL) for tensor feature requests.

### DIFF
--- a/docs/functions/tensors.rst
+++ b/docs/functions/tensors.rst
@@ -26,6 +26,17 @@ Creation
 
      Matrix([[1, 2], [3, 4]])
 
+.. js:function:: Tensor(dims, arr)
+
+   :param array dims: array of dimension sizes
+   :param array arr: array of values
+
+   Creates a tensor with dimension ``dims`` out of a flat array ``arr``.
+
+   Example::
+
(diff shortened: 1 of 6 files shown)


null-a:
Cool, the change looks sensible to me.

Since the change to adnn is not backwards compatible but only the patch version was bumped, I think fresh installs of webppl might be broken now? e.g. [here](https://travis-ci.org/probmods/webppl/jobs/160497431).

dritchie:
Derp, serves me right for not understanding how npm semantic versioning works...
Here's how I've fixed the issue: I re-published adnn 1.1.6 as 1.1.9 (which new installs can version up to), and I re-published the latest version as 2.0.0. I also submitted webppl pull request #647 which makes dev now depend on adnn 2.0.0.

Issue #631, opened 2016-08-16
Add a "list of samples" backed marginal distribution


null-a:
From [#369](https://github.com/probmods/webppl/issues/369#issuecomment-238905554):

> For the results of sampling based inference, we could have a version of the marginal distribution that stores a list of samples internally, and only builds the histogram if `support` or `score` are called. `sample` would just pick an element from the list. This has exactly the same interface as the current marginal distribution, but you don't pay the price of `stringify` (#389) if you don't need to.

PR #654, opened 2016-09-25
Attach the number of samples to Marginals

longouyang:
For some visualization methods (e.g., 2d heatmap), we need to know the number of samples that are backing a marginal distribution.

As far as I can tell, this information isn't stored anywhere, so I've modified `CountAggregator` to store it. Example:

```js
var mdist = Infer({method: 'MCMC', samples: 51},
      function() { return uniformDraw(_.range(10)) });

mdist.params.numSamples // => 51
```

### DIFF
--- a/src/aggregation/CountAggregator.js
+++ b/src/aggregation/CountAggregator.js
@@ -21,13 +21,20 @@ function normalize(hist) {
   var totalCount = _.reduce(hist, function(acc, obj) {
     return acc + obj.count;
   }, 0);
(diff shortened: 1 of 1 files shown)


null-a:
This seems OK to me.

I'm not sure whether having this as a property on `params` is going to be the right place once we've looked at #631 etc., but it seems perfectly fine for now.

I noticed that this approach automatically does the right thing when `justSample=true` which is nice, though I guess that's not important for the intended use case.

Issue #573, opened 2016-07-19
Handle more distribution types when generating default guides


null-a:
The current implementation only knows how to handle distributions where all of the parameters are either numbers or tensors.

PR #657, opened 2016-09-29
Generalize DiagCovGaussian

null-a:


null-a:
This generalizes `DiagCovGaussian` to be a distribution over tensors of any shape rather than just vectors. This change is backwards compatible.

This is then used to provide a more flexibly default guide for `TensorGaussian`.

Partially addresses #573.

dritchie:
Ooh, thanks for this! This'll definitely be useful for doing variational Bayes with neural net parameters.

PR #670, opened 2016-10-11
New ELBO estimator

### LINKED ISSUE #661: Merge new ELBO estimator
The code in the [`new-vi-estimator` branch](https://github.com/probmods/webppl/tree/new-vi-estimator) is close to been ready to merge. Still to do:

* [x] Rename `ELBO2` to `ELBO`.
* [x] Minor clean-ups, as documented in code with `TODO` comments.
* [x] Maybe add back the message suggesting step size be reduced when things blow up.
* [x] Update docs.
* [x] Switch baselines on by default.
* [x] Ensure that the the use `fs` browserifys OK. (I think it will be fine.) Have these fail in a graceful way if an attempt is made to use them in the browser?

null-a:
I think this is ready to be merged. As well as the updated estimator, this PR includes the following:

*  New interface for creating/retrieving guide parameters. There's now a single `param` method that takes an options object. The docs have been updated.
* `modelParam` helper for declaring MLE params.
* `Infer({method: 'forward'}, ...)` now computes the normalization constant when sampling from the target.
* `Optimize` gained a couple options for check-pointing params and persisting the progress made on the ELBo to disk.

Closes #661.


stuhlmueller:
Looks great, especially for such a big change!

A few thoughts that you should feel free to ignore:

- I'd consider factoring out the GraphViz dot generation code into a separate file.
- I wonder whether the user-visible function for creating guide parameters should be called `guideParam`, so that it's symmetric with `modelParam`. Using `param` for creating guide parameters makes sense implementation-wise, but from a user perspective, it seems only justified if guide parameters will be a lot more frequent than model parameters.
- On a first read the code, it wasn't clear to me what the shape of the `ELBO` `state` argument is. I assume it's just for storing baselines? It could be good to document its purpose briefly, as it isn't clear from the name.

PR #673, opened 2016-10-13
Make uniformDraw([]) behave in inference

longouyang:
Outside of inference, `uniformDraw([])` returns `undefined`. Within `Enumerate`, though, it causes an error because the support is empty.

This PR makes the behavior more consistent inside and outside of inference by immediately returning `undefined` if the input array has length 0.

### DIFF
--- a/src/header.wppl
+++ b/src/header.wppl
@@ -143,7 +143,11 @@ var delta = function(arg) {
 
 var uniformDraw = function(arg) {
   var vs = util.isObject(arg) ? arg.vs : arg;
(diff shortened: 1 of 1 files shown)


stuhlmueller:
Is there a reason we don't just throw an error when `uniformDraw` gets an empty list? I would think that this is almost always by accident.

longouyang:
In my model, it's not really by accident -- the argument to `uniformDraw` is computed dynamically and the easiest way to handle the empty case is to rely on the behavior of returning `undefined`.

Issue #62, opened 2015-03-22
Cache compiled webppl header


stuhlmueller:
Before I merged the stochastic-dependence-analysis branch, we were caching the compiled webppl header. After the merge in d94e1642ec7f7f24ff68ef62d17262fa969c7ad6, this is no longer the case. I didn't port the previous solution to caching the compiled header, since it wasn't ideal in the first place. A better solution will generate the header once when some install command is run, and subsequently only read it in from a file.

null-a:
Note that this gets trickier if we allow [user macros to transform the header](https://github.com/probmods/webppl/issues/153#issuecomment-144465613).

stuhlmueller:
I'll take that as another (not very strong on its own) argument that user macros shouldn't transform the header.

null-a:
Something else to consider is that fact that we've since added the freevars transform, which is applied to the header conditionally, depending on whether `IncrementalMH` is used. (There may be more of this kind of thing in the future. #136)

Somewhat related: The call to `sweet.loadModule(<contents of headerMacros.sjs>)` ([here](https://github.com/probmods/webppl/blob/1258853f49c8ae2b97b283e816afcfba382e7ce5/src/main.js#L65)) takes around 0.6s on my machine. (Versus 0.35s for the call to `compile`.) [headerMacros.sjs](https://github.com/probmods/webppl/blob/1258853f49c8ae2b97b283e816afcfba382e7ce5/src/headerMacros.sjs) has very little in it, most of the time here is some kind of overhead which is incurred on the first call to `loadModule`. HMC adds a second call to `loadModule` which takes around 0.3s, assuming the overhead happened in an earlier call. So, it looks like we're spending quite a bit more time loading macro modules than compiling, perhaps this can be cached too.

stuhlmueller:
That's a lot of time spent loading macros! Good to know. As a first step, would there be any problem with memoizing `sweet.loadModule` using a non-cps version of [cache](/probmods/webppl/blob/dev/src/headerUtils.js#L16)?

null-a:
> would there be any problem with memoizing sweet.loadModule using a non-cps version of cache?

I don't think so. I was going to think about doing something which would help in Node too, but that needs some thought. This only helps the browser (right?), but it's so simple that it seems worth doing. I'll take a look.

null-a:
>> would there be any problem with memoizing sweet.loadModule using a non-cps version of cache?

> This only helps the browser (right?),

[...]

PR #675, opened 2016-10-14
Remove macros

### LINKED ISSUE #270: Update sweet.js
The way I made browserify work with macros (#153) is rather ugly. In particular I had to:

1. [Patch](https://github.com/probmods/sweet.js/commit/77d2efdb8f4d81a6b57b8891bc3b1fa6e14c1d9e) sweet.js.
2. Pin us to an old version of brfs. (I mention one of the reasons for this [here](https://github.com/substack/static-module/issues/17).)
3. [Switch](https://github.com/probmods/webppl/blob/1258853f49c8ae2b97b283e816afcfba382e7ce5/compiled/Makefile#L16) to applying brfs globally. (`browserify -g` rather than `-t`.)

We'll presumably need to re-visit this at some point.

### LINKED ISSUE #484: Improve startup time
Flipping a coin on the command line currently takes 2.5s:

~~~~
parsePackageCode: 2072.600ms
compile: 724.353ms
run: 2.793ms
~~~~

### LINKED ISSUE #668: Operator precedence - possible parser issue
In the course of using WebPPL for Psych 204 at Stanford, I ran into this unexpected behavior that clashes with JavaScript operator precedence rules:

The following function returns `false`

```
var foo = function(g, x, y){
  return g(x,y) > x+y ? "yes" : "no"
}

foo(function(x,y){3*x*y},2.6,1.2)
```

However, this returns `true`, as it should

```
var bar = function(g, x, y){
  return g(x,y) > x+y
}

bar(function(x,y){3*x*y},2.6,1.2)
```

and the parentheses make it return `"yes"` as it should:

```
var baz = function(g, x, y){
  return (g(x,y) > x+y) ? "yes" : "no"
}

baz(function(x,y){3*x*y},2.6,1.2)
```

null-a:
This removes macros as discussed in #270. Summary:

- The ad macros are replaced by a new code transform.
- The macro that created distribution constructors has been replaced with a script that generates the code based on distribution metadata. This runs when the npm package is installed, i.e. at the same time we 'adify' back end code. I also took the opportunity to switch to generating the sampling helpers using this machinery too.
- Remove macros infrastructure and the dependency on sweet.js.

*Syntax errors*

One question we had before looking at this was whether dropping sweet.js would make syntax errors less useful. In the cases I've checked, it appears they are quite similar. For example, on this program:

```js
a[;
```

... we get:

```
# before
Error: Line 2: Unexpected end of input
[... ) ( )  ...]

# after
Error: Line 1: Unexpected token ;
```

*Editor*

I've tested this branch with the browser editor and it seems work OK. There may be a small issue with the formatting of syntax errors (though they are still readable) so I'll open an issue for that. (Now probmods/webppl-editor#43.)

*Start-up time*

One of the reasons for doing all of this was to try and improve the start-up. Here's the before and after for flipping a coin on my latop:

```
# before
time webppl flip.wppl --verbose
parsePackageCode: 1371.672ms
compile: 428.589ms
false
run: 1.823ms

real	0m2.132s

# after
time webppl flip.wppl --verbose
parsePackageCode: 40.592ms
compile: 477.842ms
false
run: 1.763ms

real	0m0.847s
```

Closes #270.
Closes #484. (For further improvements we'd look at #62 I think.)
Closes #668.


stuhlmueller:
Looks good!

Also remove the `headerMacros.sjs` file?

null-a:
> Also remove the headerMacros.sjs file?

Oops, yes, that would be a good idea - thanks!

PR #684, opened 2016-10-25
Split out graph related bits from elbo.js.

null-a:
The [suggestion](https://github.com/probmods/webppl/pull/670#issuecomment-253297029) on #670 was to split out the graph viz related code from `elbo.js`. That turns out to be tricky without splitting out the rest of the graph related code too, so that's what this does. Does this help?


stuhlmueller:
Yes, that's better than my suggestion.

PR #688, opened 2016-10-28
Improved interface for Infer and Optimize.

### LINKED ISSUE #629: Have Infer default to using Enumerate

null-a:
This implements the interface for `Infer` suggested in #629, allowing us to specify a default inference method. I've also updated `Optimize` to work in much the same way.

Since ES6 style object literal shorthand already works in WebPPL, then this change allows us to write things like:

``` js
Infer({model() {
  return flip();
}});

Infer({method: 'rejection', model() {
  return flip();
}});

var model = function() {};
Infer({model});
Infer({model, method: 'rejection'});
```

I've update the docs to reflect this new interface. However, the old interface still works, so I don't expect this change to break anything.

Closes #629.

### DIFF
--- a/src/inference/optimize.js
+++ b/src/inference/optimize.js
@@ -26,7 +26,25 @@ module.exports = function(env) {
     EUBO: require('./eubo')(env)
   };
 
-  function Optimize(s, k, a, wpplFn, options) {
+  function Optimize(s, k, a, fnOrOptions, maybeOptions) {
+    var wpplFn, options;
+    if (_.isFunction(fnOrOptions)) {
+      wpplFn = fnOrOptions;
+      options = maybeOptions;
+    } else if (util.isObject(fnOrOptions)) {
+      wpplFn = fnOrOptions.model;
+      options = _.omit(fnOrOptions, 'model');
+      if (!_.isFunction(wpplFn) && _.isFunction(maybeOptions)) {
+        throw new Error('Optimize: expected model to be included in options.');
+      }
+    } else {
+      throw new Error('Optimize: expected first argument to be an options object or a function.');
+    }
+
+    if (!_.isFunction(wpplFn)) {
+      throw new Error('Optimize: a model was not specified.');
+    }
+
     options = util.mergeDefaults(options, {
       params: {},
       optMethod: 'adam',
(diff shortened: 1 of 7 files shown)


src/inference/optimize.js
```diff
@@ -26,7 +26,25 @@ module.exports = function(env) {
     EUBO: require('./eubo')(env)
   };
 
-  function Optimize(s, k, a, wpplFn, options) {
+  function Optimize(s, k, a, fnOrOptions, maybeOptions) {
+    var wpplFn, options;
```
stuhlmueller: Is there a reason you renamed `wpplFn` to `model` in `Infer` but kept it here? If not, rename to `model` here as well?

tests/test-data/deterministic/models/infer.wppl
```diff
@@ -0,0 +1,9 @@
+var model = function() {
+  return flip(1);
+};
+[
+  // Test the use of ES6 object literal shorthand.
+  sample(Infer({model})),
+  sample(Infer({model: model})),
+  sample(Infer({model() { return model(); }}))
```
stuhlmueller: I was initially surprised that this works in Node v0.12, given that it doesn't support the ES6 object literal extensions. But Esprima transforms this into the corresponding standard syntax tree, so everything works out. :-)

null-a: Exactly. Note that esprima does distinguish between the long and short literal syntax with an extra flag on the property node, but this doesn't make it through our transforms, which is why code gen always outputs the long hand version.

I added these tests to make sure we find out if something occurs which breaks this. At some point I imagine switching all of the examples and inference tests over to the short hand syntax, but I figured I'd wait to see whether it actually becomes idiomatic first. (Though I'm increasingly confident that it will, so I'd be happy to change it soon if there are no objections.)

src/inference/optimize.js
null-a: I agree this is inconsistent -- thanks! I think the fix should be to change `Infer` back to using `wpplFn` rather than `model`, since that's what's used in lots of other places, including similar helpers in `header.wppl`. I've pushed a commit that does this, does that seem OK to you?

tests/test-data/deterministic/models/infer.wppl
stuhlmueller: I'd be happy with switching examples and tests now.

PR #690, opened 2016-11-02
Make error handling work for errors throw in deeply nested JS function calls.

### LINKED ISSUE #613: Strange "can't read property webppl" error msg with wrong data type in inference args
When you (incorrectly) write the samples arg in inference opts as a string, you get:

```
Cannot read property 'webppl' of undefined
```

Minimal example

```
Infer({method: 'MCMC', samples: '5000'}, function(){return flip(0.5)});
```

null-a:
Closes #613.

### DIFF
--- a/src/errors/errors.js
+++ b/src/errors/errors.js
@@ -54,14 +54,18 @@ function filterJsStackTrace(stackTrace) {
   // taking the top-most entry corresponding to a function application
   // in webppl code, and any frames above it.
 
+  // When no entry corresponding to application of a webppl function
+  // is present, we return the entire stack. The reason this is
+  // possible is that (by default in V8) only the top 10 stack frames
+  // are captured on error.
+
   // We can only ever take the top-most webppl frame as any earlier
   // frames may have been generated on a different execution path.
   // (And besides, there should only be one entry as the JS stack is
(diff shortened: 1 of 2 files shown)


src/errors/errors.js
```diff
@@ -54,14 +54,18 @@ function filterJsStackTrace(stackTrace) {
...
+
   // We can only ever take the top-most webppl frame as any earlier
   // frames may have been generated on a different execution path.
   // (And besides, there should only be one entry as the JS stack is
   // cleared between each webppl function application.)
 
   var ix = _.findIndex(stackTrace, _.matcher({webppl: true}));
-  assert.ok(ix !== undefined, 'Expected to find wppl entry in JS stack trace.');
-  return stackTrace.slice(0, ix).concat(stackTrace[ix]);
+  return (ix >= 0) ? stackTrace.slice(0, ix).concat(stackTrace[ix]) : stackTrace;
```
stuhlmueller: Why not `stackTrace.slice(0, ix+1)` instead of `stackTrace.slice(0, ix).concat(stackTrace[ix])`?

null-a: No good reason! Changed now.

PR #692, opened 2016-11-03
Use new Infer interface throughout.

null-a:
As [discussed](https://github.com/probmods/webppl/pull/688#discussion_r85776862) on #688.

### DIFF
--- a/examples/binomial.wppl
+++ b/examples/binomial.wppl
@@ -5,4 +5,4 @@ var binomial = function() {
   return a + b + c;
 };
 
-Infer({method: 'enumerate'}, binomial);
+Infer({model: binomial});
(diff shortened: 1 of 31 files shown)


stuhlmueller:
Nice! Should we update the docs, too?

null-a:
I updated them to `Infer({method: '...', model: model})`, but I didn't change that to `Infer({method: '...', model})` to avoid confusing folk who aren't aware of the short hand syntax yet. Are you suggesting we change the docs to the latter?

stuhlmueller:
I don't think we should use the shorthand syntax everywhere (for the reason you mention), but maybe we should add a sentence about it somewhere if we're planning to use it in essentially all of our code?

null-a:
I thought the same. The snag is that the sensible place for such seems to be on a page describing the language, which of course we don't have, yet.

null-a:
As a stopgap, we could give the shorthand version underneath the usage example here:
http://webppl.readthedocs.io/en/dev/inference/index.html

stuhlmueller:
OK.

PR #697, opened 2016-11-05
Dream learning

dorarad:
Try again after updating gitignore and gruntfile.


stuhlmueller:
The files `dreamEubo.js` and `dreamSample.js` are not supposed to be part of the PR, right? They're automatically generated after `npm install` from `dreamEubo.ad.js`and `dreamSample.ad.js`.

dritchie:
Got a chance to read through it in some detail. My comments/questions:

- The separation between `DreamSample` and `DREAM` means that you generate a batch of hallucinated traces, and then optimize on those. I think we want to be able to alternate between hallucinating traces and gradient updates on those traces. That's not currently possible, though it would become possible if we provide a mechanism for passing optimization state between different calls to Optimize (so that we could truly 'resume' optimization with a different coroutine). Is this what you had in mind when designing things this way? I like how simple and modular this setup is, so it'd be nice to make it work with the alternation (rather than mashing both `DreamSample` and `DREAM` together into one mega-coroutine.
- The way you're checking for whether the coroutine is inside `mapData` won't work for nested `mapData`. One alternative that will work: maintain a stack that you push on every `mapDataFetch` and pop on every `mapDataFinal`. The coroutine is inside `mapData` whenever this stack is not empty.
- `DreamSample.factor`: I think you don't actually need this, since we don't use the score of this trace, right?
- `DreamSample.observe`: I don't think the `addChoice` is necessary, since the `DREAM` scorer won't ever touch these choices (it only interacts with them with the `observations` array).
- `dreamEubo.ad.js`, line 137: you seem to have copied the comment here from `eubo.ad.js`, but not the line that actually does the un-lifting (i.e. `var _guideVal = ad.value(guideVal)`) before scoring. I think it’s safest to keep that around.
- `ELBO.mapDataFetch`: I would definitely switch to a more efficient implementation for building the sub data array out of the index array. Something like:
    
     ```
    var subData = new Array(ix.length);
    for (var i = 0; i < ix.length; i++) {
       subData[i] = data[ix[i]];
    }
    return subData;
    ```
    should do it.
- What's the motivation behind changing `_.isFinite` to `isFinite` in some places and `isFinite` to `Number.isFinite` in others?

@null-a: Paul, one item that definitely could use your attention in this PR is whether/the extent to which the change to the `mapDataFetch` interface (from returning indices to returning data items directly) will mess up anything else in the codebase.

src/headerUtils.js
```diff
@@ -153,29 +154,31 @@ module.exports = function(env) {
...
     }
 
-    var ix = env.coroutine.mapDataFetch ?
+    // var ix = env.coroutine.mapDataFetch ?
+    //     env.coroutine.mapDataFetch(data, batchSize, a) :
+    //     null;
+    var _data = env.coroutine.mapDataFetch ?
         env.coroutine.mapDataFetch(data, batchSize, a) :
-        null;
+        data;
```
null-a: IIRC, the reason I had `mapDataFetch` returning indices was so that the observation function is called with an address based on the index of the current datum. This is necessary for (automatic) mean field + mini-batches for example. I understand that for dream learning we need to inject the hallucinated observations, so perhaps we need to do something like return both data and indices? (I'm not sure I've explained that very well, let me know if it doesn't make much sense!)

null-a:
My understanding is that this initial version makes some assumptions about the structure of the program to simplify building records. I think these are: there is one `observe` per evaluation of `mapData`s observation function and there is no nested `mapData`. Maybe we should have a comment somewhere describing this? Also, we might consider trying to notice (at run time) when these assumptions are broken. This might not be important if this is going to evolve quickly, but it might be worth it if it's a simple change?

src/inference/dreamEubo.ad.js
```diff
@@ -0,0 +1,181 @@
...
+        // TODO: not sure whether it's needed or not in our case
+        if (ad.isLifted(objective)) { // Handles programs with zero random choices.
+          objective.backprop();
+        }
+
+        var grads = _.mapObject(this.paramsSeen, function(params) {
+          return params.map(ad.derivative);
+        });
+
+        return cont(grads, -ad.value(objective));
```
null-a: The second argument passed to the continuation should be the value of the optimization objective. (So something like `loqp - logq`.) I messed this up when I implemented eubo, but have just submitted #699 to fix it. I think we probably want to do something similar here. (I guess the difference is we don't want to use the score of the entire trace for `logp`, rather we want the scores from the choices in the observation function.)

dorarad:
Andreas: I added them since the automatically generated files failed in `grunt lint`, and so I ran `fixjsstyle` and then added the fixed files to the PR. The alternative which we'll go on for now is to make lint ignore them in `gruntfile.js` and remove them from the PR. However, I wonder if it may be better later to change the grunt command so that it runs `fixjsstyle` as a first step before lint?

Daniel:

- I thought that having modularity between the two phases is the correct design since these two phases can  be used on their own, e.g. `DREAM` (eubo) estimator can in principle have the traces from various sources rather than dreamSample. The separation can also allow alternations between these two steps - would be glad to talk with you more in slack / in person about that. However, I still think that the `DreamTrainThenSample` function should be provided for the same reasons that routines like the 2-in-1 `OptimizeThenSample` or the specific `ParticleFilterRejuv` are.

- You are right that current flag implementation won't work. Will change that. However, since right now we plan to consider everything that is in the outmost mapData as the local part (where we are sampling from target), I think stack is a bit of an overkill - if I understand correctly a counter of nesting level should be enough.

- `DreamSample.factor` + `DreamSample.observe` : yea that's right it's indeed not necessary for dream to work - I did it for the sake of maintaining the consistency of the trace data-structure, comparing to its usage in other places. Actually I think it may generally be better that the trace will have its own methods that are responsible for its internal fields such as numFactors and score rather than having each inference method (such as `smc`, `hmckernel` and `mhkernel`) update it separately, but let's talk about that in my next clean-up PR. For now I'll remove that.

- `dreamEubo.ad.js`, line 137: Fixed

- `elbo.mapDataFetch`:  Sure, fixed. When I implemented that I thought I would need to have some cpsLoop or other cps-based solution for the `for` loop, but I guess inside the `js` its ok to have just a standard for..

- isFinite: Ultimately we indeed should use either isFinite or Number.isFinite (not both). I changed _.isFinite to isFinite is in all places following Paul's PR 685. I thought that we should use Number.isFinite in all places since it [enforces the argument is of type Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite) but for some reason when I did this it passed tests only locally and not in travis-ci. So for now I'll have isFinite in all places, and I'll look into the Number.isFinite issue for the next PR.

Paul: Thank you very much! Yea tests and docs update are in progress and I'll submit them in a separate PR. Actually I also did some documentation for other things besides dream..Please see my responses to your comments above.

src/inference/dreamEubo.ad.js
```diff
@@ -0,0 +1,181 @@
...
+
+      }.bind(this), this.a);
+
+    },
+
+    // Since we allready have all the records from a previous pass,
```
dorarad: Fixed, thanks. I guess I should enable spell checker in sublime.. :)

src/inference/dreamEubo.ad.js
```diff
@@ -0,0 +1,181 @@
...
+    },
+
+    factor: function(s, k, a, score) {
+      // if (!isFinite(ad.value(score))) {
+      //   throw new Error('DREAM: factor score is not finite.');
+      // }
+      return k(s);
+    },
+
+    // Instead of returning the origina data of observations, we inject
```
dorarad: Fixed

src/inference/dreamEubo.ad.js
```diff
@@ -0,0 +1,181 @@
...
+      assert.notStrictEqual(val, undefined);
+
+      if (this.insideMapData) {
+
+        // We unlift guideVal to maintain the separation between the ad
+        // graph we're building in order to optimize the parameters and
+        // any ad graphs associated with the example traces. (The
+        // choices in an example trace can be ad nodes when they are
+        // generated with SMC + HMC rejuv.)
+        var _guideVal = guideDist.score(val);
```
dorarad: Totally agree

src/inference/dreamEubo.ad.js
```diff
@@ -0,0 +1,181 @@
...
+      // }
+      return k(s);
+    },
+
+    // Instead of returning the origina data of observations, we inject
+    // our hallucinated observations from the corresponding record.
+    mapDataFetch: function(data, batchSize, address) {
+      this.insideMapData = true;
+      assert(this.currRecord);
+      return currRecord.observations;
```
dorarad: yea, fixed.

src/inference/dreamSample.ad.js
```diff
@@ -0,0 +1,138 @@
...
+
+    this.params = this.opts.params;
+    this.wpplFn = wpplFn;
+    this.s = s;
+    this.k = k;
+    this.a = a;
+
+    // true when inside mapData (i.e. local model)
+    // false otherwise (i.e. global model)
+    this.insideMapData = true;
```
dorarad: Yea sorry I did it for debugging purposes. Fixed.

src/inference/dreamSample.ad.js
```diff
@@ -0,0 +1,138 @@
...
+
+    this.records = [];
+    this.currRecord;
```
dorarad: Maybe it's because I'm used to work more with other languages, but it personally seems nice to me that variables that are used throughout a module are declared in the beginning, even tough it practically does nothing here, but on the other hand it's non-js style so maybe I should change that. What do you think?

src/inference/dreamSample.ad.js
```diff
@@ -0,0 +1,138 @@
...
+
+    },
+
+    // Samples from the guide outside mapData and from the target inside mapData.
+    // Add this choice to the recorded trace.
+    sample: function(s, k, a, dist, options) {
+      var distribution = this.insideMapData ?
+          dist : (options && options.guide) || guide.independent(dist, a, env);
+      var _val = distribution.sample();
+      var val = this.ad && dist.isContinuous ? ad.lift(_val) : _val;
```
dorarad: Yea, I wasn't completely sure about when do we need lifting and so I added that since I've seen it's used in sample calls of `smc`, `mhkernel`, `hmckernel` and `initialize` (i.e. `mcmc`)..I'll remove it now and will be glad to talk with you or Daniel more about the lifting later.

src/inference/dreamSample.ad.js
```diff
@@ -0,0 +1,138 @@
...
+    },
+
+    // Instead of factoring the score of the observed given value,
+    // in dream training we sample a hallucinated value instead and
+    // push that to the corresponding record.
+    observe: function(s, k, a, dist, val) {
+      var _hallucinatedVal = dist.sample();
+      var val = this.ad && dist.isContinuous ?
+          ad.lift(_hallucinatedVal) : _hallucinatedVal;
+      this.currRecord.trace.addChoice(distribution, val, a, s, k, options);
```
dorarad: Yea, `distribution` should have been `dist`.. And I also agree it's not needed for dream - as I replied to Daniel, I did it in order to maintain consistency in the trace data-structure. Will remove that.

src/headerUtils.js
```diff
@@ -153,29 +154,31 @@ module.exports = function(env) {
...
```
dorarad: No problem. We can definitely return both from mapDataFetch if needed. I just wonder if what you've said about the observation being called with an address is already implemented in the codebase? Maybe I misunderstood something but when I changed the mapDataFetch I've seen that it's only called from Elbo and that the `ix` are used only inside `mapData`, `cpsMapData` and `elbo.mapDataFetch`. It seems to me that these will continue to work without returning indices from fetch.. Do I miss something or did you mean that we should return indices to support future additions to webppl?

src/inference/dreamEubo.ad.js
dorarad: Yea I agree. I initially thought only about the `logq` part which is relevant to the gradients but indeed the objective should also contain `logp`. Will fix.

src/headerUtils.js
null-a: Following on from our chat yesterday, here's some code that demonstrates the property we need to preserve:

```js
// The code that follows requires this definition adding to e.g.
// headerUtils.js:

// function getAddress(s, k, a) {
//   return k(s, a);
// }

var model = function() {
  display('start');
  mapData({data: 'abc'.split(''), batchSize: 1}, function(x) {
    display('address: ' + getAddress() + ' | x: ' + x);
  });
};

// When running the model, the address within the observation function
// corresponds to the current value of x.
model();

// start
// address: _159_158_$$0_156 | x: a
// address: _159_158_$$1_156 | x: b
// address: _159_158_$$2_156 | x: c

// When using mini batches with elbo, the observation function will
// only be called on a subset of the data. When the observation
// function is called with some particular value of x, the address should
// be the same as the address used for this x when not doing mini batches,
// as above. Like so:

Optimize({model, steps: 5});

// start
// address: _160_158_$$1_156 | x: b
// start
// address: _160_158_$$2_156 | x: c
// start
// address: _160_158_$$1_156 | x: b
// start
// address: _160_158_$$0_156 | x: a
// start
// address: _160_158_$$2_156 | x: c
```
If you see a way to do it that doesn't involve returning both the indices and data from `mapDataFetch` that's fine with me!

src/inference/elbo.ad.js
```diff
@@ -354,8 +354,11 @@ module.exports = function(env) {
         // Signal to mapDataFinal that the batch was empty.
         this.mapDataStack.push(null);
       }
-
-      return ix;
+      if (ix === null) {
+          return data;
+      }
+      // TODO - change to a more efficient solution?
+      return data.filter(function(val, index){ return ix.includes[index] });
```
null-a: This should be `includes(index)` rather than `includes[index]`. I realize this will change anyway, but I found it necessary to make this change when testing the code from [this comment](https://github.com/probmods/webppl/pull/697#discussion_r87580260).

src/headerUtils.js
dorarad: Cool thanks a lot! I'll look into that.

src/inference/dreamEubo.ad.js
```diff
@@ -0,0 +1,189 @@
...
+    // The local model includes anything inside of some mapData (level 1+)
+    this.mapDataNestingLevel = 0;
+
+    this.coroutine = env.coroutine;
+    env.coroutine = this;
+  }
+
+  function isInsideMapData() {
+    return this.mapDataNestingLevel > 0;
+  }
```
null-a: `this` will be undefined here. `isInsideMapData` will need to be a method on the coroutine for this to work.

src/inference/dreamEubo.ad.js
```diff
@@ -0,0 +1,189 @@
...
+      if (isInsideMapData()) {
+
+        // We unlift guideVal to maintain the separation between the ad
+        // graph we're building in order to optimize the parameters and
+        // any ad graphs associated with the example traces. (The
+        // choices in an example trace can be ad nodes when they are
+        // generated with SMC + HMC rejuv.)
+        var _guideVal = ad.value(guideVal);
+        
+        var _guideScore = guideDist.score(_guideVal);
```
null-a: I should explain why you're seeing a bunch of variable names prefixed with underscores. When a piece of code has to deal with both a node in the ad graph and the node's value, I often use similar variables for both but prefix the name of the latter with an underscore. That's what's happening here with `guideVal` and `_guideVal`. You don't need to do this for the score variable for example.

src/inference/dreamEubo.ad.js
```diff
@@ -0,0 +1,189 @@
...
+          objective.backprop();
+        }
+
+        var grads = _.mapObject(this.paramsSeen, function(params) {
+          return params.map(ad.derivative);
+        });
+
+        var logp = ad.value(this.currRecord.trace.score);
+        var logq = ad.value(this.logq);
+        return cont(grads, logp - logq);
```
null-a: Since we only accumulate the score of local choices in `this.logq`, I wonder if `logp` should also be the sum of the scores from the local choices? (i.e. not include the global choices, as I assume the trace score does.)

src/inference/dreamSample.ad.js
```diff
@@ -0,0 +1,143 @@
...
+    sample: function(s, k, a, dist, options) {
+      var distribution = isInsideMapData() ?
+          dist : (options && options.guide) || guide.independent(dist, a, env);
+      var _val = distribution.sample();
+      // var val = this.ad && dist.isContinuous ? ad.lift(_val) : _val;
+      
+      // Accumulates score of samples inside mapData only, which can be 
+      // used in the objective computation in dreamEUBO
+      if (isInsideMapData()) {
+        this.currRecord.trace.addChoice(distribution, val, a, s, k, options);
```
null-a: Do we not want to add 'global' variables to the trace? In `dreamEUBO` we still try to look them up.

Also, `val` is now undefined here?

src/inference/dreamEubo.ad.js
```diff
@@ -0,0 +1,189 @@
...
+        // generated with SMC + HMC rejuv.)
+        var _guideVal = ad.value(guideVal);
+        
+        var _guideScore = guideDist.score(_guideVal);
+        checkScoreIsFinite(_guideScore, 'guide');
+
+        this.logq += _guideScore;
+      }
+
+      return k(s, val);
```
null-a: I think `val` is undefined?

dorarad:
@null-a Thanks for the comments - I need to commit some update to dreamSample and dreamEubo will submit in a few minutes.

src/inference/dreamEubo.ad.js
dorarad: Please look at the new commit.

dorarad: Please look at the new commit.

src/inference/dreamEubo.ad.js
dorarad: Please look at the new commit.

src/inference/dreamSample.ad.js
dorarad: Please look at the new commit.

src/inference/dreamEubo.ad.js
dorarad: I see alright, thought it used for "local" variables vs params. fixed.

src/inference/dreamSample.ad.js
null-a: I understand where you're coming from. I think we perhaps either remove it for consistency with the rest of the code base, or we change it to something like `this.currRecord = null` so that it says something more like "i'm creating this property, i just don't know what its value it yet".

src/inference/dreamEubo.ad.js
```diff
@@ -0,0 +1,189 @@
...
+    this.s = s;
+    this.a = a;
+
+    // The global model includes anything outside of any mapData (level 0) 
+    // The local model includes anything inside of some mapData (level 1+)
+    this.mapDataNestingLevel = 0;
+
+    this.isInsideMapData = function () {
+      return this.mapDataNestingLevel > 0;
+    }
```
null-a: This works fine, but it's probably best to add this method as a property of `DREAM.prototype` since that's the way all of the other methods (`run`, `sample`, etc.) are added.

src/inference/dreamSample.ad.js
```diff
@@ -0,0 +1,145 @@
...
+
+    // Instead of factoring the score of the observed given value,
+    // in dream training we sample a hallucinated value instead and
+    // push that to the corresponding record.
+    observe: function(s, k, a, dist, val) {
+      var _hallucinatedVal = dist.sample();
+      // var val = this.ad && dist.isContinuous ?
+      //     ad.lift(_hallucinatedVal) : _hallucinatedVal;
+      // this.currRecord.trace.addChoice(dist, val, a, s, k);
+      this.currRecord.observations.push(val);
```
null-a: The hallucinated value is no longer added to the trace?

src/inference/dreamSample.ad.js
```diff
@@ -0,0 +1,145 @@
...
+          dist : (options && options.guide) || guide.independent(dist, a, env);
+      var val = distribution.sample();
+      // var val = this.ad && dist.isContinuous ? ad.lift(_val) : _val;
+      
+      this.currRecord.trace.addChoice(distribution, val, a, s, k, options);
+
+      // Accumulates score of samples inside mapData only, which can be 
+      // used in the objective computation in dreamEUBO
+      if (this.isInsideMapData()) {
+        this.currRecord.samplesScore = ad.scalar.add(this.currRecord.samplesScore, distribution.score(val));
```
null-a: BTW, since this whole file goes through the ad transform, you can write this as `this.currRecord.samplesScore += distribution.score(val)` and it will be transformed into what you've written here.

null-a:
> I'll test that later today but for now I'm not sure I completely understand how it's assumed that there is one observe evaluation per mapData.

My understanding is that the assumption is that there's only a single `observe` within the "observation function". Of course, this is evaluated once for each piece of data we look at. Does that clear things up?

> Regarding the assumptions: will be glad to monitor if/when the assumptions are broken + add documentation

Great -- just a comment in the code that explains what form the model has to take for this is to work would be really helpful.

src/inference/dreamEubo.ad.js
dorarad: Sure. Will fix.

src/inference/dreamSample.ad.js
dorarad: That's really cool actually, didn't know that.

dorarad:
Regarding the one `observe` per observation function. You're right. I've realized that it's indeed assumed right now. I'm planning to add support for multiple observes by creating the records array inside every mapDataEnter and mapDataLeave so that we'll have observations array per data item. And so `observations` will become an array (corresponding to data points in mapData) of arrays (corresponding to each observation per single data point). In the long run I think we should support not only arrays but also objects, by doing kind of injection of the hallucinated data to the original data points. Actually, IIUC, it should be straightforward to support that in JS with some underscore function..

src/inference/dreamSample.ad.js
dorarad: Sure - so I'll change it to the `= null` option, and let me know if you'd later prefer me to remove it.

null-a:
> I'm planning to add support for multiple observes by creating the records array inside every mapDataEnter and mapDataLeave so that we'll have observations array per data item.

I suggest we stick with the assumption of a single `observe` for this PR, and then work on relaxing that as a follow-up PR. I also think it would be worth having one or two inference tests in place before we tackle this so that it's easy to know what's working and what isn't. I'm not sure exactly what these tests will look like, so these might also come on a follow-up PR. What do you think?

BTW, since the case where there's a single `observe` is very common, I think we should make sure it continues to work without the user having to transform the array passed to `mapData` into an array of (singleton) arrays.  (This might be something you already have in mind, but I figured I'd mention it just in case.)

dorarad:
Yea I can do it in a separate PR. It's better to do that in steps rather than all at one merged PR.
Regarding tests: I've been working on some tests and planned to submit them in a follow-up PR indeed.
Also agree with your last comment. Since it's a common usecase, let's support single observe right now. In a follow up PR I'll add tests, and in a second follow-up will generalize the impl so that we have less assumptions.

ngoodman:
i'm in favor of keeping this PR restricted (eg assume just one observe). and not adding dream to docs (so people won't think they can use it). that way we can move on to putting our energy into expts to understand when dream is useful.... after that we can make it cleaner and more elegant.

dorarad:
@ndg Completely agree with you. Working on expts right now.

dorarad:
@dritchie I added support of keeping history following our discussion above. The current implementation keeps only most recent params by default, and when keepParamsHistory is true, then it records previous updates in a JSON array, removing and re-appending the ']' every write, so that this is safe against crashes.

Example of output with keepParamsHistory = false:
{"_156_$$0_155$mf$Bernoulli$p":[{"dims":[1],"length":1,"data":[1.4708630715902236]}]}

Example of output with keepParamsHistory = true:
[{"_156_$$0_155$mf$Bernoulli$p":[{"dims":[1],"length":1,"data":[0]}]},
{"_156_$$0_155$mf$Bernoulli$p":[{"dims":[1],"length":1,"data":[0.044778214342635486]}]},
{"_156_$$0_155$mf$Bernoulli$p":[{"dims":[1],"length":1,"data":[0.09448171834945263]}]},
{"_156_$$0_155$mf$Bernoulli$p":[{"dims":[1],"length":1,"data":[0.14197375650811456]}]}]

dritchie:
Cool, thanks for fixing that up.

dorarad:
@dritchie regarding your comment about DreamSample.observe: I think `addChoice` is necessary because by adding that to the trace you updating the p_score to be p(x,y) instead of p(x) in order to compute ultimately EUBO objective (rather than KL-divergence).

src/inference/dreamSample.ad.js
dorarad: It's definitely should be added yea.

dorarad:
@dritchie regarding DreamSample.factor: same thing, since we actually use the trace score for the eubo computation we need that.

src/inference/dreamEubo.ad.js
null-a: To me, this doesn't feel like it strongly conveys the idea of public/private, and been consistent with surrounding code trumps this in my mind.

`checkScoreIsFinite` and `top` are different because they don't need to access `this`.

dorarad: Alright - will change that.

dorarad:
quick FYI: following my recent commits, the PR is not ready for merge/code-review. Dream is under development so not at a stable/clean state right now. I'll add a comment once it's ready for merge/code-review.

Issue #689, opened 2016-10-29
Categorical should take array or vector


ngoodman:
Like Discrete, Categorical should take a vector or an array. Makes hierarchical models prettier.

null-a:
Closed by #706.

PR #706, opened 2016-11-15
Categorical - accept probs as vector, accept unnormalized probs

### LINKED ISSUE #686: Categorical should normalize ps or warn if not normalized
Webchurch normalized the probability arguments to its categorical distribution; people used to that will be surprised by things like:

``` js
var d = Categorical({ps: [1, 1], vs: ['a','b']});
d.score('b'); // returns 0, not -0.69
```

I think we should either normalize for the user or warn if the arguments don't (approximately) sum to 1.

More generally, we should check input arguments to base distributions.

null-a:
The main change here is that `Categorical` can now take either an array or a vector of probs, and these can be unnormalized. (Closes #686, #689.) This PR includes basic tests for this new functionality, and I've also gone back and added similar tests for `Discrete`.


src/dists.ad.js
```diff
@@ -1350,27 +1357,25 @@ var Marginal = makeDistributionType({
 
 var Categorical = makeDistributionType({
   name: 'Categorical',
-  desc: 'Distribution over elements of ``vs`` with ``P(vs[i]) = ps[i]``',
-  params: [{name: 'ps', desc: 'array of probabilities', domain: interval(0, 1)},
+  desc: 'Distribution over elements of ``vs`` with ``P(vs[i])`` proportional to ``ps[i]``',
+  params: [{name: 'ps', desc: 'array or vector of probabilities (can be unnormalized)', domain: gt(0, 1)},
```
null-a: Ah, yes, thanks! Now fixed.

Issue #660, opened 2016-10-01
Parameter-server interface for params


ngoodman:
After thinking about a extensions of the system to allow asynchronous distributed parallelism, at least for VI, I'm leaning toward refactoring the params to live in a global store (rather than passing local params objects, which I've previously championed).

By doing this and making a suitable API for the params store (something along the lines of `getParam(name)`, `setParam(name, value)`, `incrParam(name, delta)`), we could back the params by local memory, shared memory, or a network based protocol. This would allow pretty easy implementation of parallel workers....

stuhlmueller:
Here's one way to go about this:

1. Adapt the [adnn optimization methods](https://github.com/dritchie/adnn/blob/master/opt/methods.js) so that we can get back raw deltas.

2. Implement a parameter store (initially in-memory) that supports the following functions:
   
    - `getParams(id) -> params`
    - `setParams(id, params) -> void`
    - `incParams(id, deltas) -> void`

    Here, `id` is the name for a set of parameters. `params` has the same format as in the current implementation, with a name for each parameter. The parameter set ID is used to support storing multiple sets of parameters for the same program (e.g. for different runs), and to allow multiple programs to use the same simple parameter names (like `mu`). This matters most for a network-based parameter store that is used by multiple programs. By default, it could be set to a unique ID that is generated when webppl is started (so there's no re-use across runs), but alternatively we could support setting it using a command line parameter (like the random seed). If we have a runner that starts multiple parallel webppl processes, it could set this ID such that it is the same for each process.

3. At the beginning of each `Optimize` iteration, call `getParams` to retrieve the current params from the store. This replaces the current functionality for passing in `params` via `options`.

4. When an `Optimize` iteration ends:

    1. If parameters were empty at the beginning of the iteration, call `setParams` with the newly initialized parameters.

    2. Call `incParams` using the deltas from the optimizer. 

The current functionality for returning new parameters from `Optimize` goes away; instead the parameters can be retrieved using `getCurrentParams()`, which calls `getParams` with the ID for the active program execution.

Instead of having `incParams` return nothing, we could also have it return the store's current parameter set and thus merge steps 3 and 4.2 on every iteration except for the first. 

This differs from Noah's suggestion in that it only allows all parameters for a program to be set in batch. I'm not sure if this is better or worse. I mainly explored it because it seems like such a small divergence from our current implementation, but some form of batching is desirable for network-based storage anyway. Are there significant advantages to having the API be per-parameter?

[...]

Issue #634, opened 2016-08-23
`registerParams` should only lift parameters when called inside optimization(?)  


dritchie:
Currently, `registerParams` always lifts parameters to ad nodes before returning them. When sampling from a guide distribution with already-optimized params, this can lead to unintuitive behavior. Example:

``` js
var softplus = function(x) {
    return Math.log(Math.exp(x) + 1);
};

var model = function() {
    var mu = scalarParam(0, 0.1);
    var sigma = softplus(scalarParam(0, 0.1));

    var x = gaussian(mu, sigma);

    return {mu: mu, sigma: sigma};
};

console.log(sample(SampleGuide(model)));
```

This outputs

```
{ mu: 
   Node {
     x: -0.1707969210507051,
     parents: [ [Object] ],
     inputs: [ [Object], 0 ],
     backward: [Function: bw],
     outDegree: 0,
     name: 'tensorEtry',
     dx: 0 },
  sigma: 
   Node {
     x: 0.6690029642246522,
     parents: [ [Object] ],
     inputs: [ [Object] ],
     backward: [Function: bw],
     outDegree: 0,
     name: 'scalar.log',
     dx: 0 } }
```

Which is almost certainly not what the user wants to see.

Those in the know can just wrap these values in an `ad.value` and call it a day, but it'd be nice to make this just do the expected thing.

PR #708, opened 2016-11-15
Parameter store interface with in-memory and MongoDB implementations

stuhlmueller:
This PR introduces a global parameter store interface (#660), an in-memory and a MongoDB implementation, and command-line functionality for sharing parameters across processes/machines.

## Interface

Each parameter store is required to implement the following interface:

    start: () => ()
    stop: () => ()
    getParams: (id) => params
    setParams: (id, params) => ()
    incParams: (id, params, deltas) => newParams

`start` initializes the connection to the store and is executed before a WebPPL program runs. `stop` disconnects and is executed before a WebPPL program finishes. `getParams`, `setParams` and `incParams` take a parameter set `id` that decides which set of parameters to get/set/increment.

The contract for `incParams` is as follows: Both `params` and `deltas` are [param structs](https://github.com/probmods/webppl/blob/param-store/src/params/struct.js). That is, they are maps from parameter names to lists of tensors. If each of the parameter names in `deltas` is already present in the store, the `params` argument will be ignored and the deltas will be applied to the stored params. If one of the parameters is not yet present, the corresponding value given in `params` will be used to initialize it, and the delta will be applied to this newly-initialized value. (See discussion [here](https://github.com/probmods/webppl/issues/660#issuecomment-255773454).)

All of these functions are implemented in continuation-passing style to support stores that need to make asynchronous calls. That is, each of the functions above also takes the continuation function as its last argument and calls it with the return value.

## Implementations

The PR includes two implementations of this interface, an in-memory store that is used by default and that doesn't support parameter-sharing, and a MongoDB store that can be used to share parameters locally across processes or over the network.

### In-memory store

The in-memory store does a little more work than we did previously. There are two additional deep-copies [here](https://github.com/probmods/webppl/blob/param-store/src/params/store/memory.js#L43) and [here](https://github.com/probmods/webppl/blob/param-store/src/inference/optimize.js#L153). This results in a 20% slowdown for this [toy model](https://gist.github.com/stuhlmueller/98c6053b4fc1de49b977048f6e26dfc7). For [LDA](https://gist.github.com/stuhlmueller/f2b991b96fcba5c383ca04933a3ae547), the difference is maybe 10%, although I suspect that this particular result is just noise. (The runtime for LDA with the MongoDB store is close to the runtime for the old in-memory store, and if anything we'd expect MongoDB to be slower.)

The [deep-copy](https://github.com/probmods/webppl/blob/param-store/src/params/store/memory.js#L43) that is part of the memory–store could potentially be removed (or disabled based on a flag), but when I briefly experimented with this, I didn't see substantial gains.

### MongoDB store

The MongoDB store allows us to run multiple optimizers in parallel. See [here](https://github.com/probmods/carl/blob/master/figures/v2/expt-1-lda-parallel.pdf) and [here](https://github.com/probmods/carl/blob/master/figures/v2/expt-2-lda-parallel.pdf) for some early results on the effect of scaling the number of threads for LDA. To use this store, users need to have MongoDB running. By default, WebPPL will look for MongoDB at `localhost:27017` and use the collection `parameters`, but this can be changed by adjusting the environment variables `WEBPPL_MONGO_URL` and `WEBPPL_MONGO_COLLECTION`.

### Within WebPPL

For WebPPL users, the main change is that `Optimize` doesn't return its parameters anymore, and parameters are no longer passed into forward-sampling or SMC. Instead, the `guide`/`ignoreGuide` options (that were present before this PR and haven't changed) control whether the inference routine uses the current global parameters. (Compare to the change to the [`OptimizeThenSample`](https://github.com/probmods/webppl/blob/4cc8d9cbaaf382cd5be5cfe5f684a0632b08234d/src/header.wppl#L471) implementation: it is no longer necessary to return the parameters and pass them on to `SampleGuide`.)

I'm also making the following functions available from within WebPPL:

    getParams: () => params
    getParamsId: () => id
    setFreshParamsId: () => id
    setParams: (params) => ()
    setParamsId: (id) => ()

Using these functions, the previous capability to control where to use which parameters can be recovered.

### New command-line functionality

There are two new command-line options for the `webppl` script:

- By using `--param-store memory` or `--param-store mongo`, the user can control which store back-end to use. If no `--param-store` argument is given, the in-memory store is used.
- By using `--param-id 123`, the user can select which parameter set to use. Currently, this matters only for the MongoDB store that persists parameters across executions. If no such argument is given, a fresh id is generated.

I also added a `parallelRun` script to the `scripts` folder. This makes it easier to launch multiple webppl processes that all use the same parameter id. For example, to run ten processes that all execute `/path/to/model.wppl` with parameter id `my-params`, run:

    scripts/parallelRun /path/to/model.wppl 10 my-params

Any extra arguments are passed on to webppl, so this sort of thing works:

    scripts/parallelRun /path/to/model.wppl 10 my-params --require webppl-json

I haven't added the new command-line functionality to the docs yet. We probably want to test it a bit before we advertise it to the general public. We can also consider folding this functionality into the main webppl script.

## Other breaking changes

To support asynchronous stores, I'm making the trampoline runner [globally available](https://github.com/probmods/webppl/blob/param-store/src/main.js#L231). As a consequence, `prepare` doesn't need to return it anymore. This will require a small change to `webppl-editor`.


dritchie:
Wow, this is a really impressive effort!

The one high-level concern I have is how "future-proof" this setup is, as we think about implementing a proper, higher-performance tensor backend (perhaps backed by Torch). In that case, the in-memory implementation's deep copies could become very expensive when we're dealing with tensors that have hundreds of thousands of elements. I suppose we could have an 'in-place in-memory' implementation of the parameter store API, if we can find a way to get rid of any parameter struct deep copies that occur outside of parameter store code (namely, [this one](https://github.com/probmods/webppl/blob/param-store/src/inference/optimize.js#L153)). Along these lines: it seems like this could be done away with if we just had 'get' and 'set' for params. What's the gain from also having an 'incr' method?

The mongo implementation's serialization/deserialization would be even worse in the "big tensors from high-performance backend" setting (probably unusably so). There might be other ways to support distributed parameters, though (like the [parameter server](https://github.com/dmlc/ps-lite)).

Thoughts?

stuhlmueller:
> I suppose we could have an 'in-place in-memory' implementation of the parameter store API, if we can find a way to get rid of any parameter struct deep copies that occur outside of parameter store code (namely, [this one](https://github.com/probmods/webppl/blob/param-store/src/inference/optimize.js#L153)). Along these lines: it seems like this could be done away with if we just had 'get' and 'set' for params. What's the gain from also having an 'incr' method?

Yes, the current in-memory implementation can be made in-place simply by removing all the deep-copies in the store. (This requires us to be more careful about what we do to parameter structs returned from the store.)

Without `inc`, clients need to retrieve the latest parameters after each optimization step so that they can apply the gradient updates without overwriting what's in the store with outdated local params. The store seems to be in a better position to apply deltas, since it can queue them up and make sure that they all get applied, whereas simultaneous gets/sets from many clients can lead to work getting lost.

That said, the Mongo Store internally implements incremental updates using `get` and `set`, so currently the only benefit from having `inc` is that the code in `Optimize` doesn't have to implement the get/set logic.

I think you're right that using `get`/`set` directly instead of `inc` would let us remove the deep-copy in `Optimize`, so we'd recover the parameter computations as they were before this PR. I'm not certain, but my sense is that this would be somewhat better than the current version, so unless Paul or Noah feel otherwise, I'll go ahead and see if I this works out as expected.

> The mongo implementation's serialization/deserialization would be even worse in the "big tensors from high-performance backend" setting (probably unusably so).

It's only intended to work with applications that are roughly like the ones we've explored so far where, if you have big tensors, you generally have other computational costs that outweigh the (de)serialization cost.

> There might be other ways to support distributed parameters, though (like the parameter server).

Yeah, would be great if we could figure out a way to use this sort of system as a back-end for our parameters!

src/params/params.js
```diff
@@ -0,0 +1,146 @@
...
+// it directly returns the current local parameter copy.
+function get() {
+  return _params;
+}
+
+
+// Save the local parameter table to a file
+function save(filename) {
+  var s = JSON.stringify(serializeParams(_params));
+  fs.writeFileSync(options.checkpointParamsFilename, s);
```
null-a: `fs` is not required in this module, `options.checkpointParamsFilename` => `filename`?

null-a:
> I'm not certain, but my sense is that this would be somewhat better than the current version, so unless Paul or Noah feel otherwise, I'll go ahead and see if I this works out as expected.

I had mind that `inc` exists in order to support a particular approach to updating parameters @ngoodman had in mind? (I'm sketchy on the details, and may have the wrong end of the stick.)

src/params/params.js
stuhlmueller: Thanks, fixed.

null-a:
I spent a little more time with this. A few comments:

* I tried out the mongo store, and worked fine with one of my recurrent guide models.
* The only snag I hit getting it working was that after I had installed the mongodb package globally as suggested, I had to run `npm link mongodb` to get the `require` to work. Did you not need to do this?
* The `parallelRun` script is a little more picky about the format of the path to the program than the webppl script. e.g. `path/to/prog.wppl` doesn't work, but `./path/to/prog.wppl` does. One possible ugly fix is to redirect the output of `cd $(dirname $1)` to `/dev/null` in the `realpath` function. This seems to work (in my limited tested) but presumably someone who knows more bash than me could do better.
* In switching from `dev` to this branch I didn't see any appreciable difference in the run time of the program I tested. Switching from the memory store to mongo increased the run time by about a third. Running multiple processes worked, and FWIW just from eyeballing the output I'd guess it did yield a speed up.
* I previously left a TODO comment in one of the tests about #634. If you'd like to remove that, then it's fine by me. I intend to look at that issue once this PR is merged.

stuhlmueller:
I've replaced `inc` with `get`/`set` and introduced a `SAFE_MODE` that is off by default. When the memory store is used, performance should now be very close to what it was before the PR. As far as I can tell, the only relevant difference is that, when we retrieve the latest params from the store before we apply gradient updates, [we use `_.extendOwn`](https://github.com/probmods/webppl/blob/f3ca01301b2652c536ffbc9d0a44599188ea76e9/src/params/params.js#L39) instead of plain assignment of the `params` object. This is necessary, since plain assignment would remove any parameters that have been initialized locally in the most recent program run and that aren't present in the store yet.

> The only snag I hit getting it working was that after I had installed the mongodb package globally as suggested, I had to run `npm link mongodb` to get the `require` to work. Did you not need to do this?

I didn't have to do this previously (not sure why), but had to do it now on a new install. I've replaced the instructions with a message that suggests a local instead of a global install of `mongodb`. Feel free to replace it with global + link if that's better.

> The `parallelRun` script is a little more picky about the format of the path to the program than the webppl script. e.g. `path/to/prog.wppl` doesn't work, but `./path/to/prog.wppl` does.

I couldn't reproduce this on my install (running in Docker on Mac). If you want to edit the script to use the version suggested in your comment, or some other lenient version, that's fine with me.

> I previously left a TODO comment in one of the tests about #634. If you'd like to remove that, then it's fine by me.

Done.

This is ready to merge from my end now, unless anyone has further comments.

null-a:
Great, sounds good to go to me too.

PR #711, opened 2016-11-23
Add 'justSample' and 'onlyMAP' to SMC

dritchie:
This also stops MaxAggregator from returning a histogram-based distribution object (it instead uses a new 'MAP' distribution type).

This is necessary because the call to to `util.serialize` in `CountAggregator.toDist' can take *forever* for non-trivial objects (as in, multiple minutes, and then it sometimes crashes trying to allocate a string larger than what JSON.stringify can handle). I've hacked around this for graphics demos in the past, but I got fed up with it this time around and decided to actually fix it in core webppl.

### DIFF
--- a/docs/inference/methods.rst
+++ b/docs/inference/methods.rst
@@ -289,6 +289,22 @@ SMC
 
          Default: ``'MH'``
 
+      .. describe:: justSample
+
+         When ``true``, maintain an array of all samples taken. This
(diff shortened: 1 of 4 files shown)


dritchie:
Ah, looks like this broke some tests.

dritchie:
Closing this in favor of new PR Paul is about to submit.

PR #712, opened 2016-11-23
List of samples backed marginal

### LINKED ISSUE #631: Add a "list of samples" backed marginal distribution

null-a:
This adds a new "list of samples" backed marginal distribution (as discussed in #631), and uses it as the return value for all sampling based inference algorithms.

There's one interface change that comes along with this: The `justSample` option of `MCMC`/`IncrementalMH` no longer exists, since marginals returned by sampling algorithms *always* have the `samples` property available. (This just returns the list of samples that backs the marginal.)

The `onlyMAP` option still exists, and works as before. Following Daniel, I've also added this option to `SMC`.

I see two possible objections to this change as it stands:

1. There are some scenarios where building the histogram as we go might be better strategy. For example, if you are taking many, many, samples but are only sampling from a small set of simple values (e.g Bools) then building the histogram as you go saves keeping a giant array in memory. I'm not certain whether this is likely to be a problem for us in practice though? I've tested sampling `1e6` coin flips using `MCMC`, and this change doesn't noticeably change the run time.

2. The `samples` property is an array of objects like `{value: ..., score: ...}`. This format is inherited from the old `justSample` option. At present, not all algorithms compute the joint score of a sample, so for those I currently set the `score` property of these objects to zero. This seems messy, so we might want to compute the score everywhere, or if there's no value in that perhaps we just omit the `score` property when we don't have it.

Closes #631.


dritchie:
This makes me very, very happy. That's it, just wanted to say that :)

src/inference/smc.js
```diff
@@ -87,6 +89,8 @@ module.exports = function(env) {
     // Optimization: Choices are not required for PF without rejuvenation.
     if (this.performRejuv || this.saveTraces) {
       particle.trace.addChoice(dist, val, a, s, k, options);
+    } else if (this.onlyMAP) {
+      particle.trace.score = ad.scalar.add(particle.trace.score, dist.score(val));
```
stuhlmueller: Why do this only when `this.onlyMAP` is `true`?

src/aggregation/CountAggregator.js
```diff
@@ -5,35 +5,30 @@ var util = require('../util');
...
+  this.max = {value: undefined, score: -Infinity};
+  this.samples = [];
 };
 
-CountAggregator.prototype.add = function(value) {
-  var k = util.serialize(value);
-  if (this.hist[k] === undefined) {
-    this.hist[k] = { count: 0, val: value };
+CountAggregator.prototype.add = function(value, score) {
+  score = (score === undefined) ? 0 : score;
```
stuhlmueller: I think I'd go with your suggestion to omit the score property when we don't have it. (Can happen in a follow-up PR if it's not a quick change.)

src/dists.ad.js
```diff
@@ -1322,6 +1322,10 @@ var Marginal = makeDistributionType({
     this.supp = _.map(this.params.dist, function(obj) {
       return obj.val;
     });
+
+    this.getDist = function() {
```
stuhlmueller: I was a little confused by the name `getDist` at first. I assumed that, following our new-ish nomenclature, `dist` here is an instance of our distribution type -- the thing formerly known as `erp`. It seems desirable to reserve the word `dist` in the codebase for this type wherever possible. Is there a clearer name for what happens here? Maybe `toObject`?

That said, I see that the distribution parameter is also called `dist`, which may have motivated this method name. There, it's less clear that there is a good replacement, and then there are a bunch of other places (`printDist`, `_cacheddist `) where it's unclear what the alternative would be, and having inconsistent names may be worse than the status quo? I don't know. Feel free to leave it as it is.

src/dists.ad.js
```diff
@@ -1347,14 +1351,90 @@ var Marginal = makeDistributionType({
...
+    this.samples = this.params.samples;
+
+    this.getDist = function() {
+      if (this._cacheddist) {
+        return this._cacheddist;
+      } else {
+        var dist = {};
+        this.params.samples.forEach(function(obj) {
+          var val = obj.value;
+          var k = util.serialize(val);
```
stuhlmueller: Use a different letter? I don't think there's much danger of confusion here, but since `k` is almost exclusively used for continuations, maybe other uses are worth avoiding.

src/dists.ad.js
```diff
@@ -1347,14 +1351,90 @@ var Marginal = makeDistributionType({
...
+
+    this.getSupport = function() {
+      if (this.params.samples.length === 1) {
+        // Optimization: Avoid unnecessary serialization in the
+        // onlyMAP case
+        return [this.params.samples[0].value];
+      } else if (this._cachedsupport) {
+        return this._cachedsupport;
+      } else {
+        var support = _.map(this.getDist(), _.property('val'));
```
stuhlmueller: Nice! I didn't know `_.property` :-)

src/dists.ad.js
```diff
@@ -1347,14 +1351,90 @@ var Marginal = makeDistributionType({
...
+    var n = this.params.samples.length;
+    return this.params.samples[Math.floor(util.random() * n)].value;
+  },
+  score: function(val) {
+    var k = util.serialize(val);
+    var obj = this.getDist()[k];
+    return (obj === undefined) ? -Infinity : Math.log(obj.prob);
+  },
+  support: function() {
+    return this.getSupport();
```
stuhlmueller: What's the reason for having a separate `getSupport` function (as opposed to inlining the code there into the `support` method)?

src/inference/smc.js
null-a: My thinking was that we should only incur the cost of computing the score when doing so is necessary. However, we already compute the score earlier in the `sample` method, and we should just re-use that of course -- then we'll always be able to include the joint score in the bag of samples. Thanks. I'll fix this shortly.

src/dists.ad.js
null-a: No idea! Maybe it made sense at some point, but inlining it looks like a good idea to me. Thanks!

src/aggregation/CountAggregator.js
null-a: Great, thanks. Will do as part of this PR.

src/dists.ad.js
stuhlmueller: Sounds good. I think `hist` isn't clearly enough of an improvement that it's worth switching over now, but it's worth keeping in mind when we write future code that involves these concepts.

Issue #691, opened 2016-11-02
webppl.org redesign/re-branding?


dritchie:
I like the simplicity and friendliness of the current webppl.org design (and the interactive code editor front-and-center is awesome). But I also feel like the current design might be under-selling what WebPPL is capable of, especially the description of it as "a small but feature-rich" language. That may have been true of the original dippl-era implementation, but I'd argue that it's not really "small" anymore. Basically, I think that the current website presents the language as more 'toy' than it really is.

Some suggestions for low-effort changes that I think would improve WebPPL's public image:
 - New tagline (i.e. instead of "small but feature-rich")
 - Small section emphasizing what features the language has. Could include the inference algorithms it supports, nested inference, etc.
 - Small section playing up other nice properties of the language, like that you can run it in either the browser or node, and that it plays nicely with existing libraries in the excellent npm ecosystem.

Also open to other suggestions for what people would like to see!

longouyang:
I think having more impressive interactive demos would be really helpful, e.g., the inverse physics example from probmods or a procedural graphics example.

stuhlmueller:
We could also add that it now supports both inference and learning, and show some examples.

ngoodman:
i think this is a great idea. any volunteers to make it happen?

adding a couple of really cool examples (either on the main page or on separate examples pages) would help a lot....

stuhlmueller:
I'm happy to help with this, especially the part where we need to adapt the current html/css/js. Would be great to do examples and other new content collaboratively!

ngoodman:
maybe we can brainstorm (here or elsewhere) some cool examples we could do?

-some kind of procedural graphics
-captcha breaking
-some kind of natural language understanding (pragmatics, semantic parsing, puns?)
-VAE variants on mnist
...

stuhlmueller:
- agents - planning as inference
- lda
- some example that uses data generated in the browser; maybe something [like this](https://stuhlmueller.org/smartpages/)

dritchie:
Probably wouldn't be too hard to get the vines program from the NIPS paper working on the website.

As for the VAE idea: IMO, this is currently probably too slow to train in-browser. But now that I think about it, it might be neat to send already-trained parameters to the client, and then run the learned model / inference network in-browser.

stuhlmueller:
Would be cool to do vines on the page content, maybe using [html2canvas](http://html2canvas.hertzen.com/examples.html) first to get an image of the page.

longouyang:
[...]

PR #714, opened 2016-11-27
Updated webppl.org with demos section

dritchie:
Here's my first crack at updating webppl.org. New content includes:

- List of language features
- List of publications using webppl
- List of demos (currently: 2D vines/lightning from NGPM paper, old procedural spaceships demo from SIGGRAPH 2015)
- Jumbotron header points to demos

Addresses (but doesn't yet close?) #691 

Anyone have any changes they'd like to see made before this gets merged in and goes live? I'd like to do that by Tuesday morning, if possible (some Adobe people are coming to Stanford, and it'd be good to have this up and running).


ngoodman:
neat --how can i preview the new site?

dritchie:
Here's how to preview:

- Check out the `gh-pages-redesign` branch
- From the root directory of the branch, run `python -m SimpleHTTPServer` (the demo makes some AJAX requests, so it needs to be run under a webserver).
- Point your browser to `localhost:8000`

ngoodman:
nice additions! happy to have this go live today....

some suggestions:

*  it looks a bit awkward to have the "features" live in boxes, but nothing else. maybe loose the boxes but have bullets for the features?

* likewise it's not currently clear why some content is two column and other is one (and why these are intermixed)

* in "documentation" we should perhaps now put the actual docs first, and then have dippl / probmods. also we should point to probmods/v2 (which is actually in webppl).

* in publications we can list the various web books and notes that use webppl: probmods/v2, dippl, agentmodels.org, http://gscontras.github.io/ESSLLI-2016/, maybe others eg MH's BDA class?

* it would be good for the how to cite info to be a top level section separate from the publications...

dritchie:
Done. I'll merge this now. Let me know if anything else seems amiss (e.g. @stuhlmueller, did I get the authors right for agentmodels.org?)

Issue #710, opened 2016-11-21
Neural Net Helpers


null-a:
(I'm splitting this out from #663.)

Idea: by leveraging our parameter/guide infrastructure we can use function composition to build simple neural networks in webppl. This would look something like this:

```js
// make a net:
var net = compose(sigmoid, linear(10, 'name'));
// use it:
net(ones([5,1]));
// slightly fuller example:
var predictNet = stack([linear(10, 'net1'), tanh, linear(2, 'net2'), sigmoid]);
var guideParams = {ps: predict(x)};
// etc.
```

Under the hood, the parameter infrastructure takes care of doing the right thing with parameters, and we determine whether we need to put a prior over parameters by checking whether we're in a guide thunk. (By building on #652.) ~~There's a work-in-progress implementation of this [here](https://github.com/probmods/webppl/compare/dev...null-a:wppl-nets).~~

For the user, the main advantage of this over using adnn nets is that there's no need to use `nneval` or call `setTraining` on the net. Also, with this approach, it's possible to mimic the common pattern of `cache`ing a fragment of network by simply reusing network names where sharing is required.

One question I'd like to answer (in part to help convince myself that this is flexible enough to handle future requirements) is how we best handle weight decay with this scheme. You can imagine specifying weight decay with something like `linear({..., weightDecay: 'l2'})`, and for nets that are part of the model we would implement this by switching from using `ImproperUniform` to `TensorGaussian` as the prior. But for nets that are part of the guide we don't have this option, so we need to do something else. (@dritchie and I have [discussed](https://gist.github.com/null-a/a42a855efbce27da1f8c4f4248ddd790) this a little already.)

dritchie:
Re: weight decay for for guide nets: can we not just use the 'prior swapping' trick? I realize that this would mean that we'd be using a "guide inside of a guide," but there's no reason this wouldn't work, right? Currently (i.e. we haven't switched to using guide thunks yet), there's no notion of being 'inside' a guide, so something like this would work:
```js
var data = // some tensor

var guideParamWithPrior = function(data) {
   var W = sample(TensorGaussian({ mu: 0, sigma: 1, dims: [1, data.length] }), {
      guide: Delta({ v: param({ name: 'W', dims: [1, data.length]) })
   });
   var b = sample(Gaussian({mu: 0, sigma: 1}), {
      guide: Delta({ v: param({ name: 'b' })
   };
   return T.sigmoid(T.get(T.dot(W, data), 0) + b);
};

[...]

PR #715, opened 2016-11-28
Guide Thunks

### LINKED ISSUE #652: Compute guide distributions lazily / on-demand?
Currently, guide distributions are always computed, even if the program is running under a coroutine that does not use guides. When the guide construction is compute-intensive (e.g. involves big neural nets), this seems wasteful.

It might be better to encapsulate the guide in a thunk which is only evaluated when the guide is actually needed.

### LINKED ISSUE #705: Allow auto mean field guide to be used as importance dist in SMC

null-a:
This switches us to the "guides as thunks" interface suggested in #652.

This will require all code that includes a "hand written" guides to be updated. Passing a distribution directly rather than a function will generate the following error after this change: `Error: The guide is expected to be a function.`

The dream learning implementation will need updating inline with this, but I expect this to be a small change. It's also likely that this touches some of the same code as the parameter store PR (#708), so perhaps we merge that first, then I can update this PR.

In addition to avoiding evaluating the guide when it's not needed for inference, after this change we also:

1. Show a warning message when `param` is called outside of a guide thunk.
2. Throw an error if `sample` or `factor` are used directly within a guide thunk. (Though you can call `Infer` in a guide, from within which `sample` and `factor` work as normal.)

The motivation for the second of these is that in my way of looking at this, whether or not the guide is evaluated is an implementation detail of each inference algorithm, and as such shouldn't affect the structure of the model. Having used this a little, it also feels like this extra structure is helpful when thinking about what a guided model means, and for catching errors. However, this change does make a pattern that @dritchie [suggested](https://github.com/probmods/webppl/issues/710#issuecomment-262038916) rather awkward (at best), so that's something we need to talk about before this is merged. (I'm happy with this direction, but there might be use cases I'm missing.)

This PR also addresses #705. As part of that, I've added a new `importance` option to SMC to control *when* mean-field guides should be automatically added to random choices. This is described [here](https://github.com/probmods/webppl/commit/9dfdef561149204d726f4cf021d5afc3f506ee51). I've had some feedback that the values this takes aren't very intuitive -- if anyone has better suggestions I'd be happy to change this.

Closes #652.
Closes #705


dritchie:
I agree that this is a Good Idea, and that we should figure out how to work around it re: #710. BTW, I now see why what I suggested there won't work, as `sample` is forbidden inside guide thunks (otherwise the one-to-one correspondence between choices in the guide and in the target wouldn't hold).

One other possible complication: there might be scenarios where the guides for multiple random choices can naturally share computation, which will be made trickier by this change. I suppose different guide thunks can still communicate via `globalStore`. Any other thoughts about how to handle this?

null-a:
@dritchie I think I over stated how awkward this makes your suggestion from #710, since we can make any existing code work under this PR with the following simple transform:

```js
// before thunks
sample(dist, {guide: guideDist})

// after thunks
sample(dist, {guide: constF(guideDist)})
```

> there might be scenarios where the guides for multiple random choices can naturally share computation, which will be made trickier by this change.

By transforming as above, it might be messy, but not too tricky, right? The hope in making this change of course, is that we could factor out some or all of the non-shared bits and stick them in the thunk, but the extent to which that is possible is presumably problem specific.

If you have a specific example in mind, maybe we could take a look at what that looks like after this change? If not, perhaps we just do as well as we can here with what we know now, and adjust course as we learn.

dritchie:
I don't have any specifics in mind. I think it'd be fine to move forward with things as-is and adjust as necessary in the future.

PR #716, opened 2016-11-28
Fix bug recovering stack.

null-a:
This fixes a bug that I ran into when working on #715. Without this, the error message shown when a guide thunk doesn't return a distribution object doesn't include the most recent position in user code.

### DIFF
--- a/tests/test-stack-trace.js
+++ b/tests/test-stack-trace.js
@@ -137,10 +137,14 @@ var testDefs = [
   },
 
   { name: 'top most webppl entry is not a user function',
-    code: ['var f = function() { stackTraceTestFns.test() };',
+    code: ['var testFn = stackTraceTestFns.test;',
+           'var f = function() { return testFn(); };',
            'f();'
     ].join('\n'),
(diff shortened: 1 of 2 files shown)


tests/test-stack-trace.js
```diff
@@ -124,6 +136,14 @@ var testDefs = [
     limit: 1
   },
 
+  { name: 'top most webppl entry is not a user function',
+    code: ['var f = function() { stackTraceTestFns.test() };',
```
stuhlmueller: I don't follow how this test works. `stackTraceTestFns.test` is called as a method, so -- due to the way we wrap external functions -- I expect `s, k, a` above to be all `undefined`, and the inner function in `test` never to be called. Is that expected behavior, or am I missing something?

null-a: No, you're right, I messed up! That test function used to be in headerUtils where `s,k,a` made sense, but when I switched to temporarily attaching it to `global` I broke things. Actually, I'm now not entirely sure it was right before that, although as written the test did fail before the fix, and pass after. Either way, I'll push a fix shortly. Thanks!

PR #718, opened 2016-12-01
Push remaining printing logic into dist objects.

### LINKED ISSUE #698: Marginal distributions shouldn't show their internal structure when printed

null-a:
This is a minor change to the way marginals are printed under node. It removes the last remaining bit of special casing for marginal dists, from the code that prints the value of the last expression in a program. The upshot is that `var m = Infer({...}); display(m)` now prints the same thing as just doing `var m = Infer({...}); m`. Previously, the former printed the internal data structures backing the marginal, rather than the pretty output generated by the latter.

Closes #698.

### DIFF
--- a/src/dists.ad.js
+++ b/src/dists.ad.js
@@ -52,7 +52,11 @@ Distribution.prototype = {
 
   inspect: function(depth, options) {
     if (_.has(this, 'params')) {
-      return [this.meta.name, '(', inspect(this.params), ')'].join('');
+      if (this.print) {
+        return this.print();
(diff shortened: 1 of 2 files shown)


stuhlmueller:
Nice! Should we get rid of the `printWebPPLValue` function given that it only calls `console.log`?

null-a:
Oh, you're right. When I looked at that I thought we lost a useful abstraction by doing so, but since `topK` is used by both `run` and `compile` we don't. Thanks, done!

PR #722, opened 2016-12-03
Add `extend` to header

stuhlmueller:
We use this extremely commonly as part of AgentModels (more than 50 times, currently named `update`).

### DIFF
--- a/docs/functions/header.rst
+++ b/docs/functions/header.rst
@@ -29,6 +29,19 @@ Header
       var pair = function(x, y) { return [x, y]; };
       mapObject(pair, {a: 1, b: 2}); // => {a: ['a', 1], b: ['b', 2]}
 
+.. js:function:: extend(obj1, obj2, ...)
+
+   Creates a new object and assigns own enumerable string-keyed properties
+   of source objects 1, 2, ... to it. Source objects are applied from left
+   to right. Subsequent sources overwrite property assignments of previous
+   sources.
+
+   ::
+
+      var x = { a: 1, b: 2 };
+      var y = { b: 3, c: 4 };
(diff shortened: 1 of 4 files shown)


longouyang:
[underscore's extend](http://underscorejs.org/#extend) takes an arbitrary number of objects as an argument -- would it be easy to make your implementation behave this way?

stuhlmueller:
It's definitely possible:

~~~~
var extend = function() {
  return _.extend.apply(_, [{}].concat(arguments));
}
~~~~

The plain version is [10x faster](https://jsperf.com/arguments-vs-not) in plain Javascript.

One could also do this:

~~~~
var extend = function(a, b, c, d, e, f, g, h, i, j) {
  return _.extend({}, a, b, c, d, e, f, g, h, i, j);
}
~~~~

The plain version is [4x faster](https://jsperf.com/two-args-vs-many) than this one.

longouyang:
Would you be on board with writing it the slower but most flexible way? (using `_.extend.apply`)

Also:
- For me (Chrome 55), the plain version is only 5x as fast (2.8 million ops / sec) as the most flexible version (550k ops / sec).
- The most flexible version clearly works in Javascript, but would it work in webppl?

stuhlmueller:
> Would you be on board with rewriting your PR in the most flexible way? (using `_.extend.apply`)

Done.

> The most flexible version clearly works in Javascript, but would it work in webppl?

Yes.

null-a:
This seems fine to me. Obligatory mention of tests, docs?

null-a:
A thought crossed my mind reading the docs... For 'own enumerable' should this be implemented with `extendOwn`/`assign` rather than `extend`?

PR #736, opened 2016-12-19
Warn when forward sampling hits factor/clarify docs.

### LINKED ISSUE #731: Warn user when method="forward" used w/ model with observes
This seems to be a common misconception when learning how to use `Infer()`, that "forward" means "forward sample (ignoring observe, condition, factor)"

Has come up repeatedly when teaching WebPPL. [Here also](https://groups.google.com/forum/#!topic/webppl-dev/vxIOoWZnD7E)

null-a:
This is an attempt to prevent confusion about the behavior of forward sampling by issuing a warning (as suggested by @mhtess) and by clarifying the docs.

Closes #731.

### DIFF
--- a/src/inference/forwardSample.js
+++ b/src/inference/forwardSample.js
@@ -23,6 +23,8 @@ module.exports = function(env) {
     this.k = k;
     this.a = a;
 
+    this.factorWarningIssued = false;
+
     this.coroutine = env.coroutine;
     env.coroutine = this;
   }
@@ -68,6 +70,12 @@ module.exports = function(env) {
     },
 
     factor: function(s, k, a, score) {
+      if (!this.opts.guide && !this.factorWarningIssued) {
+        this.factorWarningIssued = true;
+        var msg = 'Note that factor, condition and observe statements are ' +
+            'ignored when forward sampling from a model.';
+        util.warn(msg);
+      }
       this.logWeight += ad.value(score);
       return k(s);
     },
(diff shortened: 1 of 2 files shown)


src/inference/forwardSample.js
```diff
@@ -68,6 +70,12 @@ module.exports = function(env) {
     },
 
     factor: function(s, k, a, score) {
+      if (!this.opts.guide && !this.factorWarningIssued) {
```
stuhlmueller: Why don't we issue the warning if a guide is present?

null-a: I think of the guide program as been the thing you get if you sample from guide distributions at `sample` statements and ignore `factor`. From this perspective, sampling from the guide doesn't include `factor`s by definition, and therefore `forward` isn't doing anything out of the ordinary that we need to warn people about.

stuhlmueller: Makes sense.

There's an alternative view where specifying a guide only replaces the prior when sampling, and you're otherwise still doing inference as usual (but reweighting the samples to account for the fact that you're importance sampling in one form or another). For example, you could first fit a guide program and then use it instead of the prior as the basic proposal distribution in MCMC, which obviously wouldn't ignore factors. Your view is a better fit for how guide programs are currently used in WebPPL, but I think the alternative has merits as well, and if we went down that route, I think we'd want to issue warnings in `forward` even if a guide is specified.

null-a: Hmm.. I understand that use of a guide, but quite can't manage to see why that's a good reason to show a warning here. If you're happy to go ahead with this as it, we maybe resolve this some other time?

One thing that might be worth clarifying:

> Why don't we issue the warning if a guide is present?

The condition isn't whether there's a guide present, it's whether the `{guide: true}` argument is passed to `Infer`. (There's kinda *always* a guide present -- either the one written explicitly in the program, or the mean-field thing we give you otherwise.)

Issue #530, opened 2016-06-24
Use lodash instead of underscore?


stuhlmueller:
See [here](http://stackoverflow.com/questions/13789618/differences-between-lodash-and-underscore) for a comparison.

jpchen:
There's been some talk of merging the two: https://github.com/jashkenas/underscore/issues/2182

null-a:
Closed by #737.

PR #737, opened 2016-12-20
Change underscore dependency to lodash

jpchen:
Addresses #530. Will need an `npm install` to work locally.

Tests:
Passes all existing tests.  

Ran into some merging issues and couldn't squash locally, feel free to squash before merging.


src/header.wppl
```diff
@@ -531,7 +531,7 @@ var Vector = function(arr) {
 var Matrix = function(arr) {
   var n = arr.length;
   var m = arr[0].length;
-  var t = ad.tensor.fromScalars(_.flatten(arr));
+  var t = ad.tensor.fromScalars(_.flattenDeep(arr));
```
stuhlmueller: Do we only need `_.flatten` here? I think the assumption is that the input array is a list of lists, e.g. `[[1, 2], [3, 4]]`.

src/util.js
```diff
@@ -271,7 +271,7 @@ function timeif(bool, name, thunk) {
 }
 
 function pipeline(fns) {
-  return _.compose.apply(null, fns.reverse());
+  return _.flowRight.apply(null, fns.reverse());
```
stuhlmueller: Would just `_.flow(fns)` be equivalent to this?

(I'm a little surprised to see that the current `pipeline` is mutating the `fns` argument, but I suppose `pipeline` is only used on arrays that are created for the sole purpose of being passed to it.)

jpchen: `_.flow` invokes functions left to right, whereas `_.compose`and `_.flowRight` is a normal function composition you'd expect where the last argument is the innermost fn, etc. A bit confusing, take a look here: https://lodash.com/docs/4.17.2#flow

src/header.wppl
jpchen: This was to be paranoid/safe when migrating. `flatten` in underscore is deep unless a shallow argument passed, while lodash breaks them up into two functions, `flatten` which is shallow and `flattenDeep` which is deep.  I guess if the matrix is always 2D, I can change it.

stuhlmueller: I think the matrix is always 2D, and that something else has gone wrong if it's not. (Paul, correct me if I'm wrong.) In that case, flattening only one step could make errors easier to find since `fromScalars` will hopefully throw an error if it gets a list as input.

src/util.js
stuhlmueller: But leaving out `reverse` as suggested would recover the intended order, no?

null-a:
Am I correct in thinking this will break any webppl programs that rely on functions present in underscore but not in lodash? (e.g. `_.object`.)

stuhlmueller:
Yes, such programs will have to be updated. I don't think this will be very common, the upgrade is easy, and the next release will include a number of breaking changes anyway. We can point people to the [underscore->lodash migration page](https://github.com/lodash/lodash/wiki/Migrating) in our release notes.

src/header.wppl
jpchen: Done.

src/util.js
jpchen: Ah yes my apologies, I misread your comment. Fixed.

jpchen:
Yes, I also hastily whipped up cross documentation between the two libraries, for ease of future use: jonathanpchen.com/underdash-api
You can look up the equivalent lodash and underscore methods.

It's incomplete and based off of what I used above and other functions I looked into, but I plan to update as I go.

src/util.js
```diff
@@ -271,7 +271,7 @@ function timeif(bool, name, thunk) {
 }
 
 function pipeline(fns) {
-  return _.flowRight.apply(null, fns.reverse());
+  return _.flow.apply(null, fns);
```
stuhlmueller: I think `_.flow(fns)` would be equivalent for our purposes, since `_.flow` can take arrays as well. But at that point, it's not clear that we even need the `pipeline` function, since `_.flow` is fewer characters than `util.pipeline`.

scripts/distributionDocs
```diff
@@ -1,14 +1,14 @@
-#!/usr/bin/env node
+#!usr/bin/env node
```
null-a: Is this change intentional?

src/util.js
null-a: I added `pipeline` because there wasn't an equivalent in underscore, but yeah, sounds like we can get rid of it now. (At some point.)

scripts/distributionDocs
jpchen: Fixed. Must've been something in my bash settings when I ran build...

src/util.js
jpchen: Done.

PR #740, opened 2016-12-24
Support reflection for static analysis

longouyang:
I'm starting to work on static analysis tools for webppl (e.g., `viz.model` for visualizing dependency diagrams of generative models, making `viz.auto` smarter wrt the actual supports of a distribution) and it's *far* easier to do this if I have access to both the original code, compiled code, and source map at run-time.

Example: `viz.model` ingests a function `f`, analyzes its source code via `f.toString()`, and visualizes the result. Except `f.toString()` is pretty hard to work with, as it has gone through a bunch of transforms. But.. it turns out that I can back out what `f`'s original source code was if I have access to the source map et al. This PR installs that information as global variables.

FWIW, I have hunch about a much cleaner way of supporting this sort of static analysis but I think it would be much more invasive, so this PR seems like a decent approach for the time being. (The idea: I think we can get away with doing code transforms at inference-time rather than compile-time if we change the signatures from `(s, k, a, x, ...)` to `(x, ..., s, k, a)`, so that any header functions would work both in JS and webppl)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -189,7 +189,10 @@ function compile(code, options) {
     ])(asts);
   };
 
-  return util.timeif(options.verbose, 'compile', _compile);
+  var r = util.timeif(options.verbose, 'compile', _compile);
+  global.__sourceMap__ = r.sourceMap;
+  global.__compiled__ = r.code;
+  return r;
 }
 
 function wrapWithHandler(f, handler) {
@@ -261,6 +264,10 @@ function prepare(codeAndAssets, k, options) {
 function run(code, k, options) {
   options = options || {};
   var codeAndAssets = compile(code, options);
+  global.__compiled__ = codeAndAssets.code;
+  if (_.has(codeAndAssets, 'sourceMap')) {
+    global.__sourceMap__ = codeAndAssets.sourceMap;
+  }
   util.timeif(options.verbose, 'run', prepare(codeAndAssets, k, options).run);
 }
 
--- a/src/main.js
+++ b/src/main.js
@@ -190,6 +190,7 @@ function compile(code, options) {
   };
 
   var r = util.timeif(options.verbose, 'compile', _compile);
+  // provide compiled code and sourceMap to support reflection
   global.__sourceMap__ = r.sourceMap;
   global.__compiled__ = r.code;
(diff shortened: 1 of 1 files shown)


longouyang:
(nudge)

null-a:
Assuming this is something we need to add, it seems likely to be worth spending at least a little effort trying to implement it without introducing extra global variables.

I'm not sure how your "cleaner but more invasive" approach would work, but it IIUC it would be a drastic change and not something we're likely to entertain in the short term?

The only practical alternative I can see, is to wrap the compiled program with a function and pass in the code and source map when the program is evaluated. We already do this for the runner and the object that tracks the current stack address, for much the same reason.

This would make the things you're interested in available throughout the webppl program, but not globally from JS. To get at these from JS code, I think you'd have to have some webppl entry point (e.g. `viz`) which grabbed them and passed them on. I don't know whether this would fit with the webppl-viz interface you have in mind for this?

What do you think?

(BTW, if we wanted to try this, I'd be happy to implement it if that helps.)

longouyang:
> Assuming this is something we need to add, it seems likely to be worth spending at least a little effort trying to implement it without introducing extra global variables.

I'm already using this functionality in a fork of webppl to make `viz.auto` handle more components in a posterior (it shows independent components in different graphs, rather than trying to cram every component into a single graph). So I think this reflection functionality would be nice to have in the canonical repo.

> I'm not sure how your "cleaner but more invasive" approach would work, but it IIUC it would be a drastic change and not something we're likely to entertain in the short term?

Indeed. I only brought it up because, at the time, this more drastic change was the only alternative approach I could envision. But the approach you suggested is sensible.

> The only practical alternative I can see, is to wrap the compiled program with a function and pass in the code and source map when the program is evaluated. We already do this for the runner and the object that tracks the current stack address, for much the same reason.
> 
> This would make the things you're interested in available throughout the webppl program, but not globally from JS. To get at these from JS code, I think you'd have to have some webppl entry point (e.g. viz) which grabbed them and passed them on. I don't know whether this would fit with the webppl-viz interface you have in mind for this?

Yes, thanks -- this makes sense. I'll refactor to avoid globals.

Issue #663, opened 2016-10-05
Add nneval


null-a:
This only exists in daipp at present.

ngoodman:
worth adding now as-is.

also worth figuring out how to make nn declaration and use more consistent with direct param use.

null-a:
> also worth figuring out how to make nn declaration and use more consistent with direct param use.

There's a pattern I've started to use that might suggest one possible way to approach this -- I've [described it here](https://gist.github.com/null-a/a42a855efbce27da1f8c4f4248ddd790). I'm not sure this hits the mark, but I figure it's worth sharing at least.

null-a:
After chatting with @stuhlmueller (Noah's on board too) it was agreed that we'll not add this (#752), and instead work towards standardizing on #710.

PR #747, opened 2017-01-05
Improve `params.register()`

### LINKED ISSUE #634: `registerParams` should only lift parameters when called inside optimization(?)

null-a:
This PR makes a few changes to the `params.register` method:

1. The `setParams` callback has been removed.
2. The function passed as the `getParams` argument is now expected to return unlifted parameters.
3. Parameters are only lifted when the current coroutine is performing optimization.
4. Improve comments.

Changes 1 & 2 don't affect core webppl, but they will break the implementation of `nnEval` in webppl-daipp. This will be addressed with an immediate follow-up PR that adds `nnEval` to webppl. (#663.)

Change 3 closes #634.

These changes are spread across a number of commits, which hopefully makes it easier to see that these are the only changes made. This could be squished down at merge if you like.

### DIFF
--- a/src/params/params.js
+++ b/src/params/params.js
@@ -73,10 +73,7 @@ function set(params, k) {
 }
 
 
-function register(env, name, getParams) {
-
-  // getParams is expected to be a function which is used to
-  // initialize parameters the first time they are encountered.
+function register(env, name, initParams) {
 
   var paramTable = get();
   var paramsSeen = env.coroutine.paramsSeen;
@@ -98,9 +95,9 @@ function register(env, name, getParams) {
       _params = paramTable[name];
     } else {
       // Never seen. Fetch initial values and add to store.
-      _params = getParams();
+      _params = initParams();
       assert.ok(_.every(_params, _.negate(ad.isLifted)),
-                'getParams unexpectedly returned a lifted value.');
+                'initParams unexpectedly returned a lifted value.');
       paramTable[name] = _params;
     }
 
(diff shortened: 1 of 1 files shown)


stuhlmueller:
Looks good. 

Checking for the presence of `paramsSeen` seems like a somewhat indirect way to determine whether we should return lifted params, but I think the two conditions do always coincide in our usage and we probably don't want to use an explicit argument instead, so seems OK for now.

PR #750, opened 2017-01-07
Add marginalization helper

### LINKED ISSUE #614: Add a marginalize helper function
It's often very helpful to project out individual components of a posterior

longouyang:
Would close #614 

I changed my mind about the naming again -- I think `marginalize` would be a little better than `mapDistribution`; the latter might convey the impression that the output distribution is isomorphic to the input.


stuhlmueller:
Two thoughts:

1. Would it make sense to extend this so that `project` can also be a string, in which case we perform a key lookup on the return value of `sample(dist)`? In many cases multivariate distributions are implemented as objects, so this would simplify marginalization quite a bit.

2. Add a test?

longouyang:
Okay, I made `project` work as a string and added a test and docs

stuhlmueller:
Nice, thanks!

Issue #733, opened 2016-12-19
Fix occasional syntax-related errors on travis


stuhlmueller:
Example:

~~~~
>> Error: [{name: globalStore, loc: null, type: Identifier, comments: null, typeAnnotation: null}, {"name":"_k32","loc":null,"type":"Identifier","comments":null,"typeAnnotation":null}, {callee: [object Object], arguments: [object Object], loc: null, type: CallExpression, comments: null}, {loc: [object Object], type: Identifier, name: wpplFn}, {properties: [object Object],[object Object], loc: null, type: ObjectExpression, comments: null}] does not match field "arguments": [Expression | SpreadElement] of type CallExpression
>> at add (node_modules/ast-types/lib/types.js:576:27)
>> at /home/travis/build/probmods/webppl/node_modules/ast-types/lib/types.js:589:17
>> at Array.forEach (native)
>> at Object.value [as callExpression] (node_modules/ast-types/lib/types.js:588:30)
>> at Controller.store (src/transforms/store.js:30:22)
>> at Controller.__execute (node_modules/estraverse/estraverse.js:397:31)
>> at Controller.replace (node_modules/estraverse/estraverse.js:618:31)
>> at Object.replace (node_modules/estraverse/estraverse.js:718:27)
>> at storeMain (src/transforms/store.js:41:21)
>> at /home/travis/build/probmods/webppl/node_modules/underscore/underscore.js:871:36
>> at /home/travis/build/probmods/webppl/node_modules/underscore/underscore.js:871:36
>> at _compile (src/main.js:189:7)
>> at Object.timeif (src/util.js:270:37)
>> at compile (src/main.js:192:15)
>> at Object.run (src/main.js:259:23)
>> at wpplRunInference (tests/test-inference.js:636:12)
~~~~

null-a:
Also:
https://travis-ci.org/probmods/webppl/jobs/193831121

null-a:
From #814:

https://travis-ci.org/probmods/webppl/jobs/219859137

null-a:
Possibly related:

https://travis-ci.org/probmods/webppl/jobs/221409602

PR #752, opened 2017-01-09
Add nnEval & nnEvalModel.

### LINKED ISSUE #663: Add nneval

null-a:
As promised in #747.

Closes #663.


null-a:
The [CI failure](https://travis-ci.org/probmods/webppl/jobs/190318252) looks similar to #733. I have no idea what the cause of this is, but I'll take a look.

null-a:
The tests pass locally on Node v7.0.0.

PR #769, opened 2017-01-26
Update docs to reflect change to parameter store.

null-a:
I guess we missed this when we did #708.


null-a:
The CI failure was from an unrelated inference test that happened to fall outside of its tolerance.

PR #775, opened 2017-02-15
Dream Learning

null-a:
This is @dorarad's dream learning implementation.

The changes I've made are:

* Re-work the back end interface for `mapData` in order to support ELBO and dream.
* Update for guide thunks.
* Revert to separate coroutines for sampling traces and estimating gradients. (While retaining the approach of interleaving these two phases, rather than generating a fixed bag of traces and optimizing on those.) This seems OK to me, but maybe there's a problem I'm not seeing?
* Add a `dreamBatchSize` option to `mapData` to allow control over the number of times the observation function is executed per sample. (Note that this defaults to one.)
  
As we know, the current implementation makes several assumptions about the structure of the model. I've tried to note these in comments, and where possible notice at run time when a program violates an assumption. One thing I'm not clear about is whether the current implementation does something sensible in the presence of additional factor statements. (i.e. In addition to those coming from `observe`.) So for now, we generate a run time error if we encounter such a thing, but I definitely need to understand this better.

Aside from that, this is intended to support the same class of models and optimize the same objective as the original implementation.

Usage examples:

```js
var model() {
  // global stuff
  mapData({data: someData, dreamBatchSize: 10}, function(x) {
    // local latents
    observe(someDist, x);
  };
};

Optimize({model, estimator: 'dream'});
// More samples per step:
Optimize({model, estimator: {dream: {samples: 10}}});
```


stuhlmueller:
Before I dig in more deeply, I want to make sure I understand the behavior of dream using a simple Bernoulli mixture.

### Global mixture component

First, consider a version where the mixture component is chosen globally:

~~~~{js}
var data = [false];

var model = function() {

  var bool = flip(.5);
  var p = bool ? .1 : .9;

  mapData({ data }, function(x) {
    var dist = Bernoulli({ p });
    observe(dist, x);
  });

  return bool;
};

Optimize({
  model,
  estimator: 'dream',
  steps: 10000
});

Infer({
  model,
  method: 'forward',
  samples: 10000,
  guide: true
});
~~~~

With dream, I observe:

~~~~
Marginal:
    false : 0.5059
    true : 0.4941
~~~~

By contrast, with ELBO:

~~~~
Marginal:
    true : 0.903
    false : 0.097
~~~~

My understanding: In this model, no parameters are used within `mapData`. Since those are the only parameters that dream updates, no parameter changes take place at all.

### Per-datum mixture component

Now suppose that the mixture distribution is chosen per data point:

~~~~{js}
var data = [false];

var model = function() {
  mapData({ data }, function(x) {
    var bool = flip(.5);
    var p = bool ? .1 : .9;
    var dist = Bernoulli({ p });
    observe(dist, x);
    return bool;
  });
};

Optimize({
  model,
  estimator: 'dream',
  steps: 10000
});

var dist = Infer({
  model,
  method: 'forward',
  samples: 10000,
  guide: true
});

marginalize(dist, function(x){ return _.first(x); });
~~~~

The return distribution I observe looks like this:

~~~~
Marginal:
    true : 0.5055
    false : 0.49450000000000005
~~~~

By contrast, if I use ELBO, I see:

~~~~
Marginal:
    true : 0.8974000000000001
    false : 0.1026
~~~~

My understanding: In dream, within `mapData`, both `observe` and guide parameters are ignored while sampling. We simply sample data from the prior (but values sampled outside of `mapData` may be sampled using learned parameters).

Based on this hallucinated dataset, we then update the guide parameters used within `mapData` to maximize the probability of the data under the guide distribution. In the program above, this means that we update the mean-field guide for `flip(.5)` based on samples from the prior. This results in a guide that is identical to the prior.

### Per-datum mixture component with data-dependent guide

Third, consider a version with data-dependent guide:

~~~~{js}
var data = [false];

var model = function() {
  mapData({ data }, function(x) {
    var bool = sample(Bernoulli({ p: 0.5 }), {
      guide() {
        var p = Math.sigmoid(param({ name: (x ? 'a' : 'b') }));
        return Bernoulli({ p });
      }
    });
    var p = bool ? .1 : .9;
    var dist = Bernoulli({ p });
    observe(dist, x);
    return bool;
  });
};

Optimize({
  model,
  estimator: 'dream',
  steps: 10000
});

var dist = Infer({
  model,
  method: 'forward',
  samples: 10000,
  guide: true
});

marginalize(dist, function(x){ return _.first(x); });
~~~~

With dream, I observe:

~~~~
Marginal:
    true : 0.8775
    false : 0.12250000000000001
~~~~

ELBO looks similar:

~~~~
Marginal:
    true : 0.9031
    false : 0.09690000000000001
~~~~

My understanding: In dream, we again take samples from the prior within `mapData`. Unlike in the previous case, we now have a data-dependent guide, so updating its parameters based on a dataset hallucinated from the prior results in a guide that isn't identical to the prior.

Does that sound right? Did I leave out any important considerations?

I think it might be worth adding tests for cases like the ones above (especially the last one).

dorarad:
Andreas: completely correct. Though one small thing I'm not sure about is why did you use different parameters based on x in the third case: param({ name: (x ? 'a' : 'b') })?
You can also look at the report for dream behavior on simple cases:
https://docs.google.com/document/d/1SGBkUVHGHFZQC5P50bAavC55lmSGFdwE3RNH92M-YZs/edit?usp=sharing
it would also be a good idea to use these at tests in order to ensure the pr or further changes don't break dream behavior (@null-a).

src/inference/dream/estimator.js
```diff
@@ -0,0 +1,76 @@
...
+// This estimator currently makes the following assumptions:
+
+// 1. The model contains exactly one `mapData`.
+
+// 2. Either a) each element of the data array is observed (passed to
+// `observe` as its second argument) exactly once in the corresponding
+// observation function, or b) each element of the data array is
+// itself an array, of which each element is observed exactly once,
+// and in the order in which they appear in the array. See below for
+// examples of each of these.
```
dorarad: Ultimately I think it would be a good idea to add support for giving a dictionary as an argument in order to get rid of the same-order and multiplicity assumptions.

src/inference/dream/estimator.js
```diff
@@ -0,0 +1,76 @@
...
+
+// 3. There are no factor statements. We assume we can generate
+// samples from the posterior predictive distribution directly by
+// forward sampling. If there were additional factors we'd need to
+// account for them with e.g. importance sampling.
```
dorarad: correct

dorarad:
@null-a looks fine overall! I'll look later today / tomorrow more closely and let you know if I find some bug or anything. The best test would be to see it works for the cases from the doc I've mentioned + a case with both ELBO/Dream for model learning (I tested it locally with orig version of dream). 
- Separating the stages to different coroutines shouldn't cause problems. 
- The note regarding the assumptions is correct. See my comment regarding generalization to dictionary arguments in observe. It won't damage if we used a counter for insideMapData rather than a flag but that's indeed not necessary at this stage since this alone won't support nesting. Dream won't work for importance-sampling so probably shouldn't support factor statements. 
- What is `dreamBatchSize` used for?

null-a:
Here's my two cents:

Pretty much everything you've said sounds right to me. There's just one thing, related to your second example, that could perhaps be clarified. See what you think:

> this means that we update the mean-field guide for flip(.5) based on samples from the prior. This results in a guide that is identical to the prior

The reason the guide for the latent `flip` looks like the prior is not because optimization is fitting it to the prior, but rather that optimization isn't moving the guide at all, and it just happens to be initialized to something close to the prior.

(In other words, changing the success probability of the latent `flip` to something other than 0.5 won't change the result of this program.)

The reason for this is that when using mean field, the latent `flip` associated with the observation, and the latent `flip` for a posterior predictive sample, have independent guide distributions. `dream` will optimize the latter but not the former.

If we write the guide for the latent `flip` by hand, and give the parameter a name to induce sharing across applications of the observation function, then the guide _will_ be optimized to match the prior.

(I intentionally introduced this behavior in this [change](https://github.com/probmods/webppl/pull/775/commits/88783816887d35a86e648382dc9584bfaed9ace7) to the original implementation.)

> I think it might be worth adding tests for cases like the ones above (especially the last one).

I've added your first and last examples as test cases. (The tolerances for these may need tweaking over time.)

null-a:
@dorarad: Great, thanks!

> What is dreamBatchSize used for?

It allows control over the number of times the observation function is executed per sample. (Instead of always executing it `data.length` times.)

dorarad:
@null-a so why does dream have specific flag rather than generic batchsize that can be used for elbo or other estimators?

dorarad:
@null-a I'm not sure regarding what you said: it should be checked what happens in current impl but in the orig one as can be seen in mean-field examples in the doc (page 12 for instance) when mean-field guide is used dream returns the prior correctly. The parameters are changed over training to fit that.

dorarad:
@null-a I'm not sure regarding what you said: it should be checked what happens in current impl but in the orig one as can be seen in mean-field examples in the doc (page 12 for instance) when mean-field guide is used dream returns the prior correctly. The parameters are changed over training to fit that. I think the reason for difference between the second and third cases is that when the observation is not supported as input to the guide than dream is unable to learn a function based on the observation, and so it averages the results over the hallucinated observations to fit the prior (i.e. trying to fit a constant to the target distribution rather than a function of the observation, yielding the prior).

null-a:
@dorarad:

> so why does dream have specific flag rather than generic batchsize that can be used for elbo or other estimators?

It wasn't clear to me that these two things would always coincide, so I figured it made sense to keep things flexible while we continue to explore how this works.

> Maybe the reason for the different behavior than the orig is the guide-thunks pr / was there a change recently in the way automatic mean-field is created?

Not quite:

> (I intentionally introduced this behavior in this [change](https://github.com/probmods/webppl/pull/775/commits/88783816887d35a86e648382dc9584bfaed9ace7) to the original implementation.)

stuhlmueller:
> The reason for this is that when using mean field, the latent `flip` associated with the observation, and the latent `flip` for a posterior predictive sample, have independent guide distributions. `dream` will optimize the latter but not the former.

Thanks! I missed that. So, my explanation (under **Per-datum mixture component** above) is only correct for a model that has explicit sharing, like this:

~~~~
var data = [false];

var model = function() {
  mapData({ data }, function(x) {
    var bool = sample(Bernoulli({ p: 0.5 }), {
      guide() {
        var p = Math.sigmoid(param({ name: 'p' }));
        return Bernoulli({ p });
      }
    });    
    var p = bool ? .1 : .9;
    var dist = Bernoulli({ p });
    observe(dist, x);
    return bool;
  });
};

Optimize({
  model,
  estimator: 'dream',
  steps: 10000
});

var dist = Infer({
  model,
  method: 'forward',
  samples: 10000,
  guide: true
});

marginalize(dist, function(x){ return _.first(x); });
~~~~

> Though one small thing I'm not sure about is why did you use different parameters based on x in the third case: param({ name: (x ? 'a' : 'b') })?

If you don't do this, what you get is the code above (in this reply)—your guide isn't data-dependent and fitting it with dream won't move it away from the prior.

src/inference/dream/gradients.ad.js
```diff
@@ -0,0 +1,116 @@
...
+      return this.estimateGradient(function(grad, objVal) {
+        env.coroutine = this.coroutine;
+        return this.cont(grad, objVal);
+      }.bind(this));
+    },
+
+    estimateGradient: function(cont) {
+
+      this.paramsSeen = {};
+      this.logp = this.logq = 0;
```
stuhlmueller: Why do we keep track of `logp`?

src/inference/elbo.ad.js
```diff
@@ -303,7 +303,12 @@ module.exports = function(env) {
       return k(s);
     },
 
-    mapDataFetch: function(data, batchSize, address) {
+    mapDataFetch: function(data, opts, address) {
+
+      var batchSize = opts.batchSize !== undefined ? opts.batchSize : data.length;
+      if (batchSize < 0 || batchSize > data.length) {
```
stuhlmueller: Do we want to support `batchSize` 0?

src/inference/dream/estimator.js
null-a: I don't have a clear idea about how best to implement this yet, so I'd be in favour of tackling it later. Of course, someone else may already have a good approach, but even so, doing it as a separate PR seems like a good way to go to me. (Unless it's super easy.)

src/inference/dream/sample.js
```diff
@@ -0,0 +1,152 @@
...
+      if (_.isEmpty(data)) {
+        // In general, we need data to yield to the observation
+        // function in order to generate fantasy data.
+        throw new Error('dream: data should be non empty.');
+      }
+
+      // Flag indicating whether each element of the original data is
+      // an array of observations or a single observation. (We check
+      // the first datum, and assume the rest of data would return the
+      // same.)
```
null-a: You're right. I've [re-worked](https://github.com/null-a/webppl/blob/a57f6fd210ad189e614064cdcc478d672f42a4bd/src/inference/dream/estimator.js#L10) this to unwrap single observations based on the number of times `observe` is called (in a particular application of the observation function) rather than by peeking at the data. I think this is an improvement. I've also added some [tests](https://github.com/probmods/webppl/pull/775/commits/24d447a109f6e5583d21bac89ed45b00acee7cb1) for this stuff.

src/inference/dream/sample.js
```diff
@@ -0,0 +1,152 @@
...
+      this.obsArr = _.isArray(data[0]);
```
null-a: This variable doesn't exist any more, so we're good. (But I think you were right.)

src/inference/dream/estimator.js
dorarad: don't think it has to be in the same pr necessarily but in principle: I thought that if we get observations as an object - a list of keys and values, and also extend the observe function to get an observations object and a key rather than directly the value, then we can in the first pass start every iteration with copying the object, then in every observe statement just hallucinate the value and overwrite it in the copied object and ultimately just pass this object to the second pass and it should work fine from there without additional modifications.

src/inference/dream/gradients.ad.js
null-a: Only because I expected we'd need it soon, but I've removed it for now.

src/inference/elbo.ad.js
null-a: I'm not sure. One argument for not doing so might be something like by setting it to zero we're not only specifying something about how the inference strategy, we changing the objective (by effectively removing the `mapData`). On the other hand, if there's utility in having such a thing I'd be happy to add it. Separate PR?

stuhlmueller: It looked to me like we're currently supporting `batchSize` 0 (due to checking `batchSize < 0` instead of `batchSize <= 0`), but I didn't see the utility in it / thought that it would more likely result from a mistake than from an intentional choice. I don't think we should add it if it's not supported right now.

null-a: Oops, sorry, I should have read the code and not just the question!

It looks like we didn't support this when it was first written, but then I [changed](https://github.com/probmods/webppl/pull/558/commits/0756b3d14f31941bbaf0c94926b7ab015644b513) it for some reason. Unfortunately, the commmit doesn't say why. I'm with you though, that this is most likely to be a mistake and we're likely better off without it. (Unless there's something we're missing.) I'll open an issue about changing this.

PR #781, opened 2017-02-21
Default Infer supported when no options provided

hereneric:
Introduction here: https://goo.gl/Me4ylk

Tested by wppl scripts on probmods v2. One thing I'm concerned about is the shortcut exit in enumerate.ad.js, not sure if only doing "return this.k();" is good or not.


null-a:
This is neat. I tried it on a few simple models and it worked as I expected based on your description of the heuristics. Good job!

There was one thing I didn't understand: Why is the complexity threshold for enumerate based on the total number of nodes (`num_evals` in code, I think) rather than the number of leaves? For example, in a program that just uses `flip` these differ by a constant factor (give or take), so at first glance it seems you could instead base the decision on the number of leaves, and just halve the tolerance. I guess I'm missing something though?

You've probably already thought about this, but I wonder if we should run the sampling methods (when selected) for a fixed length of time, rather than generate a fixed number of samples? Or alternatively limit how long they run for? (Not necessarily as part of this PR.)

**Thoughts on the implementation**

Reading the code, the main thought I had is that a lot of what we're adding seems quite specific to "default infer", and i wonder if some of this can be factored into more independent parts that "default infer" then builds on.

Here are some (not fully thought out) examples that came to mind, they might give you a better idea of what I'm thinking.

* Inference coroutines could take a separate option `throwOnError` that indicates whether to throw a JS exception or return some special value when an error is encountered. The implementation of this could be wrapped up in an `error` method on each coroutine. This might make it easier to add new errors in the future, and could plausibly have other uses.

* Related to this, I wonder if it might be clearer to return something other than integers when signalling an error. One option is to return dedicated values for the errors we need to distinguish. e.g. In the coroutine we might do something like `return this.error(errors.emptySupport)` and then in `DefaultInfer` we can have checks like `retVal === errors.emptySupport` etc. But even just returning a string (e.g. `'emptySupport'`) might help?

* `Rejection` could take a `minSampleRate` argument to indicate that an error should be generated when the sampling rate is too low. (What "error" means would depend on the `throwOnError` argument.) Something similar could be done for the check that `Enumerate` makes.

* I wonder if there's a way to avoid having the check for interleaved `sample` and `factor` smushed up with the rejection sampler? I can see that it's convenient right now, but if they were separate we'd avoid complicating `Rejection`, and the extra flexibility might be useful if we choose to change the decision tree changes in the future. One option is to have a separate coroutine for this, but perhaps there are other alternatives.

**Further thoughts**

* I don't think the `shuffle` method in `Enumerate` should mutate its argument. It's hard to know where else that value is used, and modifying it in-place could cause confusion or bugs. For example, this shuffling can already be observed from program code:

```js
var dist = Categorical({ps: [1,1,1,1], vs: ['a','b','c','d']});
display(dist.support());
Infer({model() { return sample(dist); }});
display(dist.support()); // <--- will be shuffled
```

* Related to this, `shuffle` should probably use `util.random` rather than `Math.random` internally. (This uses the RNG we can seed from e.g. the command line.)

* I don't think we need to worry about recovering from the assertions in `SMC`, since we'll only run into these if there's a bug in webppl. This [comment](https://github.com/probmods/webppl/issues/317#issuecomment-182476496) explains how I (try to) use `assert` vs. `throw`, and might make this clearer.

* However, [this assertion](https://github.com/probmods/webppl/pull/781/files#diff-d36c5d9c1355d9f935bcc029077ca33cR97) would be better as a `throw`. (I didn't follow my own advice here.) So if we have something like the `this.error()` idea mentioned above, we could use it here too.

* `dists.isDist()` is callable from webppl and might come in handy.

ngoodman:
these are great comments. i think @Eric-H-Ren is game to clean up and refactor, to get to the point where we can merge.

> Reading the code, the main thought I had is that a lot of what we're adding seems quite specific to "default infer", and i wonder if some of this can be factored into more independent parts that "default infer" then builds on.

relatedly, if we can expose the features that go into the decision tree in a unified way, it will help in the future if we try to use some learning to improve the heuristics. it's not totally clear to me how to think about these features, since they are often runtime features -- something to keep in mind though.

Eric-H-Ren:
Hi Paul, thank you for the great comments/thoughts. I am going to improve the code based on them. I want to go through every point and clarify some of them to make sure I get things right:

a) 

> Why not estimate the complexity by the number of leaves?

 The turnaround time for enumerate is proportional to the number of nodes, which is a constant factor times the number of leaves only if the numbers of supports are the same at each level (geometric sequence). Therefore halve the tolerance if the last sampler has 2 supports like flip. Can't alway halve though, generally should divide the tolerance by the number of supports of the last sampler. However, the number of supports usually differs at each level, say [2,2,2,100], rather than geometric. In this case there are 800 leaves but only 815 nodes in total. We can’t take the shortcut by doing 1/2 or 1/100.

b) 

> Why not run the sampling methods (when selected) for a fixed length of time, rather than generate a fixed number of samples?

The main reason I thought is because for many simple scripts the sampling methods probably don't need that much time to generate enough samples. If we set the fixed length of time to say 5 seconds, then for some scripts 20000+ of samples got through, where 1000 samples are enough to give a good visualization I guess? They could have exited in 1 seconds if you just need 1000. So I use fixed number of samples to guarantee the rejection gives a fairly good visualization as soon as it could. What do you think is the advantage of fixed time rather than fixed number of samples?

c) 

> Inference coroutines could take a separate option throwOnError that indicates whether to throw a JS exception or return some special value when an error is encountered.

Yes, I agree. In the current PR, I use `options.probe` to indicate whether to throw a JS exception or return some special value. But options.probe is also responsible for some other things to happen, not just error stuff, so I guess it is better to assign the error option to another independent agent called throwOnError to have more flexibility over error itself.

d) 

> I wonder if it might be clearer to return something other than integers when signaling an error. 

Good idea. Will get this done.

e) 

> Rejection could take a minSampleRate argument to indicate that an error should be generated when the sampling rate is too low.

 The rejection sampling is taking `options.probe` as its min sample rate. Do you mean we expose `minSampleRate` as an option to users?

f) 

> I wonder if there's a way to avoid having the check for interleaved sample and factor smushed up with the rejection sampler?

 OK, I can have a separate coroutine for it for flexibility, but it would be slower than we let rejection keep track of it along the way. Another way to check is by parsing the code before runtime, but that's not as convenient.

g) 

> I don't think the shuffle method in Enumerate should mutate its argument. 

True. Actually I think the shuffle can be removed. The reason I used it is because I want to select a completely random DFS path to estimate total nodes of enumeration tree. But I guess the randomness here is not necessary, and introduce other side effects.

f) 

> I don't think we need to worry about recovering from the assertions in SMC.

 It seems that when a model has very strict condition, SMC would be very likely to stop here:
`assert.notStrictEqual(logAvgW, -Infinity, 'All particles have zero weight.');`
Except for this, other asserts are fine without `try/catch`. I will create an `error()` function for each method, as you said, `return this.error(errors.emptySupport)`.

null-a:
> a
> which is a constant factor times the number of leaves only if the numbers of supports are the same at each level

Doh, of course! Thanks.

> b
> What do you think is the advantage of fixed time rather than fixed number of samples?

1. It seems that for a fixed number of samples, a big enough model will take a long time to sample from. A user who doesn't know how default infer works (internally) is likely to be uncertain about how long inference will run for. So it seems possible that the way people will end up using this is by trying it for a few seconds, and if no results are produced hit cancel and try specifying inference by hand. If this happens, having an upper bound on the run time would be useful.

2. I think I sometimes take lots of samples from tiny models, and only a few samples from big models. So I wonder if there are occasions where I'm approaching inference with a fixed time/computation budget in mind, and that when I'm specifying inference options by hand what I'm really doing is tweaking them to come in on budget. If so, having an automatic way to do this would be useful.

To be clear, I'm not at all sure which approach is best. And I'm totally fine with leaving things as they are and seeing what happens once we start using this.

> d

> Good idea. Will get this done, like this?
> var error = function(type) {
> // var type is a string
> if (throwOnError) {
> throw new Error(type)
> } else {
> return this.exit(type)
> }
> }

Yeah, that's the kind of thing I had in mind. (Though I guess the function will be best as a property of the coroutine so that `this` is bound to the coroutine.)

> still need to call this.exit() to return to defaultInfer right? can we directly do return this.k(this.store, type);?

I think you'll be OK to call `this.k` without going through `exit`.

(I'm happy to elaborate on this if that will be helpful.)

> e
> The rejection sampling is taking options.probe as its min sample rate. Do you mean we expose minSampleRate as an option to users?

Yes, in the sense that it would available from webppl code. (As `probe` already is, right?) But I don't think we need to add it to the docs. Does that help?

> f
> OK, I can have a separate coroutine for it for flexibility, but it would be slower than we let rejection keep track of it along the way.

If we only execute the program one or a few times(?), then I don't think this is worth worrying about?

> h
> it seems that when a model has very strict condition, SMC would be very likely to stop here:

You're right, good catch!

Eric-H-Ren:
Please see the updated code with error methods and a separate coroutine for interleaving sample and factor check etc.

null-a:
> Please see the updated code with error methods and a separate coroutine for interleaving sample and factor check etc.

The changes look great, I think they help -- what do you think?

I'll make another pass through this and add any comments inline. If this prompts work that will take more time than you have to put into the project, please let me know and we'll figure something else out.

.gitignore
```diff
@@ -1,5 +1,7 @@
 .DS_Store
-.pyc
+*~
+*.svg
+*.pyc
```
null-a: Maybe revert the changes to `.gitignore` since they aren't directly related to this PR?

src/header.wppl
```diff
@@ -469,11 +469,48 @@ var OptimizeThenSample = function(wpplFn, options) {
...
+ * the "probe" field in the option is a signal to enumerate/rejection/SMC
+ * in probe mode these methods don't throw error, but return useful information for deciding
+ * next step in the heuristics flowchart
+ */
+var DefaultInfer = function(wpplFn, options) {
+  var enumTol = 200000;
+  var minSampleRate = 250; //from 1 to sampleNum
+  var sampleNum = 1000;
+  var enumResult = Enumerate(wpplFn, {probe: enumTol, strategy: 'depthFirst'});
+  if (_.isObjectLike(enumResult)) {
```
null-a: Could we use `dists.isDist()` to check the return value here? (It might make this a little easier to understand.)

src/inference/enumerate.ad.js
```diff
@@ -19,6 +19,21 @@ module.exports = function(env) {
     options = util.mergeDefaults(options, {
       maxExecutions: Infinity
     });
+
+    this.throwOnError = options.throwOnError !== undefined ? options.throwOnError : true;
```
null-a: I think you will be able to use the existing call to `mergeDefaults` (above) to more compactly specify the default value for `throwOnError`.

src/inference/enumerate.ad.js
```diff
@@ -47,6 +62,17 @@ module.exports = function(env) {
...
   };
 
+  // Error function for error handling
+  // this.throwOnError is true: directly throw error
+  // this.throwOnError is false: return error (string) as infer result
+  Enumerate.prototype.error = function(errType) {
+    if (this.throwOnError) {
+      throw new Error(errType);
+    } else {
+      return this.k(this.store, errType + '..quit enumerate');
```
null-a: I think that these `error` methods should be general purpose. The `".. quit enumerate"` string seems like something specific to "default infer", so I wonder if it shouldn't be somewhere else? Perhaps as part of `DefaultInfer` in `header.wppl`?

src/inference/enumerate.ad.js
```diff
@@ -67,21 +93,33 @@ module.exports = function(env) {
...
-      throw new Error('Enumerate encountered a distribution with empty support!');
+      return 'Enumerate encountered a distribution with empty support.';
     }
-
     return supp;
   };
 
   Enumerate.prototype.sample = function(store, k, a, dist) {
     var support = getSupport(dist);
+    if (_.isString(support)) {
```
null-a: I wonder if we should just call `error` directly from within `getSupport`, rather than having this check on the return type?

src/inference/rejection.js
```diff
@@ -22,6 +21,17 @@ module.exports = function(env) {
       maxScore: 0,
       incremental: false
     });
+
+    this.throwOnError = options.throwOnError !== undefined ? options.throwOnError : true;
+    // the value of options.probe is the min sample rate
+    this.probe = options.probe;
```
null-a: As you make progress on making the things that "default infer" is implemented on top of more independent, I think it will make sense to replace these generic `probe` options with something more specific. e.g. Here, rejection could have a `minSampleRate` option.

src/inference/rejection.js
```diff
@@ -22,6 +21,17 @@ module.exports = function(env) {
...
+    if (this.probe) {
+      // not throw error in probe mode
+      this.throwOnError = false;
```
null-a: I think `DefaultInfer` should pass in `{throwOnError: false}` rather than assuming that when we 'probe' we'll always want this particular error behavior.

src/inference/rejection.js
```diff
@@ -22,6 +21,17 @@ module.exports = function(env) {
...
+    }
+    this.numSamplesBak = options.samples;
```
null-a: I was a little confused by the way we track the number of samples that have been taken so far. This is largely down to the existing code in `Rejection`, but I think calling `numSamplesBak` something like `numSamplesTotal` might have helped a little. What do you think?

src/inference/smc.js
```diff
@@ -134,7 +155,7 @@ module.exports = function(env) {
   };
 
   function resampleParticles(particles) {
-
+    // return env.coroutine.error('nothing');
```
null-a: I think it's better to remove this rather than comment it out.

src/inference/interleavingSF.js
```diff
@@ -0,0 +1,79 @@
...
+  InterleavingSF.prototype = {
+
+    run: function() {
+
+      return util.cpsLoop(
+          this.opts.samples,
+
+          // Loop body.
+          function(i, next) {
+            this.logWeight = 0;
```
null-a: I don't think this is used. Remove if not?

.gitignore
Eric-H-Ren: I think the `.pyc` was supposed to be `*.pyc`? and the other two lines are also used for similar purpose - ignore irrelevant files that may appear during local development. Many `*~` and `*.svg` (viz output) appears in the directory when I was working on this. I guess these changes are useful? or not?

src/header.wppl
Eric-H-Ren: Agree!

src/inference/enumerate.ad.js
Eric-H-Ren: Yes. I think so.

src/inference/enumerate.ad.js
Eric-H-Ren: Yes. Will move it to DefaultInfer.

src/inference/rejection.js
Eric-H-Ren: I think `options.probe: true` means both `minSampleRate: 250` and `throwOnError: false`. So I can replace `probe` with these two settings.

src/inference/rejection.js
Eric-H-Ren: Yes. can do `{minSampleRate: ..., throwOnError: false}`.

src/inference/rejection.js
Eric-H-Ren: OK. Such renaming is good. I guess this part is confusing because it counts down rather than counts up?

src/inference/interleavingSF.js
Eric-H-Ren: Yes, you are right.

src/inference/enumerate.ad.js
Eric-H-Ren: I tried to do so at first. But it seems the error function called in `getSupport()` like `return env.coroutine.error('...');` doesn't interrupt the execution of enumeration? The enumeration keeps running rather than return error msg to header.wppl. I'm a little confused.

Eric-H-Ren:
Hi Paul, thanks for the reviews. No worries about my time. I'll fix the problems and update the PR.

Eric-H-Ren:
I have fixed all problems except for the .gitignore and calling error in `getSupport`. I have some questions about these two as I replied above.

.gitignore
null-a: I can imagine making a case for adding `*.svg`, but I'm not sure what`*~` files are? If this isn't something many people will run into, adding a local exclusion to `.git/info/exclude` might be a better approach?

(BTW, the only reason I suggested reverting these was that in general, having separate PRs for independent changes seems to work well. But now that we've started, I'm happy to finalize this here.)

Eric-H-Ren: OK, I'd remove if not many people need. *~ files are temporary files generated by vim.

src/inference/enumerate.ad.js
null-a: One way to make this work is to adjust `getSupport` to take a continuation. This is a drag, but it seems better than expecting every caller of `getSupport` to check the return value and call `this.error` as appropriate? Having the signature be `getSupport(dist, cont)` would be consistent with other similar code I think. (`cont` rather than `k` because this would be a JS continuation rather than a webppl one. e.g It doesn't take a `store` argument.)

null-a: Also, I think the call to `getSupport` in `sampleWithFactor` will need updating to work with error handling.

src/inference/smc.js
```diff
@@ -71,6 +79,20 @@ module.exports = function(env) {
...
+  // this.throwOnError is true: directly throw error
+  // this.throwOnError is false: return error (string) as infer result
+  SMC.prototype.error = function(errType) {
+    if (this.throwOnError) {
+      throw new Error(errType);
+    } else {
+      this.err = errType;
+      // the way to return differs from enumerate and rejection because of CPS
+      // directly calling this.k doesn't work
+      return this.finish();
```
null-a: I think it would be nice if these `error` methods worked in the same way. What doesn't work when calling `this.k` directly here? (I've tried to figure it out be can't see it.)

src/inference/enumerate.ad.js
Eric-H-Ren: OK, I see... What are we going to pass to `getSupport(dist, cont)` as `cont`?

src/inference/smc.js
Eric-H-Ren: I guess this is a similar problem as `getSupport`. Still trying to figure out the continuation thing.
Calling `this.k()` doesn't interrupt the execution of smc, neither does `this.finish()`. But it seems using `this.finish()` the error function works fine on the test set.

src/inference/rejection.js
null-a: I was imagining the other coroutines would be similarly updated. Does that sound reasonable?

null-a:
I remembered that the docs [mention](http://docs.webppl.org/en/master/inference/index.html) that enumeration is the default inference method. Are you OK to update that sentence to reflect the changes here? (See `docs/inference/index.rst`.)

src/inference/rejection.js
Eric-H-Ren: For enumeration, the probe means three things: a) estimate the complexity and quit if it's above threshold, b) quit when time out, c) `throwOnError: false`

So what are we going to do with a) and b), using two additional options like `maxComplexity: 200000` and `timeOut: 5000`?

src/inference/smc.js
null-a: I'd like to recreate the problem, but when I tried changing `this.finish()` to `this.k(this.s, errType)` then all the tests passed, and writing test programs to generate errors behaved as I expected. Any idea what I'm missing?

(I think all the places we call `this.error` in SMC are in CPS already, so I don't think it's the same problem as `getSupport`, but I could be mistaken.)

src/inference/rejection.js
null-a: Sounds good to me.

src/inference/enumerate.ad.js
null-a: We need to pass a function that represents the computation that happens after `getSupport` -- this is the continuation. So the body of the `sample` method will go from something like this:
```js
var supp = getSupport(dist);
// rest of body
```
to this:
```js
return getSupport(dist, function(support) {
  // rest of body
});
```

.gitignore
null-a: Gotcha, thanks. If it's something that helps vim users, then I see no harm in keeping it.

src/inference/smc.js
Eric-H-Ren: I tested another time and `this.k(this.s, errType)` is working fine. I guess I might have be mistaken.

src/inference/enumerate.ad.js
Eric-H-Ren: Awesome. It is working now.

Eric-H-Ren:
The code has been updated. Is this the version we can safely merge? I added a simple description for default infer in `docs/inference/index.rst`. I can do more if needed.

src/inference/smc.js
```diff
@@ -25,14 +25,17 @@ module.exports = function(env) {
...
+      throwOnError: true
     });
 
     if (!_.includes(validImportanceOptVals, options.importance)) {
       var msg = options.importance + ' is not a valid importance option. ' +
           'Valid options are: ' + validImportanceOptVals;
       throw new Error(msg);
     }
+    this.throwOnError = options.throwOnError;
+    this.err = undefined;
```
null-a: Am I right in thinking we don't need this property now we call `this.k` directly from `this.error`?

docs/inference/index.rst
```diff
@@ -15,14 +15,18 @@ function of zero arguments representing a stochastic computation and
...
-   }});
-
-``Infer`` will perform inference using :ref:`enumeration<enumerate>`
-by default, but several other implementations of marginal inference
-are also built into WebPPL. Information about the individual methods
-is available here:
+   });
+
+This example has no inference options specified. By default, ``Infer``
+will perform inference using one of the method among enumeration,
```
null-a: I like this. One small thing, should it instead be "one of the methods"?

src/inference/smc.js
```diff
@@ -145,7 +159,10 @@ module.exports = function(env) {
     var logW = util.logsumexp(_.map(particles, 'logWeight'));
     var logAvgW = logW - Math.log(m);
 
-    assert.notStrictEqual(logAvgW, -Infinity, 'All particles have zero weight.');
+    if (logAvgW === -Infinity) {
+      // do not return, execution continues
+      env.coroutine.error('All particles have zero weight.');
+    }
```
null-a: I agree that this assertion should be an error, but I believe the remaining assertions in SMC are better left unchanged. (See my original comment for an attempt to explain my thinking.) Are you OK to revert these other ones?

null-a:
> Is this the version we can safely merge?

I think we're really close. I've added a few additional comments inline.

I noticed that a few models in `/examples` assume that enumerate is the default inference method, by specifying a `maxExecutions` but omitting a `method`. I think we should probably update these models so that they explicitly request enumeration, otherwise these `maxExecutions` options aren't doing anything. (And for a few of these default infer is giving up on enumeration, as without `maxExecutions` the timeout threshold is exceeded.)

```
geometric.wppl:5:Infer({maxExecutions: 10, model() {
hmmIncremental.wppl:25:Infer({maxExecutions: 100, model() {
hmmSampleWithFactor.wppl:28:Infer({maxExecutions: 500, model() {
pcfg.wppl:38:Infer({maxExecutions: 300, model() {
pcfgIncremental.wppl:43:Infer({maxExecutions: 300, model() {
pragmaticsWithSemanticParsing.wppl:151:  Infer({maxExecutions: 100, model() {
pragmaticsWithSemanticParsing.wppl:160:  Infer({maxExecutions: 100, model() {
pragmaticsWithSemanticParsing.wppl:169:  Infer({maxExecutions: 100, model() {
```

One last thought: I suspect that at some point it will be useful to have the option of having default infer work silently, without writing to the console. (A concrete example would be if we want to include default infer in inference tests without cluttering up the output of the test runner.) I don't think we need this right now, so we can delay adding it until we do, but I figured the thought was worth sharing.

src/inference/smc.js
Eric-H-Ren: OK, you are right.

src/inference/smc.js
```diff
@@ -71,6 +73,18 @@ module.exports = function(env) {
...
   };
 
+  // Error function for error handling
+  // this.throwOnError is true: directly throw error
+  // this.throwOnError is false: return error (string) as infer result
+  SMC.prototype.error = function(errType) {
+    if (this.throwOnError) {
+      throw new Error(errType);
+    } else {
+      this.err = errType;
```
stuhlmueller: What's the reason for assigning `this.err` here?

src/header.js
```diff
@@ -34,6 +34,7 @@ try {
   var incrementalmh = require('./inference/incrementalmh');
   var optimize = require('./inference/optimize');
   var forwardSample = require('./inference/forwardSample');
+  var interleavingSF = require('./inference/interleavingSF');
```
stuhlmueller: I'd rename `interleavingSF` to `checkSampleAfterFactor`, or something similar that is more explicit about what the coroutine does.

src/header.wppl
```diff
@@ -469,11 +469,49 @@ var OptimizeThenSample = function(wpplFn, options) {
...
 };
 
-var Infer = function(options, maybeFn) {
-  if (!util.isObject(options)) {
-    util.fatal('Infer: expected first argument to be an options object.');
+/*
+ * DefaultInfer() called when no options are specified
+ * var enumTol: upper bound for enumeration tree size, enumeration ends when above threshold.
+ * var minSampleRate: lower bound for #sample/sec, rejection ends when below threshold.
+ * var sampleNum: number of samples/particles used for default rejection/SMC/MCMC.
```
stuhlmueller: Is the comment in the next two lines stale, or am I just not finding what it refers to?

src/header.wppl
```diff
@@ -469,11 +469,49 @@ var OptimizeThenSample = function(wpplFn, options) {
   return SampleGuide(wpplFn, opts);
 };
 
-var Infer = function(options, maybeFn) {
-  if (!util.isObject(options)) {
-    util.fatal('Infer: expected first argument to be an options object.');
+/*
+ * DefaultInfer() called when no options are specified
+ * var enumTol: upper bound for enumeration tree size, enumeration ends when above threshold.
+ * var minSampleRate: lower bound for #sample/sec, rejection ends when below threshold.
```
stuhlmueller: Use `samples` instead of `sampleNum` below, to make it analogous to the name we use for the corresponding MCMC/Rejection option?

Eric-H-Ren:
The code is updated again. As for working silently, I think the console.log things for default infer are mostly in header.wppl. The silence feature can be easily implemented when needed.

src/header.wppl
```diff
@@ -469,11 +469,49 @@ var OptimizeThenSample = function(wpplFn, options) {
   return SampleGuide(wpplFn, opts);
 };
 
-var Infer = function(options, maybeFn) {
-  if (!util.isObject(options)) {
-    util.fatal('Infer: expected first argument to be an options object.');
+/*
+ * DefaultInfer() called when no options are specified
```
stuhlmueller: Use `maxEnumTreeSize` or something similarly explicit instead of `enumTol`? Also, use `estimatedTreeSize`/`maxTreeSize` in the enumeration file instead of "complexity"?

src/inference/enumerate.ad.js
```diff
@@ -47,6 +57,17 @@ module.exports = function(env) {
     return this.wpplFn(_.clone(this.store), env.exit, this.a);
   };
 
+  // Error function for error handling
+  // this.throwOnError is true: directly throw error
+  // this.throwOnError is false: return error (string) as infer result
+  Enumerate.prototype.error = function(errType) {
```
stuhlmueller: If you don't want to return strings when inference functions fail, you could also do something like this:

```js
var err = new Error(errType);
if (this.throwOnError) {
  throw err;
} else {
  return this.k(this.store, err);
}
```

Then you can check `returnVal instanceof Error` to determine whether the inference function failed.

src/inference/enumerate.ad.js
```diff
@@ -17,8 +17,18 @@ module.exports = function(env) {
...
+      maxExecutions: Infinity,
+      throwOnError: true,
+      timeOut: Infinity,
+      maxComplexity: Infinity
     });
+
+    this.throwOnError = options.throwOnError;
+    this.timeOut = options.timeOut; // Time upper threshold for enumeration
+    this.startTime = Date.now();
+    this.first_path = true; // whether enumeration has reached the first leaf/exit
```
stuhlmueller: Use `firstPath` and `levelSizes` to make the naming scheme consistent?

src/inference/enumerate.ad.js
```diff
@@ -17,8 +17,18 @@ module.exports = function(env) {
...
     options = util.mergeDefaults(options, {
-      maxExecutions: Infinity
+      maxExecutions: Infinity,
+      throwOnError: true,
+      timeOut: Infinity,
+      maxComplexity: Infinity
     });
+
+    this.throwOnError = options.throwOnError;
+    this.timeOut = options.timeOut; // Time upper threshold for enumeration
```
stuhlmueller: Use `maxRuntime` or `maxRuntimeInMilliseconds` instead of `timeOut`?

src/inference/enumerate.ad.js
```diff
@@ -103,27 +133,48 @@ module.exports = function(env) {
...
+            // Call the next state on the queue
+            return env.coroutine.nextInQueue();
+          },
+          support);
+    });
   };
 
+  var getComplexity = function(sizes) {
+    // Estimate enumeration tree size by support length at each level
+    var num_nodes = 1;
```
stuhlmueller: Use `numNodes` and `numEvals`?

src/inference/interleavingSF.js
```diff
@@ -0,0 +1,78 @@
...
+    this.opts = {samples: 1};
+    this.wpplFn = wpplFn;
+    this.s = s;
+    this.k = k;
+    this.a = a;
+
+    // has seen at least one factor operation
+    this.hasFactor = false;
+    // interleaving samples and factor
+    this.interleavingSampleFactor = false;
```
stuhlmueller: `this.hasSampleAfterFactor` would be more explicit.

src/inference/rejection.js
```diff
@@ -33,28 +41,49 @@ module.exports = function(env) {
...
       util.warn('Rejection: Reduce maxScore to zero for better performance.');
     }
   }
 
   Rejection.prototype.run = function() {
+    if (!_.isNumber(this.numSamples) || this.numSamples <= 0) {
+      return this.error('"samples" should be a positive integer.');
+    }
+    var elapseSec = (Date.now() - this.startTime) / 1000.0;
+    if (elapseSec > 2) {
```
stuhlmueller: `startTime` isn't reset on each iteration, so it looks like `elapseSec > 2` will always be true after it has been true once and `elapseSec` will increase monotonically. Did you intend to divide by `elapseSec` instead of `2` below?

src/inference/smc.js
Eric-H-Ren: Oh, this line should have been deleted.

src/header.js
Eric-H-Ren: OK.

src/header.wppl
Eric-H-Ren: OK.

src/inference/enumerate.ad.js
Eric-H-Ren: The `err` is going to be printed to console in header.wppl. I guess we don't want see a few info messages starting with "Error" when doing default infer?

src/inference/rejection.js
Eric-H-Ren: I think the logic here is we only care about what happens in the first 2 seconds. If `sampleRate` is good then we'll let rejection execute till it finishes, rather than check every time. Is it better to set a flag here so that `if (elapseSec > 2) {...}` will be skipped later on?

src/inference/enumerate.ad.js
stuhlmueller: If `err` is an `Error` instance, you can still access the message as `err.message`, so this should be a small change to `header.wppl`, right?

src/header.wppl
Eric-H-Ren: Yes they are stale. Should have deleted.

src/inference/rejection.js
stuhlmueller: I see. I'm not sure what's best here. One idea that doesn't require an extra flag would be to replace the conditional with `(this.minSampleRate > 0) && (elapseSec > 2)` and to set `this.minSampleRate` to 0 once the `sampleRate` check succeeds. This would make it explicit that you're only checking once after about two seconds and then accept any sample rate.

Eric-H-Ren: This idea is good.

src/inference/smc.js
null-a: Also, I just noticed that because we're not doing `return env.coroutine.error()` here, this won't do the right thing when we're not throwing errors. Inference will (attempt to) continue, rather than returning early with an `Error` instance. For example, this generates an error as expected:

```js
display(Infer({
  method: 'SMC',
  throwOnError: true,
  particles: 2,
  model: function() {
    condition(false);
  }
}));
```

but with `{throwOnError: false}` it returns a (incorrect) marginal, not an `Error`.

This is similar to the situation we found ourselves in with `getSupport`, and similar solutions will apply. I guess that switching to CPS will again be the least worst option.

src/header.wppl
```diff
@@ -469,11 +469,47 @@ var OptimizeThenSample = function(wpplFn, options) {
...
+ */
+var DefaultInfer = function(wpplFn, options) {
+  var maxEnumTreeSize = 200000;
+  var minSampleRate = 250; //from 1 to samples
+  var samples = 1000;
+  var enumResult = Enumerate(wpplFn, {maxEnumTreeSize: maxEnumTreeSize, maxRuntimeInMilliseconds: 5000,
+    throwOnError: false, strategy: 'depthFirst'});
+  if (dists.isDist(enumResult)) {
+    return enumResult;
+  } else {
```
stuhlmueller: Here and below, I'd check that the inference result is an error object (`enumResult instanceof Error`) and throw an error if it's neither a distribution nor an error object.

src/header.wppl
```diff
@@ -469,11 +469,47 @@ var OptimizeThenSample = function(wpplFn, options) {
...
+ * var maxEnumTreeSize: upper bound for enumeration tree size, enumeration ends when above threshold.
+ * var minSampleRate: lower bound for #sample/sec, rejection ends when below threshold.
+ * var samples: number of samples/particles used for default rejection/SMC/MCMC.
+ * next step in the heuristics flowchart
+ */
+var DefaultInfer = function(wpplFn, options) {
+  var maxEnumTreeSize = 200000;
+  var minSampleRate = 250; //from 1 to samples
+  var samples = 1000;
+  var enumResult = Enumerate(wpplFn, {maxEnumTreeSize: maxEnumTreeSize, maxRuntimeInMilliseconds: 5000,
```
stuhlmueller: You use `maxRuntimeInMilliseconds` here, but `maxRuntimeInMS` in `enumerate.ad.js`, so the option given here will be ignored.

src/inference/rejection.js
```diff
@@ -33,28 +40,51 @@ module.exports = function(env) {
...
 
   Rejection.prototype.run = function() {
+    if (!_.isNumber(this.numSamples) || this.numSamples <= 0) {
+      return this.error('"samples" should be a positive integer.');
+    }
+    var elapseSec = (Date.now() - this.startTime) / 1000.0;
+    if (this.minSampleRate > 0 && elapseSec > 2) {
+      // count the number of samples collected in ~2 secs
+      // compute number of samples per sec
+      var sampleRate = (this.numSamplesTotal - this.numSamples) / 2;
```
stuhlmueller: Even if you only run this check once, it's probably still a good idea to divide by `elapseSec` here (instead of `2`), so that what you compute is closer to "samples per sec", especially if the first runs for some reason take substantially longer than two seconds.

src/inference/enumerate.ad.js
```diff
@@ -17,8 +17,18 @@ module.exports = function(env) {
   function Enumerate(store, k, a, wpplFn, options) {
     util.throwUnlessOpts(options, 'Enumerate');
     options = util.mergeDefaults(options, {
-      maxExecutions: Infinity
+      maxExecutions: Infinity,
+      throwOnError: true,
+      maxRuntimeInMS: Infinity,
+      maxEnumTreeSize: Infinity
```
stuhlmueller: It's a little weird that `maxExecutions` is treated differently from `maxRuntimeInMS` and `maxEnumTreeSize`. (The former returns a partial distribution, the latter error objects.) We might want to unify their treatment in the future (but probably not in this PR).

src/header.wppl
Eric-H-Ren: CPS doesn't allow throw statement in header.wppl. Any alternatives?

null-a: [`util.fatal`](https://github.com/probmods/webppl/blob/dev/src/util.js#L289) wraps `throw`, you could use that.

src/inference/smc.js
Eric-H-Ren: Like this?
```
  SMC.prototype.sync = function() {
    // Called at sync points factor and exit.
    // Either advance the next active particle, or if all particles have
    // advanced, perform re-sampling and rejuvenation.
    if (!this.atLastParticle()) {
      this.advanceParticleIndex();
      return this.runCurrentParticle();
    } else {
      this.step += 1;
      this.debugLog('***** sync :: step = ' + this.step + ' *****');

      // Resampling and rejuvenation are applied to all particles.
      // Active and complete particles are combined here and
      // re-partitioned after rejuvenation.
      var allParticles = this.allParticles();
      assert(this.particlesAreInSync(allParticles));
      return resampleParticles(allParticles, function(resampledParticles){
        assert.strictEqual(resampledParticles.length, env.coroutine.numParticles);

        var numActiveParticles = _.reduce(resampledParticles, function(acc, p) {
          return acc + (p.trace.isComplete() ? 0 : 1);
        }, 0);

        if (numActiveParticles > 0) {
          // We still have active particles, wrap-around:
          env.coroutine.particleIndex = 0;
          return env.coroutine.rejuvenateParticles(resampledParticles, function(rejuvenatedParticles) {
            assert(env.coroutine.particlesAreInSync(rejuvenatedParticles));

            var p = _.partition(rejuvenatedParticles, function(p) { return p.trace.isComplete(); });
            env.coroutine.completeParticles = p[0];
            env.coroutine.particles = p[1];
            env.coroutine.debugLog(p[1].length + ' active particles after resample/rejuv.\n');

            if (env.coroutine.particles.length > 0) {
              return env.coroutine.runCurrentParticle();
            } else {
              return env.coroutine.finish();
            }
          }.bind(env.coroutine));
        } else {
          // All particles complete.
          env.coroutine.particles = [];
          env.coroutine.completeParticles = resampledParticles;
          return env.coroutine.finish();
        }
      });
    }
  };
```
It seems to cause error in util.js at `t = t()`, unlike `getSupport`.

Eric-H-Ren: It makes `t` sometimes not a function, but as follows: 
```
[ { trace:
     { wpplFn: [Function],
       initialStore: {},
       exitK: [Function],
       baseAddress: '_170_164_169_153',
       choices: [],
       addressMap: {},
       length: 0,
       score: -0.6931471805599453,
       numFactors: 0,
       value: false,
       store: undefined,
       k: undefined },
    logWeight: 0,
    proposalBoundary: 0 } ]
```

Eric-H-Ren: I'm looking into it.

Eric-H-Ren: Fixed now.

Eric-H-Ren:
The code has been updated.

src/inference/enumerate.ad.js
Eric-H-Ren: `maxEnumTreeSize` is used at the very beginning of the enumeration, when there's no partial distributions available.

src/inference/smc.js
```diff
@@ -242,36 +254,36 @@ module.exports = function(env) {
...
+              return env.coroutine.finish();
+            }
+          }.bind(env.coroutine));
+        } else {
+          // All particles complete.
+          env.coroutine.particles = [];
+          env.coroutine.completeParticles = resampledParticles;
+          return env.coroutine.finish();
+        }
+      });
```
null-a: You can use `this` rather `env.coroutine` to access the coroutine instance if you replace the continuation with `function(resampledParticles) { /* body */ }.bind(this)`. I've used this pattern a lot, so it might be worth using it here too for consistency?

null-a:
> The code has been updated.

Looks great! I've added one more comment about a minor cosmetic issue, but aside from that I think we're good to merge.

src/inference/smc.js
Eric-H-Ren: OK, I think can do this to `getSupport` as well.

Eric-H-Ren:
Have changed `env.coroutine` to `this`.

null-a:
@hereneric Thanks for making all of these changes, this looks good to me. Thanks for your patience too. :)

hereneric:
@null-a Thank you for your guidance Paul!

Issue #672, opened 2016-10-12
Add laplace distribution


ngoodman:
...because it makes sparse regression easy to express. This should be straightforward since it's basically two exponentials, and we do have Exponential already.

Also we should add re-parametrizations for Laplace and Exponential for ELBO (might already for the latter?).

dritchie:
Yep, Exponential already supports the reparameterization trick.

null-a:
A distribution over tensors where each element is independently Laplace distributed might also be useful. i.e. The Laplace analog of `TensorGaussian`.

Issue #630, opened 2016-08-16
Clean-up distribution parameter names


null-a:
> eg “a” is a bad param name….

Try to avoid inscrutable names, also check our names are somewhat similar to those used on Wikipedia since we link there.

This came up in the webppl meeting 16 Aug.

PR #783, opened 2017-02-22
Add Laplace and TensorLaplace Distribution

jpchen:
This addresses: #672. 

I derived the score from the [wiki page](https://en.wikipedia.org/wiki/Laplace_distribution), but the tests don't cover things like `score` or `transform`, so I'm looking for a good way to do so.

I will also add this to the documentation.


null-a:
This looks pretty close to me!

One additional thing we should do is add the new sampling helpers to [this list](https://github.com/probmods/webppl/blob/28e3ad25fc42dbbb045f95f0758fb387fdeecd27/src/transforms/caching.js#L12).

> but the tests don't cover things like score or transform, so I'm looking for a good way to do so.

If we also your run your new test cases with ELBO optimization, then we'll be indirectly testing `score` and reparameterization. (As well as our ability to generate mean field guides for these distributions.) And while ideally we might like to do more, this will be inline with what we have for most of the other distributions.

However, I'd guess that such a test won't pass yet, because it doesn't look like we can backprop through [`Math.abs()`](https://github.com/dritchie/adnn/blob/92d2672be854eebdfa787447fc638ab0f3a61dd4/ad/derivatives.js#L136). So we'll need to address that too I expect.

src/dists.ad.js
```diff
@@ -517,6 +517,44 @@ var DiagCovGaussian = makeDistributionType({
   }
 });
 
+function laplaceSample(mu, b) {
+  // Generated from goo.gl/3BxCGd (wiki)
+  var z = util.random();
+  var u = (1 - z) * -0.5 + z * 0.5;
```
null-a: Simplifies to `u=z-0.5`?

src/dists.ad.js
```diff
@@ -517,6 +517,44 @@ var DiagCovGaussian = makeDistributionType({
...
+  score: function(val) {
+    return laplaceScore(this.params.mu, this.params.b, val);
+  },
+  base: function() {
+    return new Laplace({mu: 0, b: 1});
+  },
+  transform: function(x) {
+    var mu = this.params.mu;
+    var b = this.params.b;
+    return ad.scalar.add(ad.scalar.mul(b, x), mu);
```
null-a: You could use the ad transform on this method allowing you to write `return b * x + mu;`.

src/dists.ad.js
```diff
@@ -717,6 +751,59 @@ var TensorGaussian = makeDistributionType({
...
+  while (n--) {
+    x.data[n] = laplaceSample(mu, b);
+  }
+  return x;
+}
+
+function tensorLaplaceScore(mu, b, dims, x) {
+  var _x = ad.value(x);
+  var scOp = ad.scalar;
+  var tnOp = ad.tensor;
```
null-a: * You could use the ad transform here, then use `*/+-` for the scalar ops. This will also fix what looks like a bug in the current implementation, that arises from doing `2*b` when `b` is an ad graph node.
* `T` is possible already bound to `ad.tensor` here.

src/dists.ad.js
```diff
@@ -517,6 +517,44 @@ var DiagCovGaussian = makeDistributionType({
...
+function laplaceScore(mu, b, x) {
+  'use ad';
+  return -1 * (Math.log(2 * b) + Math.abs(x - mu) / b);
+}
+
+var Laplace = makeDistributionType({
+  name: 'Laplace',
+  desc: 'Distribution over ``[-Infinity, Infinity]``',
+  params: [{name: 'mu', desc: 'mean', type: types.unboundedReal},
+           {name: 'b', desc: 'scale', type: types.positiveReal}],
```
null-a: I wonder whether we should call the parameters `location` and `scale` instead of `mu` and `b`? (I'm thinking about #630 here.)

dritchie:
FYI, I just published a patch to adnn that provides derivatives for abs().

src/dists.ad.js
jpchen: Done

src/dists.ad.js
jpchen: Done, should I do this for the other ones as well? The style seems to be split among distributions (eg :517 or :615)

src/dists.ad.js
jpchen: Discussed on slack. Changed for Laplace, left the other ones, though I could easily change them for consistency if this is the new convention

src/dists.ad.js
jpchen: Oop took that from `uniform()`. Fixed

src/dists.ad.js
null-a: > The style seems to be split among distributions (eg :517 or :615)

Yeah, I think that's just because we wrote lots of these by hand before we had the transform.

src/dists.ad.js
null-a: Cool. I think you need to run `grunt generate-docs` to update the documentation.

src/transforms/caching.js
```diff
@@ -33,6 +34,7 @@ var cacheExempt = [
   'poisson',
   'randomInteger',
   'tensorGaussian',
+  'tensorLaplacian',
```
null-a: These need to match the name of the (auto-generated) sampling helpers, so `laplace` and `tensorLaplace`.

tests/test-inference.js
```diff
@@ -29,6 +29,8 @@ var tests = [
       uniform: { args: { samples: 10000 } },
       beta: true,
       exponential: true,
+      laplace: { args: { samples: 10000 } },
+      tensorLaplace: true,
```
null-a: Are you OK to add these to the `OptimizeELBO` tests too? (You'll probably have to fiddle around with the inference args to get them passing reliably without having super wide tolerances.)

src/transforms/caching.js
jpchen: Good catch! not sure what I was thinking...

tests/test-inference.js
jpchen: Done.

tests/test-inference.js
```diff
@@ -638,6 +640,23 @@ var tests = [
           verbose: false
         }
       },
+      tensorLaplace: {
+        std: { tol: 0.15 },
+        args: {
+          stepSize: 0.002,
```
null-a: `stepSize` is a property of the optimization method, so you'll need something like `{optMethod: {adam: {stepSize: 0.002}}}` here.

null-a:
> adjust tolerance. passes ~90% of the time

Individual tests probably need to be at more like 99.99%, otherwise the test suite as a whole will fail too frequently.

null-a:
At some point we probably want to figure out what a sensible default guide for `TensorLaplace` is. At the moment it will be guided by a `TensorLaplace`, which isn't very useful since its location and scale are shared across all dims. One option is to add the equivalent of `DiagCovGaussian` giving us location and scale per dim. Or perhaps using `DiagCovGaussian` itself will be a better choice? Thoughts anyone? If we don't do this as part of this PR, we can add it to #573.

jpchen:
Added it to the linked issue for tracking and to not block the PR. Will sync up on slack on the best guide dist to use.

PR #794, opened 2017-03-15
Make observe pass options to sample

### LINKED ISSUE #787: observe should take options to pass on to sample
I find myself using this replacement for `observe` all the time:

```js
var observe = function(dist, val) {
  if (val !== undefined) {
    factor(dist.score(val));
    return val;
  } else {
    return sample(dist, { guide() {
      return dist;  // don't use auto-guide; don't complain about no guide
    }});
  }
};
```

It would be preferable to do the same thing via options passed to built-in `observe`.

stuhlmueller:
Would close #787.


jpchen:
is there a test for this?

null-a:
Looks good to me.

If we did want a test, I suppose we could add something like this monstrosity:

```js
// expected to return 'ok' with this pr
sample(Infer({method: 'forward', samples: 1, guide: true, model() {

  return observe(
    Discrete({ps: [1]}),
    undefined,
    {guide() { return Delta({v: 'ok'}); }});

}}));
```

PR #801, opened 2017-03-27
Make param work from within nested Enumerate.

### LINKED ISSUE #789: Enumerate with model parameters
Consider the following, which does MLE for the parameter `p`:

```js
Optimize({
  steps: 10000,
  model() {
    var dist = Bernoulli({p: Math.sigmoid(param({name: 'p'}))});
    observe(dist, true);
    observe(dist, false);
    observe(dist, false);
  }
});
Math.sigmoid(param({name: 'p'})); // => 0.33
```

Intuitively, it seems reasonable to expect that we can replace the `Bernoulli` distribution with a marginal distribution computed by `Enumerate`, without changing the result. Like so:

```js
Optimize({
  steps: 10000,
  model() {
    var dist = Enumerate(function() {
        sample(Bernoulli({p: Math.sigmoid(param({name: 'p'}))}));
    });
    observe(dist, true);
    observe(dist, false);
    observe(dist, false);
  }
});
Math.sigmoid(param({name: 'p'})); // => ??
```

However, this _does not_ compute the same thing as the first example at present. The problem is that `param` will only "notify" `Optimize` about parameters encountered during the evaluation of a model when `Optimize` is the current coroutine. Here, `Enumerate` is the current coroutine, so `param` assumes that we're not doing optimization, and `Optimize` is not aware of these parameters. I think there's a simple fix that will make this particular example work -- instead of only checking the current coroutine we can walk the entire coroutine stack looking for `Optimize`. This also seems like a reasonable thing to do in general.

(Katherine actually ran into essentially this problem, where `Enumerate` came from a RSA sub-model, and the parameters belong to a neural net.)

I think there are a couple of further problems related to this:

1. If `name` is omitted from `param`, we default to using something like _"the bit of the stack address that was added since entering the current coroutine"_ as the parameter name. That's a problem here, since names will be relative to `Enumerate` and not `Optimize`, inducing unexpected parameter sharing. The most direct fix I've thought of for this particular example is to base these relative stack addresses on the inner most coroutine, ignoring `Enumerate`. (i.e. Walk the coroutine stack until we find something that isn't `Enumerate`.) But this seems kinda odd, and might not be the right thing to do in general.
2. We can't swap `param` for `modelParam` in the example above. The problem is that `modelParam` will try to sample from the (continuous) prior from within `Enumerate`, which will generate an error since `Enumerate` only works with discrete choices. I'm not sure what to do about this yet.

null-a:
This makes `param` within `Enumerate` within `Optimize` work as I think we expect it to. (Closes #789.)

There are two parts to this:

1. Previously, `param` would only notify `Optimize` that a parameter should be updated if the current coroutine was one of the gradient estimators. Therefore, parameters created within `Enumerate` were not optimized. The fix implemented here is to have `param` walk the coroutine stack from the current coroutine out, looking for a coroutine that is doing optimization. If such a coroutine is found, the parameter is added to its list of "parameters to be optimized". (`paramsSeen` in code.)

2. Something similar applies to automatically generated parameter names. Previously the part of the stack address added since entering the inner most coroutine was used as the default parameter name. (I call this a relative address.) Now, we walk the coroutine stack looking for a coroutine with the `isParamBase` flag set, and use its entry address as the point from which to generate a relative address. In this way, we step over `Enumerate` when generating parameter names.

The test I've added requires both of these bits of functionality in order to pass.

Note that this PR is only intended to address the specific (and important) case of nesting `Enumerate` within `Optimize`, as identified in #789. It definitely isn't a complete solution to the problem of combining `param`, `Optimize` and `Infer` in arbitrary ways.


src/params/params.js
```diff
@@ -116,6 +116,35 @@ function fetch(name, env) {
   }
 }
 
+function findCoroutine(predicate, coroutine) {
+  if (predicate(coroutine)) {
+    return coroutine;
+  } else if (_.has(coroutine, 'coroutine')) {
+    return findCoroutine(predicate, coroutine.coroutine);
```
stuhlmueller: - I wonder if we should rename `.coroutine` to `.parentCoroutine`, or something like that?
- I also wonder if there are any circumstances where `findCoroutine` might run into an infinite loop (with two coroutines having each other as parent).

stuhlmueller:
Looks good—thanks for the detailed description!

src/params/params.js
null-a: > I wonder if we should rename .coroutine to .parentCoroutine, or something like that?

I agree.

> I also wonder if there are any circumstances where findCoroutine might run into an infinite loop (with two coroutines having each other as parent).

Since we use this like a stack, pushing the current coroutine on entry, popping on exit, I think we're good?

Issue #709, opened 2016-11-18
Default params for optimize


ngoodman:
seem like adam with step 0.01 is better than the current default in most of the little models i've tried. change?

also, change default number of forward samples to 100 (or something greater than 1)?

dorarad:
Regarding the stepSize - I also got significant improvement with 0.01, especially for dream..

null-a:
I've also noticed that small models typically benefit from a larger step size, so if that's the use case we're optimizing for, increasing it seems reasonable to me. Some [history](https://github.com/probmods/webppl/issues/548#issuecomment-234958465).

BTW, we currently inherit the default `stepSize` from adnn.

ngoodman:
wait -- is the default adam with 0.001? (i thought default was sgd.) if so then maybe i'm fine as is...

dorarad:
What about the default number of forward samples? is 1 good?

ngoodman:
right, that was confusing. should be 100 probably.

dorarad:
@ngoodman is there a reason not to change default from 0.001 to 0.01? In all cases I checked 0.01 was better. @null-a changed the default some time ago from 0.1 to 0.001 (dritchie/adnn@4f4dba2) - I guess since indeed 0.1 is unstable. But 0.001 seems to be too low..

Issue #605, opened 2016-07-29
OptimizeELBO - guidedGaussian test failed


stuhlmueller:
On 1c3cd24147bbeede50d42f24ab666d64a9ac41f9:

~~~~
Running "nodeunit:all" (nodeunit) task
Testing test-caching.js..............OK
Testing test-deterministic.js.................OK
Testing test-error-handling.js...OK
Testing test-examples.js..............OK
Random seed: 1393378568
Testing test-inference.js..............................................................................................................................................................................................................................................F..
>> OptimizeELBO - guidedGaussian
>> Message: Expected mean: -4, actual: -3.4735990233865213, tolerance: 0.2
>> Error: false == true
>> at Object.testWithinTolerance (tests/helpers/helpers.js:56:8)
>> at Object.testFunctions.mean (tests/test-inference.js:644:13)
>> at /Users/stuhlmueller/Projects/webppl/tests/test-inference.js:624:28
>> at Function._.each._.forEach (node_modules/underscore/underscore.js:158:9)
>> at Object.performTest (tests/test-inference.js:615:5)
>> at executeBound (node_modules/underscore/underscore.js:701:67)
>> at Object.bound (node_modules/underscore/underscore.js:733:14)
~~~~

null-a:
Also [158225176](https://travis-ci.org/probmods/webppl/jobs/158225176).

null-a:
We probably want to bump up the step size in the ELBO tests, after which we'll probably be able to reduce some of samples per step/steps/tolerance, which all seem quite high now I'm looking at it again.

dorarad:
It may be a good idea to change the default in adam from 0.001 to 0.01. In experiments I've done so far 0.01 was consistently better..@null-a you changed the default in July from 0.1 to 0.001 (https://github.com/dritchie/adnn/commit/4f4dba21f498f929e5626876fa94d8ff78a3400c). Indeed 0.1 causes many cases of divergence, but I think 0.001 is too small..

null-a:
@dorarad We also discussed this [here](https://github.com/probmods/webppl/issues/709#issuecomment-261751870).

dorarad:
@paul right. Did you have a case in which 0.001 was better?

jpchen:
I ran this ~10 times it seems to pass... `tol = 0.2` though, not sure if we're still looking to change the default step size?

null-a:
Also [228774965](https://travis-ci.org/probmods/webppl/jobs/228774965).

PR #808, opened 2017-04-04
Have rejection and forward sampling take 100 samples by default.

null-a:
In #709 it was suggested that the number of samples taken when forward sampling should be changed from 1 to 100. This PR does that.

Once forward sampling is updated, rejection would be the only inference algorithm left that doesn't take 100 samples by default, so I updated it as well.


null-a:
CI failure was same as #605.

PR #815, opened 2017-04-10
Add forward and forwardGuide helpers.

null-a:
I sometimes find myself writing:

`sample(Infer({model, method: 'forward', samples: 1, guide: true}))`

... which is a bit verbose. This PR abstracts that as a `forwardGuide(model)` helper for convenience. For completeness, it also adds an analogous `forward` helper for forward sampling from the target. Do you think these are these worth adding?


stuhlmueller:
Looks good, I think this a useful shorthand.

PR #816, opened 2017-04-12
Update dev dependencies & switch to eslint

### LINKED ISSUE #225: Switch to eslint

### LINKED ISSUE #463: Upgrade dependencies to eliminate deprecation warnings
```
npm WARN deprecated graceful-fs@1.2.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.
npm WARN deprecated lodash@1.0.2: lodash@<3.0.0 is no longer maintained. Upgrade to lodash@^4.0.0.
```

null-a:
This addresses #225 and #463.

The eslint configuration is based on a builtin [recommended configuration](http://eslint.org/docs/user-guide/configuring#using-eslintrecommended), with some customisations to make it better fit our current conventions. (To avoid having to make too many changes at once.) This seems like an OK starting point, though we may choose to tweak this over time of course. This could include:

* Enabling the [`no-redeclare`](http://eslint.org/docs/rules/no-redeclare) and [`no-used-vars`](http://eslint.org/docs/rules/no-unused-vars) rules. I turned these off as we'll need to make several changes the code to have them pass. (These are part of the configuration @stuhlmueller suggested we might use in #225, so this would be a step towards using that.)
* Adding the [`strict`](http://eslint.org/docs/rules/strict) rule.

Note that we now need to run `grunt lint --fix --force` to attempt to automatically fix issues identified by the linter, rather than `grunt fixjsstyle`. (I couldn't see a neat way to retain the old method.)

Follow-up tasks:

* Create issues for the todo comments removed by this PR.
* Define `T` and `ad` in `dists.ad.js`, then remove these as globals from the  eslint config. (I'll do this as a follow up PR.)

Closes #225.
Closes #463.


null-a:
CI failed because eslint doesn't support Node 0.12. To fix this, we could arrange for the linter to only run on node >= 4, or if it's not too painful, we could just drop support for node 0.12 from WebPPL. Any thoughts?

stuhlmueller:
I don't know of good reasons to keep support for node v0.12. A reason to drop support for older versions is that the more recent versions support a larger subset of ES6, so if we want to, we can start using it (for implementation, not webppl programs).

stuhlmueller:
Looks good, thanks! I'm glad we're finally using a non-ancient linter.

PR #822, opened 2017-04-18
Add file-based param store backend

### LINKED ISSUE #768: Add file-based param store backend
To simplify single-process parameter optimization with persistence across runs, it could be useful to have a param store backend that reads from a file on startup, behaves like the in-memory store at runtime, and writes to the same file on shutdown.

null-a:
* This raises the question of whether we still need the `checkpointParams` functionality that's baked into `Optimize`?

* One thing you can't do with the file store which you can do if you know `setParams` exists is read in parameters, do some optimization, and *not* write any updated parameters to disk once the program terminates. I wonder if some kind of read only mode might be a useful extension to this?

Closes #768.

### DIFF
--- a/docs/optimization/filestore.rst
+++ b/docs/optimization/filestore.rst
@@ -1,0 +1,31 @@
+.. highlight:: none
+
+Persistence
+===========
+
+The file store provides a simple way to persist :ref:`parameters
+<parameters>` across executions. Parameters are read from a file
+before the program is executed, and written back to the file once the
+program finishes. Enable it like so::
+
+  webppl model.wppl --param-store file --param-id my-parameters
+
+The file used takes its name from the ``param-id`` command line
+argument (appended with ``.json``) and is expected to be located in
+the current directory. A new file will be created if this file does
+not already exist.
+
+An alternative directory can be specified using the
+``WEBPPL_PARAM_PATH`` environment variable.
+
+A random file name is generated when the ``param-id`` argument is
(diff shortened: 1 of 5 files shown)


stuhlmueller:
Looks good. I think this will be very useful.

It could be helpful to add a comment to `getParams` saying that the reason we don't re-read from the file is that we don't support parallel use.

As far as I can tell, `checkpointParams` is now obsolete. The only reason I see why someone might still want it is if they wanted to use the Mongo store, but also occasionally save the params to disk. This doesn't seem like a strong-enough reason to keep it. Do you see others?

I agree that a read-only mode could be useful.

In general I feel that we haven't quite figured out the ideal way to select and parameterize parameter stores yet. Evidence: 

1. Customizing (only) via environment variables doesn't feel user-friendly.
2. We might want to combine stores (similar to how we combine MCMC kernels), but this would be awkward to specify under the current system (if it has to happen via command-line args and environment variables).
3. I'm still unclear on what part of the parameters interface should be accessible from within WebPPL, and what part should only be accessible at startup time.

null-a:
> It could be helpful to add a comment to getParams saying that the reason we don't re-read from the file is that we don't support parallel use.

See #825.

> As far as I can tell, checkpointParams is now obsolete.

See #826.

> In general I feel that we haven't quite figured out the ideal way to select and parameterize parameter stores yet.

Agreed.

PR #826, opened 2017-04-19
Remove checkpointParams option from Optimize.

null-a:
This is a follow up to #822.

>> This raises the question of whether we still need the checkpointParams functionality that's baked into Optimize?

> As far as I can tell, checkpointParams is now obsolete. The only reason I see why someone might still want it is if they wanted to use the Mongo store, but also occasionally save the params to disk. This doesn't seem like a strong-enough reason to keep it. Do you see others?

I suppose one thing you might like to do is read the initial parameters from one file, and then save the updated parameters to some other file. This is a little bit fiddly with the file store since it always reads and writes from the same file, so we'd probably suggest making a copy of the file containing the initial parameters before running the program to achieve something similar. But this doesn't seem like a strong enough reason to keep it around either?

### DIFF
--- a/src/inference/optimize.js
+++ b/src/inference/optimize.js
@@ -56,11 +56,7 @@ module.exports = function(env) {
 
       logProgress: false,
       logProgressFilename: 'optimizeProgress.csv',
-      logProgressThrottle: 200,
-
-      checkpointParams: false,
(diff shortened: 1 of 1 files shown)


stuhlmueller:
Agreed. Let's add "read/write different files" to the list of things that would be desirable for the next iteration of the parameter store interface.

PR #827, opened 2017-04-19
Handle multimodality in sampler tests

null-a:
This addresses the TODO notes removed by #816.


longouyang:
Looks good. Could you add back somewhere my explanation for the reason that we can't test modes for binomial?

null-a:
> Could you add back somewhere my explanation for the reason that we can't test modes for binomial?

Sure, done.

PR #837, opened 2017-04-27
Only store a single tensor per parameter name

null-a:
Before this change, each parameter name had associated with it an array of tensor valued parameters. Previously, this was used as the basis for supporting the use of adnn networks in WebPPL programs. However, this [isn't something we currently support](https://github.com/probmods/webppl/issues/663#issuecomment-272113800), so these arrays are redundant.

The only potential problem I see with this is that after the change, it won't be possible to load parameters persisted to file/mongo using the old format. I'm not sure this is worth worrying about though?

### DIFF
--- a/src/inference/dream/gradients.ad.js
+++ b/src/inference/dream/gradients.ad.js
@@ -45,9 +45,7 @@ module.exports = function(env) {
           objective.backprop();
         }
 
-        var grads = _.mapValues(this.paramsSeen, function(params) {
-          return params.map(ad.derivative);
-        });
+        var grads = _.mapValues(this.paramsSeen, ad.derivative);
 
         return cont(grads, -ad.value(this.logq));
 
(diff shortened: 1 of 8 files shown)


null-a:
Another potential problem is that it's possible that code has been written that depends on this internal structure. e.g. `getParams()[someName][0]`. Such code would need updating after this change, in this case to `getParams()[someName]`. This isn't common practice as far as I'm aware, so I imagine we're OK to go ahead, and maybe just mention this in the release notes.

stuhlmueller:
Nice, this makes a lot of code more readable!

> The only potential problem I see with this is that after the change, it won't be possible to load parameters persisted to file/mongo using the old format.

> Another potential problem is that it's possible that code has been written that depends on this internal structure. e.g. `getParams()[someName][0]`. Such code would need updating after this change, in this case to `getParams()[someName]`.

I don't think we should worry about either of these. If necessary, people can always resort to the prior version of webppl to access stored parameters and run models that rely on the structure of the params objects. (I don't know that anyone besides me has made use of this functionality anyway.)

PR #844, opened 2017-05-03
Show warning when Infer/Optimize receives unexpected option

### LINKED ISSUE #542: Show warning when Infer/Optimize receive unexpected parameter
This is useful because unexpected parameters likely arise because of typos, mis-remembering param names, interface changes etc. Sometimes the effects are subtle and go unnoticed. A warning would help guard against this.

For example `Infer({method: 'enumerate', maxexecutions: n}, model)` should print something like `Infer: Unrecognized option "maxexecutions"`.

null-a:
Closes #542.

For example, running this:

```js
Infer({model, method: 'enumerate', maxexecutions: 10});
```

... now generates the following warning:

```
Warning: Unused option "maxexecutions" given to Enumerate.
```

Similar warnings are generated for `Optimize`, and for the options passed to MCMC kernels and gradient estimators. Note that the options passed to optimization methods are not checked in this way, as these are implemented in adnn rather than WebPPL.

The most contentious part of the implementation is the introduction of an additional wrapper function around kernels & estimators ([example](https://github.com/probmods/webppl/blob/0f54d4d1a00fb446ba9498a15bb70937b147f14e/src/inference/hmckernel.js#L23-L34)), as a way of arranging for the checking of options/args to only occur once. (So that we only do the work of performing the checks and issue warnings once, rather than on each step.) I'm happy with this, but there are alternatives of course, and I'm happy to discuss those if this is unsatisfactory.


src/inference/kernels.js
```diff
@@ -49,11 +50,10 @@ module.exports = function(env) {
 
     var name = _.isString(obj) ? obj : _.keys(obj)[0];
     var options = _.isString(obj) ? {} : _.values(obj)[0];
-    var kernel = kernels[name];
+    var kernel = kernels[name](options);
 
-    return _.assign(function(cont, oldTrace, extraOptions) {
-      var allOptions = _.assign({}, options, extraOptions);
-      return kernel(cont, oldTrace, allOptions);
+    return _.assign(function(cont, oldTrace, runOpts) {
```
null-a: You're right, it's now redundant, well spotted!

src/inference/hmckernel.js
```diff
@@ -20,18 +20,28 @@ var sum = generic.sum;
 
 module.exports = function(env) {
 
-  function HMCKernel(cont, oldTrace, options) {
-    var options = util.mergeDefaults(options, {
+  function kernel(options) {
```
null-a: Sure, let's go with that.

PR #851, opened 2017-05-25
Make `cache` callable as `dp.cache`

### LINKED ISSUE #850: Make cache work with mutual recursion
As an important special case of #61, we should make `cache` work in the presence of mutual recursion. This is currently feasible using [a package](https://github.com/stuhlmueller/webppl-dp), but should be integrated into core.

null-a:
Closes #850.


stuhlmueller:
Thanks, especially for the added tests + docs!

Issue #836, opened 2017-04-23
Compile webppl-nn package into homepage


stuhlmueller:

null-a:
The webppl-nn package has now been merged into core WebPPL.

PR #854, opened 2017-06-01
Merge webppl-nn

### LINKED ISSUE #710: Neural Net Helpers

null-a:
This merges [`webppl-nn`](https://github.com/null-a/webppl-nn) into core. I've added a few basic tests and tried to improve the documentation along the way.

Closes #710, #836.

@jpchen: Would merging this now cause problems with anything you're working on?


jpchen:
Hm it would, but only slightly. I have a working webppl branch and a separate webppl-nn and moved some function calls like `relu()` and `lrelu()` to AD so that `nn` can be agnostic to Tensor type. Temporarily, people can just use that branch and `require` webppl-nn in the old way.
To not block anyone, this is good for merging by me, and I can just change the necessary functions in util.js after.

Side note: did you take out leaky relu? And for relu, is `nn.relu.eval(x)` any different than the `relu` [operation](https://github.com/null-a/webppl-nn/blob/master/src/nn.js#L39) in webppl-nn?

null-a:
> did you take out leaky relu?

Yes. (l)relu seemed like something that would be better in adnn (or its successor) than webppl. I think we should add the leaky version to adnn (and add the trivial wrapper to webppl) if we want it.

> And for relu, is nn.relu.eval(x) any different than the relu operation in webppl-nn?

When writing webppl-nn I'd missed that adnn already had relu. It's only exposed as a "network" rather than an AD op, hence the `eval` call, but they're the same function.

PR #858, opened 2017-06-27
Expand docs on conditioning

### LINKED ISSUE #602: add and document observe and condition helpers
while `factor` is the most general form, in practice it is often more pleasant to use `condition` (already in header) or `observe`. in addition, `observe` yields better inference results for many algorithms. we should make sure both are implemented and documented.

(by `observe` i mean a helper that represents an immediate observation of a sampled value. so `observe(dist,val)` corresponds semantically to `factor(dist.score(val))` and `condition(sample(dist)===val)`. it may be useful to have a form, as in daipp, that samples when `val` is undefined and always returns the result (`val` or sampled).)

### LINKED ISSUE #655: specify log space in factor docs
The current [`factor` docs](http://webppl.readthedocs.io/en/master/inference/index.html#factor) says that it "arbitrarily weights particular executions", but doesn't explicitly mention that the weighting happens in log space. Maybe we don't want people thinking about this too carefully (thus the recent push for `observe`, saving `factor` for cases where it's really heuristic), but there's still a lot of papers in print that use this fact to implement principled exponential relationships and it's created some confusion going between the code and the mathematical formula, e.g.

```
factor(alpha * literalListener(utterance,qud).score(qValue))
```

### LINKED ISSUE #665: documentation on conditioning
Hi guys, overall documentation and quickstart from rwebppl was awesome. Thank you! 

The one comment I had on the documentation was that the [section on conditioning](http://webppl.readthedocs.io/en/master/inference/index.html#factor) was awful brief (I spoke with @ngoodman and he encouraged me to pass on this comment). In particular, there was no discussion of the syntax for `condition` or `observe` and even for someone with some probabilistic modeling experience, I found the discussion of `factor` a little bit terse. The guiding principle of "insert your conditioning statement in the generative model" is a little bit unintuitive and so new users could likely benefit from a bit more guidance here.

null-a:
These aren't perfect, but I think they're an improvement on what we have now as I've tried to address issues #602, #655, and #665.

Closes #602
Closes #655
Closes #665

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -21,6 +21,7 @@ WebPPL Documentation
    sample
    distributions
    inference/index
+   conditioning
    optimization/index
    functions/index
    globalstore
--- a/docs/inference/conditioning.rst
+++ b/docs/inference/conditioning.rst
@@ -1,0 +1,63 @@
+.. _conditioning:
+
+Conditioning
+============
+
+Conditioning is supported through the use of the ``condition``,
+``observe`` and ``factor`` operators. Only a brief summary of these
+methods is given here. For a more detailed introduction, see the
+`Probabilistic Models of Cognition chapter on conditioning
+<https://probmods.org/chapters/03-conditioning.html>`_.
+
+Note that because these operators *interact* with inference, they can
+only be used *during* inference. Attempting to use them outside of
+inference will produce an error.
+
+.. js:function:: condition(bool)
+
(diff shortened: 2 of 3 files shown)


null-a:
CI failure is unrelated to this change, and addressed by #859.

stuhlmueller:
That is much clearer, thanks!

PR #870, opened 2017-07-06
Warn when impossible observations are made.

### LINKED ISSUE #621: Have observe issue a warning if factor score is -Infinity
Such situations are likely to be caused by a bug that the user would like to know about. I think?

null-a:
For example, this:

```js
observe(Uniform({a: 0, b: 1}), 2);
```

... now generates the following warning:

```
Warning: One or more values given to observe have a
score of -Infinity under the Uniform distribution.
```

One thing that's not so great about this is that it may not be clear where in the program this is coming from. The easiest way I can think of to address this would be to turn the warning into an error so that we get a "stack trace", but I'm not sure that this would always be desirable?

Note that any coroutine supplying its own `observe` implementation will have to generate a similar warning if appropriate. Currently only dream does this, and I don't think it requires this warning.

Closes #621.

### DIFF
--- a/src/header.js
+++ b/src/header.js
@@ -114,7 +114,12 @@ module.exports = function(env) {
         var factorK = function(s) {
           return k(s, val);
         };
-        return env.factor(s, factorK, a, dist.score(val));
+        var score = dist.score(val);
+        if (ad.value(score) === -Infinity) {
+          util.warn('Warning: One or more values given to observe have a score of ' +
+                    '-Infinity under the ' + dist.meta.name + ' distribution.', true);
+        }
+        return env.factor(s, factorK, a, score);
       } else {
         return env.sample(s, k, a, dist, options);
       }


null-a:
CI failure is unrelated to this change.

```
npm ERR! network read ECONNRESET
npm ERR! network This is most likely not a problem with npm itself
npm ERR! network and is related to network connectivity.
```

mhtess:
I think sometimes when initializing Mcmc chains, -infinity scores are produced (when it's exploring a terrible part of the parameter space). What that also throw an error? (If so, not desirable)

null-a:
> What that also throw an error? 

Ah, yes, it would have been an error during MCMC init too. I agree this isn't what we want. It sounds like even a warning might not be so useful during MCMC init, which makes me wonder whether there aren't more places this doesn't make sense? I originally opened this issue when working with VI, where as soon as you get a score of -Infinity you're in trouble. In that setting, knowing that the problem occurred at an `observe` seemed like it might be useful. Perhaps it's not so useful elsewhere. @mhtess What use case did you have in mind?

null-a:
Closing this. Will re-open once I'm clear what the desired behaviour is.

PR #877, opened 2017-07-27
Add mixture distribution primitive

### LINKED ISSUE #713: Mixture distribution primitives
Normally, a mixture distribution can be expressed via a simple webppl program (first choose the mixture component, then sample from that component). However, there are (at least) a couple scenarios where having a "primitive" mixture distribution comes in handy:

1. The 'enumerating out' trick for reducing variance of the ELBo estimator only works with discrete mixtures. For continuous mixtures (such as GMMs), users have to do the marginalization manually. It'd be nice if this was taken care of automatically.
2. Using mixture distributions as guides (e.g. guiding a `Gaussian` with a `GaussianMixture`, b/c the posterior may be multimodal). This was useful in the neurally-guided procedural modeling work.

There may be other ways to achieve these two goals, but having primitive mixture distribution types seems like the most direct way.

For the NGPM demo I'm building right now, I'm just adding a `GaussianMixture` distribution, but it would be nice to build a more general-purpose constructor that takes a base distribution type and builds a mixture distribution out of it (also: I'd like if my demo ran under "real" webppl, and not my hacked-up version). This would either be callable in webppl code (i.e. a higher-order distribution constructor), or we'd have an npm post-install step that would statically create all the potentially useful mixture types. Thoughts on which is preferable?

null-a:
First attempt at a `Mixture` distribution primitive. Allows things like:

```js
var g1 = Gaussian({mu: 2, sigma: 1});
var g2 = Gaussian({mu: -2, sigma: 0.5});
var mix = Mixture({dists: [g1, g2], ps: [0.3, 0.7]});
var x = sample(mix);
```

The scorer is differentiable assuming the scorers of the components are, so this will work with gradient based algos. (Addressing the 2 use cases mentioned by @dritchie in #713.)

Closes #713.

### DIFF
--- a/docs/primitive-distributions.txt
+++ b/docs/primitive-distributions.txt
@@ -135,6 +135,13 @@
 
   `Wikipedia entry <https://en.wikipedia.org/wiki/Logit-normal_distribution>`__
 
+.. js:function:: Mixture({dists: ..., ps: ...})
+
+  * dists: array of component distributions
+  * ps: component probabilities (can be unnormalized) *(vector or real array [0, Infinity))*
+
+  A finite mixture of continuous distributions. The component distributions should all share a common support.
+
 .. js:function:: Multinomial({ps: ..., n: ...})
 
   * ps: probabilities *(real array with elements that sum to one)*
--- a/docs/primitive-distributions.txt
+++ b/docs/primitive-distributions.txt
@@ -140,7 +140,7 @@
   * dists: array of component distributions
   * ps: component probabilities (can be unnormalized) *(vector or real array [0, Infinity))*
 
-  A finite mixture of continuous distributions. The component distributions should all share a common support.
+  A finite mixture of distributions. The component distributions should be either all discrete or all continuous. All continuous distributions should share a common support.
 
 .. js:function:: Multinomial({ps: ..., n: ...})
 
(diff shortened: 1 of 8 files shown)


stuhlmueller:
Looks very useful!

- Why require that the supports are equal?
- Would it be difficult to extend this so that it also works for discrete distributions? (I realize that this can be achieved using enumerate, but a single unified interface for constructing mixtures seems desirable, unless there is good reason not to go that route.)

null-a:
> Why require that the supports are equal?

I had a couple of things in mind:

1. I was nervous about the discontinuity in the density of e.g. a mixture of `Gamma(1,1)` and `Uniform(-1,0)`. It's not obvious to me that such things will always work correctly with gradient based algos.
2. It's not immediately clear (to me) how best to handle a mixture of two distributions where the support of the mixture isn't a single bounded region. e.g. mixture of `Uniform(0,1)` and `Uniform(2,3)`. In particular, HMC can handle a single bounded region, but not multiple isolated regions.

The restriction to allow only equal supports side steps these, and doesn't seem very restrictive in practice. But I'm definitely happy work on relaxing this if that will be useful.

> Would it be difficult to extend this so that it also works for discrete distributions?

No, I don't think so. We'd probably just need to compute the support and make sure the continuous/discrete flag is set correctly. I'll take a look.

Issue #30, opened 2015-01-13
Support primitive functions in more contexts


stuhlmueller:
This works:
~~~~
var exp = function(x){return Math.exp(x)}
var xs = [1, 2, 3];
map(exp, xs)
~~~~

This doesn't:
~~~~
var xs = [1, 2, 3];
map(Math.exp, xs)
~~~~

It should be possible to use primitive functions in all contexts where compound (webppl) functions are supported.

PR #881, opened 2017-09-01
Clean-up language docs, add to TOC.

null-a:
My primary motivation for this is to have *something* on docs.webppl.org about which function calls are interpreted as JS function calls, and a hint about the seemingly common gotcha of trying to call `parseInt` and the like.


stuhlmueller:
Thanks, I think this will be helpful. (I think even after reading this, people might still run into #30, but this is much less common than the plain function call case.)

PR #888, opened 2017-09-26
[docs] Fix T.dot example

longouyang:
The old example no longer runs; vectors now have dimension `[m,1]` and I believe variable re-declarations (`x` is re-declared on the third line) stopped working when we upgraded some parsing stuff a while back. This should fix that.


null-a:
I doubt this has ever run, but either way, making it work seems like a good thing!

One question, how come you removed the `// instead of ad.tensor.dot(x, y)` bit, that seems helpful given the point of the example?

longouyang:
Ah, inadvertent -- added it back.

null-a:
Great, looks good to me!

PR #905, opened 2018-07-24
Add AIS

null-a:
This PR adds an implementation of annealed importance sampling. It closely follows the AIS implementation added over in [this fork](https://github.com/siddancha/webppl/tree/b607efe714d78c44f763ffd36324c0b67de96f56). I've updated that code to work with the most recent WebPPL, added some very basic tests and docs, and that's about it.

There are minor modifications to `Trace` and `MHKernel`, and everything else is contained in a new `AIS` coroutine. I don't there's anything too controversial in here.

It would be nice to make this work with kernels other than MH, but that would require more work which I'm not sure could happen any time soon. But given that this is a relatively noninvasive change, I think it's probably acceptable as is.

One possible gotcha is that because `factor` scores are annealed, this won't produce the results one might expect with models that use hard factors/`condition`:

```js
display(AIS(function() {
  var x = flip();
  var y = flip(x ? 0.8 : 0.1);
  condition(y == true);
  return x;
}));
// => 0

display(AIS(function() {
  var x = flip();
  observe(Bernoulli({p: x ? 0.8 : 0.1}), true);
  return x;
}));
// => -1.05....
```

One thing I don't like is the `AISforInfer` method I added to `header.wppl` in order to use the existing inference testing machinery to test AIS. I'd happily change this if someone can suggest a better way to structure the tests.


null-a:
After this PR, we will be storing the score of each individual choice/sample as part of the trace. (See [here](https://github.com/null-a/webppl/blob/4b2bc20b00191a1b1bb577ae897e84ad39ddd134/src/trace.js#L72).) It looks like we could re-use these (rather than re-scoring) when computing MH transition probs, saving some work. (See [here](https://github.com/null-a/webppl/blob/4b2bc20b00191a1b1bb577ae897e84ad39ddd134/src/inference/mhkernel.js#L234).) I'll open a separate PR or issue for this if this PR is merged.

stuhlmueller:
Looks good.

Add a note to the docs saying that this won't give correct estimates for models with hard conditioning? This seems like an important gotcha, since running it won't give any indication that something is wrong, it'll just give you the wrong number. (Ideally, we'd throw an error if `-Infinity` factors are encountered.)

null-a:
> Add a note to the docs saying that this won't give correct estimates for models with hard conditioning?

Done.

> Ideally, we'd throw an error if -Infinity factors are encountered.

Also done.

(Aside: It's nice that AIS reuses the MH kernel, but it's a shame that the kernel has to know about scaling factor scores and disallowing hard factors. At some point it might be interesting to investigate whether these things can be handled by a separate coroutine, which then just relays `sample`, `factor`, etc. to some other coroutine to do the bulk of the work. This would presumably make it easier to use the HMC kernel with AIS too.)

stuhlmueller:
Thanks for taking the extra step to disallow hard factors!

null-a:
Great, thanks for reviewing this PR.

PR #908, opened 2018-09-18
updated README.md to add instructions for using packages

marielajennings:
Added instructions for using packages with WebPPL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,18 @@ Run WebPPL programs:
 Upgrade WebPPL:
 
     npm update -g webppl
+## Using packages with WebPPL
+To use packages, first navigate to your home directory and create a folder named ```.webppl```. This is where WebPPL programs will find packages.
+
+To install a package, run ```npm install --prefix ~/.webppl myPackage```, where myPackage is the name of the package you would like to use. 
+
+To run a WebPPL program with a package, run the following command:
+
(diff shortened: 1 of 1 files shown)


null-a:
This looks good to me. One question -- could we avoid duplication by linking to similar information that's already in [the docs](http://docs.webppl.org/en/master/packages.html#packages)? (Or is saving users from having to click through to the docs the goal?)

marielajennings:
We used WebPPL to teach a class last semester and the students had a hard time finding this information. We wanted to have it included in the README if possible to make it more visible this time around.

null-a:
> students had a hard time finding this information

It sounds like the documentation isn't as easy to use/navigate as we'd hope. It would be great if we could improve the usability of the docs directly, rather than duplicate information to workaround the problem, but I don't know how to do that. (Suggestions welcome!) Given that, what do you think to the following as an alternative solution -- we could extend the "Quick start" section of the readme to include something like:

<hr/>

Require a WebPPL package:

```
webppl myprogram.wppl --require packageName
```
See the [packages](http://docs.webppl.org/en/master/packages.html#packages) section of the documentation for more information.

<hr/>

The idea is to fix your problem without incurring too much duplication. What do you think?

marielajennings:
That would be great, thank you!

null-a:
@marielajennings Great! Addressed in #909.

PR #909, opened 2018-09-21
Mention packages in the "quick start" section of the README.

null-a:
This is the change I proposed in #908.


null-a:
Given how minor this change is, I'm going to go ahead and merge without review. I'm very happy to revisit if anyone has objections/sees problems.

Issue #910, opened 2018-10-04
Does `LogNormal` already exist?


erindb:
We have `LogitNormal` and its multivariate generalization `LogisticNormal`, but I can't find `LogNormal`. Am I just not looking correctly? or does it actually not exist? If it doesn't exist, I'm happy to implement it. I think it would be useful for people (like me) who sometimes forget that the Log-Normal density function at x isn't just the Normal density function evaluated at log(x). (Doh!)

null-a:
No, there isn't a `LogNormal` at present. Here are a few pointers that might help make the process of adding a new distribution go a little smoother, or at least help you get started, should you choose to add one.

- Distributions live [here](https://github.com/probmods/webppl/tree/dev/src/dists). There's one file per distribution. Any new distribution would also need adding to `index.ad.js` in the same directory.
- These are `*.ad.js` files, so during development you'll need to run `grunt build` to transform them to `*.js`. See these [docs](http://docs.webppl.org/en/master/development/workflow.html#modifying-ad-js-files).
- `grunt build` also automatically creates the sampling helper functions. e.g. It creates the `gaussian(..)` helper, which is short for `sample(Gaussian(...))`.
- The documentation for a distribution exists as properties on the distribution object itself. Running `grunt generate-docs` will extract this and update the documentation. This needs to be run manually once the distribution's docs are finalized, and the resulting changes to [this file](https://github.com/probmods/webppl/blob/dev/docs/primitive-distributions.txt) should be committed to the repo.
- The `base` and `transform` properties of a distribution are how we implement the reparameterization trick for variational inference. These could be omitted initially.
- See e.g. `gamma.ad.js` for how to indicate that the log normal has bounded support. (This information is used by the HMC sampler.)
- There's some infrastructure for testing samplers/scorers [here](https://github.com/probmods/webppl/tree/dev/tests), see `test-samples.js` and `test-scoreres.js`. I'm only pointing this out in case it's useful. Not all samplers are exercised by this yet, and I've not used it myself.

I'm happy to expand on this, answer other questions, etc. Good luck!

ngoodman:
@null-a those are awesome pointers! maybe add them to the readme (or some dev docs linked from there) for future reference?

PR #911, opened 2018-10-30
add lognormal

erindb:
#910

### DIFF
--- a/src/dists/logNormal.ad.js
+++ b/src/dists/logNormal.ad.js
@@ -1,0 +1,55 @@
+'use strict';
+
+var ad = require('../ad');
+var base = require('./base');
+var types = require('../types');
+var util = require('../util');
+var numeric = require('../math/numeric');
+var gaussian = require('./gaussian');
+
+var LOG_2PI = numeric.LOG_2PI;
+
+function score(mu, sigma, x) {
+  'use ad';
+  var logx = Math.log(x);
+  return (
+    -logx - Math.log(sigma) - 0.5 * (
+      LOG_2PI + (logx - mu) * (logx - mu) / (sigma * sigma)
+    )
+  )
+}
+
+function sample(mu, sigma) {
+  var gaussian_sample = gaussian.sample(mu, sigma);
+  return Math.exp(gaussian_sample);
+}
+
+var LogNormal = base.makeDistributionType({
+  name: 'LogNormal',
+  desc: 'Distribution over non-negative reals where the log of the random ' +
+    'variable is normally distributed.',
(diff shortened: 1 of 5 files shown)


null-a:
This is great! The sampler, scorer, tests, docs, all look good to me. I can see just one subtle issue that needs addressing before this is merged. When using gradient based inference the parameters of a distribution can be AD values rather than regular JS numbers, which at present the logNormal sample doesn't handle. For example, the following program will generate an error at present:

```js
var model = function() {
  return logNormal(0, 1);
};

Infer({model, method: 'optimize'});
```

Thankfully the fix is easy, you just need to use `ad.value(...)` to get the underlying values out of the AD objects before calling the sampling routine. I'll add a comment inline to indicate where this ought to happen.

It would also be good to have a test for this. The easiest way to achieve this is by adding a new inference test. You'd need to add two new files (one for the model and one to describe the expected marginal) and also an entry to the 'OptimizeELBO' section of `test-inference.js`. Something analogous to this exponential test would work:

https://github.com/null-a/webppl/blob/dev/tests/test-data/stochastic/models/exponential.wppl
https://github.com/null-a/webppl/blob/dev/tests/test-data/stochastic/expected/exponential.json
https://github.com/null-a/webppl/blob/c757f25adfc99bf0da88d02987beb26460974703/tests/test-inference.js#L666

One final thing, it would be great to add reparameterisation to this at some point. If you're interesting in doing that here or as a follow-up and could use some pointers I'd be happy to help.

src/dists/logNormal.ad.js
```diff
@@ -0,0 +1,55 @@
...
+var LogNormal = base.makeDistributionType({
+  name: 'LogNormal',
+  desc: 'Distribution over non-negative reals where the log of the random ' +
+    'variable is normally distributed.',
+  params: [{name: 'mu', desc: 'mean', type: types.unboundedReal},
+           {name: 'sigma', desc: 'standard deviation', type: types.positiveReal}],
+  wikipedia: 'Log-normal_distribution',
+  mixins: [base.continuousSupport],
+  sample: function() {
+    return sample(this.params.mu, this.params.sigma);
```
null-a:
```suggestion
    return sample(ad.value(this.params.mu), ad.value(this.params.sigma));
```

`ad.value(...)` get the under lying value out of an AD node. If its argument is a number, it acts like the identity. This should allow the example in my comment to run without error.

null-a:
@erindb Thanks for adding the `ad.value` fix.

erindb:
Thanks! Sorry I never got around to the other stuff! :)

PR #917, opened 2019-01-11
Add inference tests for logNormal.

null-a:
The motivation is to ensure that logNormal handles ad nodes as parameters. This test would have caught the bug we spotted during #911.


null-a:
I'll go ahead and merge this PR without review since it only adds tests, and I therefore judge doing so to be low risk.

PR #919, opened 2019-01-29
Add inference callbacks to support streaming samples to disk.

null-a:
@mhtess has long wanted to be able to stream samples to disk rather than accumulate them in memory. This PR aims to make that possible by adding a new callback to our sampling based inference methods that is called whenever a sample is collected. With this, it's possible to implement the desired functionality as [a simple package](https://github.com/null-a/webppl-sample-writer/tree/use-sample-callback). This package would be used in conjunction with the `onlyMAP` option so that WebPPL doesn't keep all collected samples in memory.

**Other approaches**

Obviously this isn't the only way to go about this. A more direct approach would be to have inference take a new parameter, `saveSamples` say, that accepts a filename and arranges for the samples to be written to disk. The actual work could happen in `SampleBasedMarginal`. While this would remove the need for the additional package, I'm inclined to favor the callback based approach because it's simpler on the WebPPL side, and this functionality could plausibly be used for other things. Let me know if you feel otherwise.

**Performance**

I've included in this PR a commit which reduces the overhead associated with the previous callback implementation. Previously invoking a hook would always involve traversing a (tiny) array, even when there was no work to do, whereas now we only make a call to `_.noop` in such cases. This should ensure that the overhead from adding these new callbacks is negligible.

**Next steps**

As a first step it would be good to get feedback on the general approach. If we stick with this, and we decide that we'd like it to be part of the public interface, I can consider trying to write some documentation as a follow-up PR.

### DIFF
--- a/src/header.wppl
+++ b/src/header.wppl
@@ -468,8 +468,8 @@ var SampleGuide = function(wpplFn, options) {
 };
 
 var OptimizeThenSample = function(wpplFn, options) {
-  Optimize(wpplFn, _.omit(options, 'samples', 'onlyMAP'));
-  var opts = _.pick(options, 'samples', 'onlyMAP', 'verbose');
+  Optimize(wpplFn, _.omit(options, 'samples', 'onlyMAP', 'callbacks'));
+  var opts = _.pick(options, 'samples', 'onlyMAP', 'callbacks', 'verbose');
   return SampleGuide(wpplFn, opts);
 };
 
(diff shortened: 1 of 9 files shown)


null-a:
I ran this by @stuhlmueller and he was OK with the general approach. (Though he hasn't performed a detailed review of the PR.) I'm going to merge this as it's useful, the tests pass, and it seems low risk.

Issue #902, opened 2018-05-09
Test against latest Node versions


null-a:
It looks like Node v8 will just work, but changes (to our error handling/stack trace bits) will be required to support Node v10.

null-a:
Closed by #920.

PR #920, opened 2019-02-08
Update Node versions exercised by CI.

null-a:


null-a:
Closes #902.

null-a:
This PR is a change to CI config. only, so will merge without review.

PR #925, opened 2019-03-12
Fix bug in Multinomial support.

### LINKED ISSUE #924: Support of multinomial broken by deprecated object() method from lodash
The `support()` method of multinomial (possibly other distributions?) [calls on `object()` method from lodash](https://github.com/probmods/webppl/blob/dev/src/dists/multinomial.ad.js#L50), which has also been [also called `zipObject` in earlier versions](https://lodash.com/docs/3.10.1#zipObject), though now it seems that `object()` is out and [`zipObject()` is still around](https://lodash.com/docs/4.17.11#zipObject). This results in the error `_.chain(...).map(...).object is not a function` with minimal example:

```
Infer({model: function(){
  var x = multinomial([0.5, 0.5], 2)
  return x
}, method: 'enumerate'})
```

Probably, this just requires replacing calls to lodash `object()` with `zipObject()`

null-a:
Fixes #924.


null-a:
@mhtess confirmed this fix is good.

PR #926, opened 2019-03-21
Added a quick start guide from source

kwanghoon:
Added a quick start guide from source

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Upgrade WebPPL:
 
 Clone WebPPL using git:
 
-    git clone https://github.com/kwanghoon/webppl
+    git clone https://github.com/probmods/webppl
 
 Change into the cloned directory:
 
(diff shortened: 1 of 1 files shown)


null-a:
Hi, thanks for submitting this PR! I have a couple of thoughts:

* There's already a section similar to this in the docs [here](http://docs.webppl.org/en/master/development/install.html). Perhaps you missed this, which would suggest that it's hard to find this information? If so, perhaps we could link to the existing section from [Installation](http://docs.webppl.org/en/master/installation.html), rather than add a near duplicate?
* I don't think it's necessary to run the scripts manually, they should be run by `npm install` with recent versions of `npm`.

kwanghoon:
Oh, I see. I was not able to find the existing section for installation from GitHub. 
You are right. It is better just to link to the section. Thanks for your reply, and for this wonderful software!

null-a:
> It is better just to link to the section.

OK, great. I've merged #927 which does this. I'll therefore close this PR, but let us know if #927 doesn't hit the mark. Thanks again for the PR!

PR #936, opened 2020-10-02
Update README.md(please accept my pull req.)

jatin-eleven:


null-a:
This doesn't look like an obvious improvement to me. In particular, we typically refer to the npm package as `webppl`, so changing this here makes things inconsistent at best. The other changes are mostly stylistic and seem like changes for changes sake. Please feel free to re-open/continue this discussion if this addresses something I'm not seeing. Thanks.

